PR #13, opened 2019-08-13
impl fused future for Next

matklad:
This allows using `st.next` in `select!`

### DIFF
--- a/src/stream/stream.rs
+++ b/src/stream/stream.rs
@@ -138,6 +138,15 @@ impl<T: futures::Stream + Unpin + ?Sized> Future for NextFuture<'_, T> {
     }
 }
 
+impl<T> futures::future::FusedFuture for NextFuture<'_, T>
+where
+    T: futures::stream::FusedStream + Unpin + ?Sized,
+{
+    fn is_terminated(&self) -> bool {
+        self.stream.is_terminated()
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
If we switch to use #208 I wonder if this is still needed.

matklad:
Probably not! It would be cool if we can live without fusing, as that’s super confusing. However, if we don’t merge this, perhaps we don’t need https://github.com/async-rs/async-std/pull/40/files either?

yoshuawuyts:
> perhaps we don’t need https://github.com/async-rs/async-std/pull/40/files either

Oh yeah, we probably don't need it for sure, but given the patch was out and pretty much done I figured it wouldn't hurt to merge (for parity with [`Iterator::fuse`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fuse)!) Agreed fusing is confusing tho; and happy to reduce our usage of it as much as possible.

yoshuawuyts:
If we land https://github.com/async-rs/async-std/issues/398 this PR will no longer be necessary, as we'd no longer use `select! {}` which in turn means no more fusing!

stjepang:
Agree with @yoshuawuyts here. Let's close this.

Issue #18, opened 2019-08-14
Change the timeout API?


stjepang:
Timeouts are confusing. @spacejam recently wrote an example that contains the following piece of code:

```rust
stream
    .read_to_end(&mut buf)
    .timeout(Duration::from_secs(5))
    .await?;
```

The problem here is that we need two `?`s after `.await` and it's easy to forget that.

I think the confusing part is in that the `.timeout()` combinator looks like it just *transforms* the future in a similar vein to `.map()` or `.and_then()`, but it really does not!

Instead, `.timeout()` *bubbles* the result of the future so that its type becomes `Result<Result<_, io::Error>, TimeoutError>`.

Perhaps it would be less confusing if `timeout()` was a free-standing function in the `future` module rather than a method on the `time::Timeout` extension trait?

```rust
future::timeout(
    stream.read_to_end(&mut buf),
    Duration::from_secs(5),
)
.await??;
```

This `timeout()` function would stand alongside `ready()`, `pending()`, and maybe some other convenience functions in the `future` module.

Here's another idea. What if we had `io::timeout()` function that resolves to a `Result<_, io::Error>` instead of bubbling the results? Then we could write the following with a single `?`:

```rust
io::timeout(
    stream.read_to_end(&mut buf),
    Duration::from_secs(5),
)
.await?;
```

Now it's also more obvious that we're setting a timeout for an I/O operation and not for an arbitrary future.

In addition to that, perhaps we could delete the whole `time` module? I'm not really a fan of it because it looks nothing like `std::time` and I generally dislike extension traits like `Timeout`.

Note that we already have a time-related function, `task::sleep()`, which is not placed in the `time` module so we probably shouldn't worry about grouping everything time-related into the `time` module. I think it's okay if we have `io::timeout()` and `future::timeout()`.

Finally, here's a really conservative proposal. Let's remove the whole `time` module and only have `io::timeout()`. A more generic function for timeouts can then be left for later design work. I think I prefer this option the most.

sdroege:
> What if we had `io::timeout()` function that resolves to a `Result<_, io::Error>` instead of bubbling the results?

In addition to the current variant? IMHO the current variant is useful on its own for non-IO futures too. You might have some more complicated user defined futures (which might or might not be around IO futures) that don't return a plain `io::Error` but some more complex error type (that might contain an `io::Error` variant somewhere though).

Removing the current variant would seem suboptimal to me.

DoumanAsh:
[...]

PR #20, opened 2019-08-14
Add future::timeout()

stjepang:
cc #18

### DIFF
--- a/src/future/timeout.rs
+++ b/src/future/timeout.rs
@@ -1,0 +1,78 @@
+use std::error::Error;
+use std::fmt;
+use std::pin::Pin;
+use std::time::Duration;
+
+use futures_timer::Delay;
+use pin_utils::unsafe_pinned;
+
+use crate::future::Future;
+use crate::task::{Context, Poll};
+
+/// Awaits a future or times out after a duration of time.
+///
+/// # Examples
+///
+/// ```no_run
+/// # #![feature(async_await)]
+/// # fn main() -> std::io::Result<()> { async_std::task::block_on(async {
+/// #
+/// use std::time::Duration;
+///
+/// use async_std::future;
+///
+/// let never = future::pending::<()>();
+/// let dur = Duration::from_secs(5);
+/// assert!(future::timeout(dur, never).await.is_err());
+/// #
+/// # Ok(()) }) }
+/// ```
+pub async fn timeout<F, T>(dur: Duration, f: F) -> Result<T, TimeoutError>
+where
+    F: Future<Output = T>,
+{
+    let f = TimeoutFuture {
+        future: f,
+        delay: Delay::new(dur),
+    };
+    f.await
+}
+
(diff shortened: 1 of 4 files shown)


src/future/timeout.rs
```diff
@@ -0,0 +1,78 @@
...
+impl<F> TimeoutFuture<F> {
+    unsafe_pinned!(future: F);
+    unsafe_pinned!(delay: Delay);
+}
+
+impl<F: Future> Future for TimeoutFuture<F> {
+    type Output = Result<F::Output, TimeoutError>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.as_mut().future().poll(cx) {
```
kpp: Does it mean that in case of a CPU-bound future that is blocking the current thread you will never try to poll the timer and return TimeoutError?

src/future/timeout.rs
```diff
@@ -0,0 +1,78 @@
...
+/// ```no_run
+/// # #![feature(async_await)]
+/// # fn main() -> std::io::Result<()> { async_std::task::block_on(async {
+/// #
+/// use std::time::Duration;
+///
+/// use async_std::future;
+///
+/// let never = future::pending::<()>();
+/// let dur = Duration::from_secs(5);
```
yoshuawuyts:
```suggestion
/// let dur = Duration::from_millis(5);
```

If we make the duration real short we can actually run this assertion! (:

src/future/timeout.rs
```diff
@@ -0,0 +1,78 @@
...
+use pin_utils::unsafe_pinned;
+
+use crate::future::Future;
+use crate::task::{Context, Poll};
+
+/// Awaits a future or times out after a duration of time.
+///
+/// # Examples
+///
+/// ```no_run
```
yoshuawuyts:
```suggestion
/// ```
```
If we accept the suggestion below we can actually run this test!

kpp:
Also notice how this code can be reduced: https://github.com/async-rs/async-std/pull/113

kpp:
BTW what's the difference between future::timeout() and io::timeout()?

src/future/timeout.rs
stjepang: @kpp That's a good point - do you think we should check for timeout before polling the future?

I worry a bit that this would add overhead to futures that are quick to poll or immediately ready. And if the future is CPU-bound and takes a really long time, we can still get stuck if the timeout occurs while that future is running.

All things considered, I think I prefer the current behavior. Wdyt?

kpp: > All things considered, I think I prefer the current behavior.

Yep... I don't see a better solution for this.

Still I would rewrite the current implementation as `f.select(Delay::new(dur))...` ;)

stjepang:
I put this function behind the new `unstable` feature flag.

PR #21, opened 2019-08-14
Add a small example on how to integrate threads

skade:
Wrote a small example on how to integrate threads and use a oneshot channel to figure out if a thread is finished. It wraps that all in a way that allows for panicking threads and makes sure that you can get the thread result as you are used to.

This is my first take at it, does it make sense/is it useful?

One jarring oddity is `join().await`, but I didn't want to go the whole way and implement a future in whole. We could do it, though, it's just a proxy.

### DIFF
--- a/examples/integrate-thread.rs
+++ b/examples/integrate-thread.rs
@@ -1,0 +1,74 @@
+#![feature(async_await)]
+
+use async_std::task;
+
+use std::{thread,time};
+use futures::channel::oneshot;
+
+struct AsyncHandle<T> {
+    handle: thread::JoinHandle<T>,
+    notifier: oneshot::Receiver<()>,
+}
+
+impl<T> AsyncHandle<T> {
+    fn thread(&self) -> &std::thread::Thread {
+        self.handle.thread()
+    }
+
+    async fn join(self) -> std::thread::Result<T> {
+        // ignore results, the join handle will propagate panics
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
I feel this is somewhat superseded now that we have the `task::spawn_blocking` API. This example could now more or less be rewritten as:

```rust
fn main() {
    task::block_on(async move {
        let handle = task::spawn_blocking(move || {
            thread::sleep(time::Duration::from_millis(1000));
            String::from("Finished")
        });

        println!("waiting for sleepy thread");
        match handle.await {
            Ok(s) => println!("Result: {}", s),
            Err(e) => println!("Error: {:?}", e),
        }

        let handle = task::spawn_blocking(move || {
            thread::spawn(|| {
                panic!("aaah!");
                String::from("Finished!")
            }).join()
        });

        println!("waiting for panicking thread");
        match handle.await {
            Ok(s) => println!("Result: {}", s),
            Err(e) => println!("Error: {:?}", e),
        }
    });
}
```

I'm not sure how much we'd want to have an example of this (could be useful!) or perhaps polish the docs a bit more to make it apparent this exists (discoverability can be tricky). But overall I think it'd be nice if we could find a resolution to this PR to not keep it open indefinitely. Thanks!

dignifiedquire:
Closing for now, please reopen if you want to update it to the latest version.

PR #32, opened 2019-08-15
Stagger background thread spin-down to avoid a thundering herd

spacejam:


spacejam:
@yoshuawuyts This is an autoscaling system. One of the most common mistakes when building such systems is to release processing resources too quickly, repeatedly incurring latency hiccups during spin-up when load bursts.

That said, this would probably be better if it followed a stack-based threadpool approach, rather than a FIFO which will keep threads alive as long as they receive any work before their timeout, keeping 10k threads open without all that much actually happening.

PR #34, opened 2019-08-15
Stream::cycle implementation

vertexclique:
* Adds `stream::cycle` implementation
* Contains an example for stream cycle.

### DIFF
--- a/examples/stream-cycle.rs
+++ b/examples/stream-cycle.rs
@@ -1,0 +1,25 @@
+//! Repeats given stream values and sum them
+
+#![feature(async_await)]
+
+use async_std::io;
+use async_std::prelude::*;
+use async_std::stream;
+use async_std::task;
+
+fn main() -> io::Result<()> {
+    task::block_on(async {
+        let mut s = stream::cycle(vec![6, 7, 8]);
+        let mut total = 0;
+
+        while let Some(v) = s.next().await {
+            total += v;
+            if total == 42 {
+                println!("Found {} the meaning of life!", total);
+                break;
+            }
+        }
+
+        Ok(())
+    })
+}
--- a/src/stream/cycle.rs
+++ b/src/stream/cycle.rs
@@ -1,0 +1,57 @@
+use std::pin::Pin;
+use std::sync::Mutex;
+
+use crate::task::{Context, Poll};
+
+/// Creates a stream that yields the given elements continually.
+///
+/// # Examples
+///
+/// ```
+/// # #![feature(async_await)]
+/// # fn main() { async_std::task::block_on(async {
+/// #
+/// use async_std::prelude::*;
+/// use async_std::stream;
+///
+/// let mut s = stream::cycle(vec![1, 2, 3]);
+///
+/// assert_eq!(s.next().await, Some(1));
+/// assert_eq!(s.next().await, Some(2));
+/// assert_eq!(s.next().await, Some(3));
+/// assert_eq!(s.next().await, Some(1));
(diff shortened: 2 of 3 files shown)


src/stream/cycle.rs
```diff
@@ -0,0 +1,57 @@
...
+///
+/// assert_eq!(s.next().await, Some(1));
+/// assert_eq!(s.next().await, Some(2));
+/// assert_eq!(s.next().await, Some(3));
+/// assert_eq!(s.next().await, Some(1));
+/// assert_eq!(s.next().await, Some(2));
+/// #
+/// # }) }
+/// ```
+pub fn cycle<T>(items: Vec<T>) -> Cycle<T>
```
kpp:
```
pub fn cycle<I>(i: I) -> impl Stream<Item = <I as IntoIterator>::Item> where
    I: IntoIterator, 
{
    futures::stream::iter(i.into_iter().cycle())
}
```

kpp: No need to over engineer the issue. BTW I suggest you adding this method into futures-rs.

examples/stream-cycle.rs
```diff
@@ -0,0 +1,25 @@
...
+            total += v;
+            if total == 42 {
+                println!("Found {} the meaning of life!", total);
+                break;
+            }
+        }
+
+        Ok(())
+    })
+}
```
yoshuawuyts: This example is cool, but I think it's probably fine to just have the doc test for now. In #131 and #129 there are quite a few more methods, and I think it'd be wise to try and keep our line count down where possible.

src/stream/cycle.rs
yoshuawuyts: We want the `Cycle` counterpart to stdlib. Also we've clarified in a different issue that we do actually want our own concrete impls.

src/stream/cycle.rs
```diff
@@ -0,0 +1,57 @@
...
+///
+/// ```
+/// # #![feature(async_await)]
+/// # fn main() { async_std::task::block_on(async {
+/// #
+/// use async_std::prelude::*;
+/// use async_std::stream;
+///
+/// let mut s = stream::cycle(vec![1, 2, 3]);
+///
```
yoshuawuyts: In std the API is:

```rust
let a = [1, 2, 3];

let mut it = a.iter().cycle();

assert_eq!(it.next(), Some(&1));
assert_eq!(it.next(), Some(&2));
assert_eq!(it.next(), Some(&3));
assert_eq!(it.next(), Some(&1));
assert_eq!(it.next(), Some(&2));
assert_eq!(it.next(), Some(&3));
assert_eq!(it.next(), Some(&1));
```

I'd expect this API to be (with #125 landed):
```rust
let a = [1, 2, 3];

let mut it = a.into_stream().cycle();

assert_eq!(it.next().await, Some(&1));
assert_eq!(it.next().await, Some(&2));
assert_eq!(it.next().await, Some(&3));
assert_eq!(it.next().await, Some(&1));
assert_eq!(it.next().await, Some(&2));
assert_eq!(it.next().await, Some(&3));
assert_eq!(it.next().await, Some(&1));
```

Similar, but subtly different in that it's more generic, which should make it more widely applicable!

yoshuawuyts:
#125 has landed now which means this can be modeled after std's API. @vertexclique do you have any interest in continuing with this patch in that direction?

yoshuawuyts:
It's been a while, and I'm going to assume this PR will not be continued. Thanks for the work! Going to go ahead and close this.

PR #40, opened 2019-08-16
Add initial Fuse implementation for Stream

spacejam:
@matklad does this address your use case?

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -27,7 +27,7 @@ pub use from_stream::FromStream;
 pub use into_stream::IntoStream;
 pub use once::{once, Once};
 pub use repeat::{repeat, Repeat};
-pub use stream::{Scan, Stream, Take, Zip};
+pub use stream::{Fuse, Scan, Stream, Take, Zip};
 
 mod double_ended_stream;
 mod empty;
--- a/src/stream/stream/fuse.rs
+++ b/src/stream/stream/fuse.rs
@@ -1,51 +1,29 @@
 use std::pin::Pin;
+use std::task::{Context, Poll};
 
-use crate::task::{Context, Poll};
-
-/// A `Stream` that is permanently closed
-/// once a single call to `poll` results in
-/// `Poll::Ready(None)`, returning `Poll::Ready(None)`
-/// for all future calls to `poll`.
+/// A `Stream` that is permanently closed once a single call to `poll` results in
+/// `Poll::Ready(None)`, returning `Poll::Ready(None)` for all future calls to `poll`.
 #[derive(Clone, Debug)]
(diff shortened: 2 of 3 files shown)


src/stream/mod.rs
```diff
@@ -25,7 +25,7 @@
 pub use empty::{empty, Empty};
 pub use once::{once, Once};
 pub use repeat::{repeat, Repeat};
-pub use stream::{Stream, Take};
+pub use stream::{Fuse, Stream, Take};
```
yoshuawuyts: Sticking with the `std` shouldn't this be `async_std::stream::FusedStream` as a counterpart to `std::iter::FusedIterator`? I know we haven't been really naming things that way, but figured it'd be worth bringing up.

cc/ @stjepang thoughts on naming?

skade: It should.

matklad: FusedIterator is a trait,the struct is named Fuse: https://doc.rust-lang.org/std/iter/struct.Fuse.html.

yoshuawuyts: Ah my bad, thanks for catching that @matklad!

src/stream/stream.rs
```diff
@@ -171,3 +203,54 @@ impl<S: futures::Stream> futures::Stream for Take<S> {
...
+    pub fn is_done(&self) -> bool {
+        self.done
+    }
+
+    /// Consumes this `Fuse` and returns the inner
+    /// `Stream`, unfusing it if it had become
+    /// fused.
+    pub fn into_inner(self) -> S
+    where
+        S: Sized,
```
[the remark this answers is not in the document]
yoshuawuyts: Added back the `where Self: Sized` bound in the `fn fused()` impl. I couldn't get this to compile with `S: ?Sized` in the impl bound.

src/stream/stream.rs
```diff
@@ -114,6 +114,38 @@ pub trait Stream {
...
+    /// # fn main() { async_std::task::block_on(async {
+    /// #
+    /// use async_std::prelude::*;
+    /// use async_std::stream;
+    ///
+    /// let mut s = stream::repeat(9).take(3);
+    ///
+    /// while let Some(v) = s.next().await {
+    ///     assert_eq!(v, 9);
+    /// }
```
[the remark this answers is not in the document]
yoshuawuyts: I've updated the example!

yoshuawuyts:
I've rebased @spacejam's excellent work on master; it's all passing now. Also removed the two helper methods on the `Fuse` struct that don't have counterparts in std so we only add the minimal API surface for now.

Given the amount of review this has had, I think this should be good to merge. Ref #129. 

bors r+

PR #45, opened 2019-08-16
Implement a simple Map transformer for Stream

spacejam:


yoshuawuyts:
Ref #129

yoshuawuyts:
cc/ @montekki probably the methods we're missing the most right now are map/for_each and try_map/try_for_each. This PR is a bit stale, and I think we should close it. But would you be willing to implement those methods for Stream? 

Thanks heaps!

montekki:
@yoshuawuyts ok

montekki:
@yoshuawuyts is `try_map` a thing? Other three are ready.

yoshuawuyts:
@montekki `try_map` isn't a thing, unfortunately. I think it's probably okay to leave it out for now.

PR #51, opened 2019-08-16
Fix links in concepts/futures.md

luizirber:
Fixes some missing links in concepts/future.md


luizirber:
There is also a link missing for [Future](https://github.com/async-rs/async-std/pull/51/files#diff-a13e1331c067da90c2a2dd2624f976baL29), but not sure where to point it to.

skade:
Thanks!

skade:
I think the [Future] link should go to std-futures.

PR #52, opened 2019-08-16
Rename 'task' variable in example

skorgu:
The collision between the 'task' variable and the 'task' import is confusing (task::block_on(task) especially).


skade:
Right, thank you here, too :)

PR #53, opened 2019-08-16
Fix some stray formatting

skorgu:


skade:
Thanks again for those 3 PRs!

PR #56, opened 2019-08-16
small typos and edits in first few pages

dcarosone:
Some small typos, markdown lints and other wording tweaks I noticed while reading. Hopefully nothing controversial.

Just a few pages in so far; I'll add more as I go.

One question: the current content uses "directional" apostrophes, it’s vs it's, that gets caught by my spell-checker. Before I go changing them, is this deliberate / preferred?


skade:
On apostrophes, I _think_ that's what the markdown implementation does... Or can you point me at a line that does that?

skade:
Fixed a merge problem.

dcarosone:
> On apostrophes, I _think_ that's what the markdown implementation does... Or can you point me at a line that does that?

I don't mean in the rendered output (from the markdown implementation), I mean in the source, eg:

https://github.com/async-rs/async-std/blame/master/docs/src/concepts/futures.md#L108

skade:
Right, I see. That's a bug, it only occurs at 3 places. I'll fix that, thanks!

jamesmunns:
Thanks!

PR #61, opened 2019-08-16
Manual: automated link checking using mdbook-linkcheck

ngirard:
MDBook LinkCheck is a backend for mdbook which check both internal and
external links (https://github.com/Michael-F-Bryan/mdbook-linkcheck).

This PR leaves external links unchecked for performance reasons ;
they can be checked by setting `follow-web-links = true` in the
`book.toml` config file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ env:
 
 before_script:
   - rustup component add rustfmt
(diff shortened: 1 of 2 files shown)


skade:
Hi, I'll rebase and check that later, currently got terrible wifi. I wanted to add this, but ran out of time. Thanks for doing it!

PR #71, opened 2019-08-17
Switch on testing for the book; fix #6

### LINKED ISSUE #6: Switch on testing for the book
Required:

* [ ] all code examples compile
* [ ] `mdbook build docs` is added to `.travis.yml`

freebroccolo:
Run the tests with the following:

```
cargo build    # first build async_std
mdbook test -L ./target/debug/deps docs
```

The `-L` flag is needed so that external crates can be resolved in the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,12 @@ language: rust
 env:
   - RUSTFLAGS="-D warnings"
 
-before_script:
-  - rustup component add rustfmt
-
 matrix:
   fast_finish: true
   include:
     - rust: nightly
       os: linux
-      env: BUILD_DOCS=1
+      env: BUILD_DOCS=1 BUILD_BOOK=1
     - rust: nightly
       os: osx
       osx_image: xcode9.2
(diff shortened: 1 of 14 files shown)


skade:
Wow, thanks already for that :).

rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2019-08-16
```
freebroccolo: I've pinned the toolchain to `nightly-2019-08-16` because the build fails for some of the rustdoc tests on current nightly with an error message about not being able to find `proc_macro_call_0`. Seems related to this: https://github.com/rust-lang/rust/issues/63651

freebroccolo:
I haven't tested the changes to the travis build but otherwise I think this is ready.

skade:
This will fail ;). What's missing is a `before_script` similar to here: https://rust-lang-nursery.github.io/mdBook/continuous-integration.html

freebroccolo:
I'm not sure how to fix this latest error from `mdbook test` about multiple matching crates for `async_std`. I saw something similar locally but only once and it seemed to resolve itself. Any ideas? By the way, feel free to make whatever changes needed to this, stylistically or otherwise.

freebroccolo:
Note that this fixes #89.

.travis.yml
```diff
@@ -3,24 +3,28 @@ language: rust
...
+  - rustup component add rustfmt
+  - (test -x $HOME/.cargo/bin/cargo-install-update || cargo install cargo-update)
+  - (test -x $HOME/.cargo/bin/mdbook || cargo install --vers "^0.3" mdbook)
+  - cargo install-update -a
+
 script:
-  - cargo check --all --benches --bins --examples --tests
-  - cargo test --all
+  - if ![[ -n "$BUILD_BOOK" ]]; then cargo check --all --benches --bins --examples --tests && cargo test --all; fi
+  - if  [[ -n "$BUILD_BOOK" ]]; then cargo test  --all --benches --bins --examples --tests; fi
```
freebroccolo: I've had to modify the travis config to where, when the book should be built, we don' run `cargo check` but instead just run `cargo test`.

This might cause ci to not fail as quickly on the linux build but the problem is that running `cargo check` followed by `cargo test` produces two entries for `async_std` in `./target/debug/deps` which then causes `mdbook test` to fail when trying to resolve `extern crate async_std`.

I don't know of another way to fix that issue. If this is an acceptable workaround then I think this PR is good to go.

skade:
This needs to be updated to remove all uses of `#[feature(async_await)]`

freebroccolo:
I've removed the `#[feature(async_await)]` and squashed some of the commits.

freebroccolo:
I've rebased again and fixed the latest set of merge conflicts. /cc @skade @stjepang

PR #73, opened 2019-08-18
Clarify async fn return type

U007D:
You guys are absolutely amazing. Thank you for all your hard work.
<3


skade:
Thanks for the PR!

PR #74, opened 2019-08-18
Implement an async version of ToSocketAddrs

### LINKED ISSUE #50: TcpListener::bind, etc. may block
TCP/UDP listeners and streams take `A: std::net::ToSocketAddrs`. Some `ToSocketAddrs` impls will resolve domain names. However, `ToSocketAddrs` is not futures-aware, so DNS lookups are synchronous.

Possible solutions:
- Spawn a blocking task a la filesystem operations (simpler)
- Replace `ToSocketAddrs` with an async version (probably more efficient)

DCjanus:
Resolve #50


src/net/addr.rs
```diff
@@ -0,0 +1,303 @@
...
+    /// to.
+    type Iter: Iterator<Item = SocketAddr> + Send;
+    /// Returned Future
+    type Output: Future<Output = crate::io::Result<Self::Iter>> + Send;
+    /// Converts this object to an iterator of resolved `SocketAddr`s.
+    ///
+    /// The returned iterator may not actually yield any values depending on the
+    /// outcome of any resolution performed.
+    ///
+    /// Note that this function may block the current thread while resolution is
```
skade: This note should go, as the function should _not _ block the current thread :).

DCjanus: That's my stupid mistake, I'm going to fix this.

src/net/addr.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// Returned Future
+    type Output: Future<Output = crate::io::Result<Self::Iter>> + Send;
+    /// Converts this object to an iterator of resolved `SocketAddr`s.
+    ///
+    /// The returned iterator may not actually yield any values depending on the
+    /// outcome of any resolution performed.
+    ///
+    /// Note that this function may block a backend thread while resolution is
+    /// performed.
+    fn to_socket_addrs(&self) -> Self::Output;
```
stjepang: This method would ideally be an `async fn`, which is currently a feature supported by the language. So it'd be best if we could design the API in such a way that switching to `async fn` is not a breaking change once that becomes possible.

I suggest replacing the `Output` associated type with a hack similar to `async_std::stream::Stream` and other traits in the library, where we return a hidden concrete type, but "lie" it's actually an `impl Future`  in the generated docs.

src/net/addr.rs
```diff
@@ -0,0 +1,302 @@
+use crate::future::Future;
+use crate::task::blocking::JoinHandle;
+use futures::future::{ready, Ready};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+pub use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
+
+/// A trait for objects which can be converted or resolved to one or more
+/// [`SocketAddr`] values.
```
stjepang: This link is missing an URL.

DCjanus: I've reexport `SocketAddr`, so maybe the URL is not required? (I've tried `cargo doc` and the link isn't broken.)

src/net/addr.rs
DCjanus: Pushed some code for this, there is a different with other `ret!`, but I think that's ok.

DCjanus:
Force pushd for rebase master.

@stjepang Hi, after a long day, I've created a commit to replace `std::net::ToSocketAddrs` with `async_std::net::ToSocketAddrs`, would you mind review again. Thanks.

stjepang:
Could you perhaps incorporate changes from https://github.com/DCjanus/async-std/pull/1 here? In particular, let's remove the new re-exports and move tests into the `tests` directory. At the same time, let's also delete unit tests that are not relevant to `ToSocketAddrs`.

src/net/addr.rs
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    // FIXME: figure out why this fails on openbsd and fix it
+    #[test]
+    #[cfg(not(any(windows, target_os = "openbsd")))]
+    fn to_socket_addr_str_bad() {
+        assert!(tsa("1200::AB00:1234::2552:7777:1313:34300").is_err());
+    }
+
+    #[test]
```
kpp: Irrelevant tests starting from here. No need to test `std::net::...`

DCjanus: Thanks for your help, this should be resolved.

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+
+use crate::future::Future;
+use crate::io;
+use crate::task::blocking;
+use crate::task::{Context, Poll};
+
+cfg_if! {
+    if #[cfg(feature = "docs")] {
+        #[doc(hidden)]
+        pub struct ImplFuture<T>(std::marker::PhantomData<T>);
```
kpp: @skade @stjepang do we really need to play with feature=docs?

tests/addr.rs
```diff
@@ -0,0 +1,77 @@
...
+
+fn tsa<A: ToSocketAddrs>(a: A) -> Result<Vec<SocketAddr>, String> {
+    let socket_addrs = task::block_on(a.to_socket_addrs());
+    match socket_addrs {
+        Ok(a) => Ok(a.collect()),
+        Err(e) => Err(e.to_string()),
+    }
+}
+
+pub fn sa4(a: Ipv4Addr, p: u16) -> SocketAddr {
```
kpp: Still you did not test ToSocketAddrs::to_socket_addrs for (Ipv4Addr, u16). This is a useless function to create SocketAddr

tests/addr.rs
```diff
@@ -0,0 +1,77 @@
...
+    #[cfg(target_env = "sgx")]
+    let _ = a;
+}
+
+#[test]
+fn to_socket_addr_string() {
+    let a = sa4(Ipv4Addr::new(77, 88, 21, 11), 24352);
+    assert_eq!(Ok(vec![a]), tsa(&*format!("{}:{}", "77.88.21.11", "24352")));
+    assert_eq!(Ok(vec![a]), tsa(&format!("{}:{}", "77.88.21.11", "24352")));
+    assert_eq!(Ok(vec![a]), tsa(format!("{}:{}", "77.88.21.11", "24352")));
```
kpp: No need to test how format! works

tests/addr.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    let s = format!("{}:{}", "77.88.21.11", "24352");
+    assert_eq!(Ok(vec![a]), tsa(s));
+    // s has been moved into the tsa call
```
kpp: No need to test Rust ownership

tests/addr.rs
```diff
@@ -0,0 +1,77 @@
...
+fn to_socket_addr_ipaddr_u16() {
+    let a = Ipv4Addr::new(77, 88, 21, 11);
+    let p = 12345;
+    let e = SocketAddr::V4(SocketAddrV4::new(a, p));
+    assert_eq!(Ok(vec![e]), tsa((a, p)));
+}
+
+#[test]
+fn to_socket_addr_str_u16() {
+    let a = sa4(Ipv4Addr::new(77, 88, 21, 11), 24352);
```
kpp: Please inline `SocketAddr::V4(SocketAddrV4::new(a, p))` instead of sa4

tests/addr.rs
```diff
@@ -0,0 +1,77 @@
...
+    #[cfg(target_env = "sgx")]
+    let _ = a;
+}
+
+#[test]
+fn to_socket_addr_str() {
+    let a = sa4(Ipv4Addr::new(77, 88, 21, 11), 24352);
+    assert_eq!(Ok(vec![a]), tsa("77.88.21.11:24352"));
+
+    let a = sa6(Ipv6Addr::new(0x2a02, 0x6b8, 0, 1, 0, 0, 0, 1), 53);
```
kpp: Please inline `SocketAddr::V6` instead of sa46

tests/addr.rs
```diff
@@ -0,0 +1,77 @@
+use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
+
+use async_std::net::ToSocketAddrs;
+use async_std::task;
+
+fn tsa<A: ToSocketAddrs>(a: A) -> Result<Vec<SocketAddr>, String> {
```
kpp: rename tsa -> blocking_resolve

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// The returned iterator may not actually yield any values depending on the outcome of any
+    /// resolution performed.
+    ///
+    /// Note that this function may block a backend thread while resolution is performed.
+    fn to_socket_addrs(&self) -> ret!(ToSocketAddrsFuture, Self::Iter);
+}
+
+#[doc(hidden)]
+#[allow(missing_debug_implementations)]
+pub enum ToSocketAddrsFuture<I: Iterator<Item = SocketAddr>> {
```
kpp: Looks like https://docs.rs/futures-preview/0.3.0-alpha.18/futures/future/enum.Either.html where A = locking::JoinHandle<io::Result<I>> , B = Ready<io::Result<I>>

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+    Join(blocking::JoinHandle<io::Result<I>>),
+    Ready(Ready<io::Result<I>>),
+}
+
+impl<I: Iterator<Item = SocketAddr>> Future for ToSocketAddrsFuture<I> {
+    type Output = io::Result<I>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
kpp: https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-util/src/future/either.rs#L52

src/net/addr.rs
DCjanus: There is no `impl Stream for JoinHandle`, which means `Either<JoinHandle<io::Result<I>, Ready<io::Result<I>>>>` isn't `Stream`.

DCjanus: I don't want to change this, because in the future, there might be more impl and more `Future` types, which means `Either` might not be a good idea.

tests/addr.rs
```diff
@@ -3,75 +3,85 @@ use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6};
...
 #[test]
 fn to_socket_addr_string() {
-    let a = sa4(Ipv4Addr::new(77, 88, 21, 11), 24352);
-    assert_eq!(Ok(vec![a]), tsa(&*format!("{}:{}", "77.88.21.11", "24352")));
-    assert_eq!(Ok(vec![a]), tsa(&format!("{}:{}", "77.88.21.11", "24352")));
-    assert_eq!(Ok(vec![a]), tsa(format!("{}:{}", "77.88.21.11", "24352")));
+    let a = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(77, 88, 21, 11), 24352));
+    assert_eq!(
+        Ok(vec![a]),
+        blocking_resolve(&*format!("{}:{}", "77.88.21.11", "24352"))
```
kpp: These tests with `format!("{}:{}"` are identical to `blocking_resolve(&format!("{}:{}", "77.88.21.11", "24352"))`. No need to duplicate tests

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    fn to_socket_addrs(&self) -> ret!(ToSocketAddrsFuture, Self::Iter) {
+        ToSocketAddrsFuture::Ready(ready(std::net::ToSocketAddrs::to_socket_addrs(self)))
+    }
+}
+
+impl ToSocketAddrs for SocketAddrV4 {
+    type Iter = std::option::IntoIter<SocketAddr>;
+
+    fn to_socket_addrs(&self) -> ret!(ToSocketAddrsFuture, Self::Iter) {
```
kpp: This one is not tested. Use https://github.com/roblabla/cargo-travis/ and `cargo coverage` and open report `target/kcov/index.html`.

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+        ToSocketAddrsFuture::Ready(ready(std::net::ToSocketAddrs::to_socket_addrs(self)))
+    }
+}
+
+impl ToSocketAddrs for SocketAddrV6 {
+    type Iter = std::option::IntoIter<SocketAddr>;
+
+    fn to_socket_addrs(&self) -> ret!(ToSocketAddrsFuture, Self::Iter) {
```
kpp: This one is not tested. Use https://github.com/roblabla/cargo-travis/ and `cargo coverage` and open report `target/kcov/index.html`.

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+        ToSocketAddrsFuture::Ready(ready(std::net::ToSocketAddrs::to_socket_addrs(self)))
+    }
+}
+
+impl ToSocketAddrs for (IpAddr, u16) {
+    type Iter = std::option::IntoIter<SocketAddr>;
+
+    fn to_socket_addrs(&self) -> ret!(ToSocketAddrsFuture, Self::Iter) {
```
kpp: This one is not tested. Use https://github.com/roblabla/cargo-travis/ and `cargo coverage` and open report `target/kcov/index.html`.

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+        ToSocketAddrsFuture::Ready(ready(std::net::ToSocketAddrs::to_socket_addrs(self)))
+    }
+}
+
+impl ToSocketAddrs for (Ipv6Addr, u16) {
+    type Iter = std::option::IntoIter<SocketAddr>;
+
+    fn to_socket_addrs(&self) -> ret!(ToSocketAddrsFuture, Self::Iter) {
```
kpp: This one is not tested. Use https://github.com/roblabla/cargo-travis/ and `cargo coverage` and open report `target/kcov/index.html`.

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+        let join =
+            blocking::spawn(async move { std::net::ToSocketAddrs::to_socket_addrs(&socket_addrs) });
+        ToSocketAddrsFuture::Join(join)
+    }
+}
+
+impl<'a> ToSocketAddrs for &'a [SocketAddr] {
+    type Iter = std::iter::Cloned<std::slice::Iter<'a, SocketAddr>>;
+
+    fn to_socket_addrs(&self) -> ret!(ToSocketAddrsFuture, Self::Iter) {
```
kpp: This one is not tested. Use https://github.com/roblabla/cargo-travis/ and `cargo coverage` and open report `target/kcov/index.html`.

src/net/addr.rs
stjepang: It makes generated docs look prettier and describes the API's intention better.

We just really want to make docs be as approachable as possible, especially given the fact that docs is an area async programming in Rust has historically not excelled at.

tests/addr.rs
DCjanus: Those code are from std, which test different types: `String`、`&String`、`&str`, I've changed them, new code would be like this:

```Rust
    let s: &str = "77.88.21.11:24352";
    assert_eq!(Ok(vec![a]), blocking_resolve(s));

    let s: &String = &"77.88.21.11:24352".to_string();
    assert_eq!(Ok(vec![a]), blocking_resolve(s));

    let s: String = "77.88.21.11:24352".to_string();
    assert_eq!(Ok(vec![a]), blocking_resolve(s));
```

src/net/addr.rs
DCjanus: To be honest, I don't want to test those code, they are too simple. @stjepang  what do you think?

src/net/addr.rs
yoshuawuyts: Agree with @dcjanus here.

src/net/addr.rs
yoshuawuyts: @dcjanus I think it's okay to be pragmatic about what we test; probably okay to skip this (:

skade:
@stjepang is this ready to merge?

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// Returned iterator over socket addresses which this type may correspond to.
+    type Iter: Iterator<Item = SocketAddr> + Send;
+
+    /// Converts this object to an iterator of resolved `SocketAddr`s.
+    ///
+    /// The returned iterator may not actually yield any values depending on the outcome of any
+    /// resolution performed.
+    ///
+    /// Note that this function may block a backend thread while resolution is performed.
+    fn to_socket_addrs(&self) -> ret!(ToSocketAddrsFuture, Self::Iter);
```
stjepang: Let's parametrize the `ToSocketAddrsFuture` by the lifetime in `&self` so that it becomes `ToSocketAddrsFuture<'_>`.

The reason is that this method will become an `async fn` at some point and the returned future in `async fn` gets desugared so that it borrows all function arguments (or, rather, their lifetimes).

src/net/addr.rs
```diff
@@ -0,0 +1,159 @@
...
+
+/// A trait for objects which can be converted or resolved to one or more [`SocketAddr`] values.
+///
+/// This trait is an async version of [`std::net::ToSocketAddrs`].
+///
+/// [`std::net::ToSocketAddrs`]: https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html
+/// [`SocketAddr`]: https://doc.rust-lang.org/std/net/enum.SocketAddr.html
+pub trait ToSocketAddrs {
+    /// Returned iterator over socket addresses which this type may correspond to.
+    type Iter: Iterator<Item = SocketAddr> + Send;
```
stjepang: Can we remove the `Send` bound here and instead use it in the `ToSocketAddrs for &T` impl block?

docs/src/tutorial/accept_loop.md
```diff
@@ -6,23 +6,20 @@ First of all, let's add required import boilerplate:
...
 ```rust,edition2018
 # extern crate async_std;
-use std::net::ToSocketAddrs; // 1
 use async_std::{
-    prelude::*, // 2
-    task, // 3
-    net::TcpListener, // 4
+    prelude::*, // 1
+    task, // 2
+    net::{TcpListener, ToSocketAddrs}, // 3
```
stjepang: Not sure why these docs got modified by this PR. Can you try rebasing on the master branch or do whatever is necessary to remove these changes in docs?

DCjanus: In this PR, implemented an async ToSocketAddrs, so I replaced std::net::ToSocketAddrs to async_std::net::ToSocketAddrs. If you want to show "we're able to use block API", I'm going to revert this.

stjepang: Oh I'm sorry, I wasn't taking a good look! Please ignore this comment :)

src/net/addr.rs
DCjanus: removed

src/net/addr.rs
DCjanus: dirty code in this commit: https://github.com/async-rs/async-std/pull/74/commits/83636b9bfed28a4bbadd0f271cdfc14a520612fd

I'm not sure is that a code style?

PR #76, opened 2019-08-18
Revert "Manual: automated link checking using mdbook-linkcheck"

jamesmunns:
Reverts async-rs/async-std#61

This broke the netlify build.


jamesmunns:
Already fixed in https://github.com/async-rs/async-std/commit/3b9ee1ad483678627dcef00b17fff045f684718a

PR #80, opened 2019-08-19
Fix a couple of typos in the book

wezm:
Just a few small fixes for typos I noticed while reading through the book.

The difference is the first diff is the insertion of, ", and", in "the language side, and expects you to enforce the "losing access" behaviour yourself."


skade:
Thanks!

PR #92, opened 2019-08-20
draft Buf writer

### LINKED ISSUE #83: Add BufWriter
There is already a `BufReader`, so people would expect to have a `BufWriter` too. A lazy implementation is to simply wrap `BufWriter` from `futures-0.3`.

Drevoed:
This pull request will resolve #83 when merged.

It includes asynchronous BufWriter and LineWriter with tests.
~Currently I'm having problems with implementing `poll_write` for BufWriter, because trying to call 
`self.buf().write(buf).poll(cx)` results in~
```rust
error[E0599]: no method named `poll` found for type `io::write::WriteFuture<'_, std::vec::Vec<u8>>` in the current scope
   --> src/io/buf_writer.rs:239:33
    |
239 |             self.buf().write(buf).poll(cx)
    |                                 ^^^^
    |
   ::: src/io/write.rs:149:1
    |
149 | pub struct WriteFuture<'a, T: Unpin + ?Sized> {
    | --------------------------------------------- method `poll` not found for this
    |
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following traits define an item `poll`, perhaps you need to implement one of them:
            candidate #1: `core::future::future::Future`
            candidate #2: `futures::future::Future`
            candidate #3: `futures::stream::Stream`
```

~I assume that's because WriteFuture has trait bounds on ?Sized but I'm not sure how to get around that.~

The above error was fixed by wrapping the buf in Pin explicitly.

Previous PR was closed because I forgot to mark it as draft.


src/io/buf_writer.rs
```diff
@@ -0,0 +1,547 @@
...
+/// [`TcpStream::write`]: ../net/struct.TcpStream.html#method.write
+/// [`TcpStream`]: ../net/struct.TcpStream.html
+/// [`flush`]: trait.Write.html#tymethod.flush
+pub struct BufWriter<W> {
+    inner: W,
+    buf: Vec<u8>,
+    written: usize,
+}
+
+impl<W: AsyncWrite + Unpin> BufWriter<W> {
```
stjepang: I think we can remove the `Unpin` bound here and in other impl blocks.

src/io/buf_writer.rs
```diff
@@ -0,0 +1,547 @@
...
+    pub fn buffer(&self) -> &[u8] {
+        &self.buf
+    }
+
+    /// Poll buffer flushing until completion
+    ///
+    /// This is used in types that wrap around BufWrite, one such example: [`LineWriter`]
+    ///
+    /// [`LineWriter`]: struct.LineWriter.html
+    pub fn poll_flush_buf(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
```
stjepang: Let's make this method private. I'm not against making it public, but would consider it after publishing version 1.0. For now, let's just match the standard library's API...

yoshuawuyts:
Ref #131 also (tracking issue for `async_std::io`)

yoshuawuyts:
@Drevoed o/ wanted to check in how things are. Is there anything we can do to help you get unblocked?

Drevoed:
@yoshuawuyts I'm currently a bit busy with day-time job but will try to implement requested changes ASAP, I don't think I will have any problems, but if I will - I am going to ask for assistance surely :)

src/io/buf_writer.rs
Drevoed: done!

PR #100, opened 2019-08-22
[docs] clearer variable name for broker_handle

zvkemp:
- use `broker_handle` as a variable name to avoid overloading the `broker` function
- ~~keep the `Result` from `client_writer` in `writers` (makes the loop with `writer.await?;` work correctly)~~
- ~~return `Result` from `brokers` in the final code example~~

### DIFF
--- a/docs/src/tutorial/clean_shutdown.md
+++ b/docs/src/tutorial/clean_shutdown.md
@@ -25,7 +25,7 @@ async fn server(addr: impl ToSocketAddrs) -> Result<()> {
     let listener = TcpListener::bind(addr).await?;
 
     let (broker_sender, broker_receiver) = mpsc::unbounded();
-    let broker = task::spawn(broker(broker_receiver));
+    let broker_handle = task::spawn(broker(broker_receiver));
     let mut incoming = listener.incoming();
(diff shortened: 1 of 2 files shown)


docs/src/tutorial/clean_shutdown.md
```diff
@@ -60,7 +60,7 @@ async fn broker(mut events: Receiver<Event>) -> Result<()> {
                     Entry::Vacant(entry) => {
                         let (client_sender, client_receiver) = mpsc::unbounded();
                         entry.insert(client_sender);
-                        let handle = spawn_and_log_error(client_writer(client_receiver, stream));
+                        let handle = client_writer(client_receiver, stream);
```
stjepang: I think it'd be best to still spawn the task here, otherwise futures will not execute in parallel.

If we just collect these futures into a `Vec` and then await them at line 72, they will execute serially. But if we spawn them first, they will execute in the background on other threads, and at line 72 we just await these tasks to complete.

zvkemp:
Thanks @stjepang . Looks like one issue was addressed in another PR; I've rebased and reduced this to a single change.

PR #105, opened 2019-08-23
Book: Select wants a FusedFuture for broker() in tutorial

clemensw:
The select! in client_writer() uses fuse(), but the final change to the broker currently doesn't and won't compile.

### DIFF
--- a/docs/src/tutorial/handling_disconnection.md
+++ b/docs/src/tutorial/handling_disconnection.md
@@ -219,11 +219,11 @@ async fn broker(mut events: Receiver<Event>) {
 
     loop {
         let event = select! {
-            event = events.next() => match event {
+            event = events.next().fuse() => match event {
                 None => break, // 2
                 Some(event) => event,
             },
-            disconnect = disconnect_receiver.next() => {
+            disconnect = disconnect_receiver.next().fuse() => {
                 let (name, _pending_messages) = disconnect.unwrap(); // 3
                 assert!(peers.remove(&name).is_some());
                 continue;


marcusklaas:
Ran into this as well. It would be great if the tutorial briefly went into what these fuses are doing and why they are required too!

jamesmunns:
Maybe CC @matklad? I know I've seen you chatting about fused things in the past.

matklad:
cc https://github.com/async-rs/async-std/pull/13 and https://github.com/async-rs/async-std/pull/40 :)

Fuse is definitely a pain point. #13 and #40 apply surgical fixes to it, but the underlying problem that one must call `.fuse` manually remains. It would be ideal if, instead of explaining the current quirks, we could just design a better API that doesn't require `.fuse`. 

Suggested course of action:

* merge #13 as it's a simple fix
* write an RFC/issue for general stream integration and fuseing (cc @stjepang @yoshuawuyts ). My tentative proposal: just make all streams&futeres exposed by async-std fused. It seems adding fuse for futures is much cheaper than for iterators, so it's ok to always do this?

stjepang:
> It would be ideal if, instead of explaining the current quirks, we could just design a better API that doesn't require `.fuse`.

I think the "proper" solution would be to port `crossbeam-channel` to futures and provide a `select` macro that works with async channels.

clemensw:
While improving the story around .fused() is desireable, merging this PR would make the example in the book compile (and be consistent with previous examples), which would help lower confusion for newbies like me.

matklad:
Unfortunately, the fix in this PR is not correct: as per select’s doc, fused should be called outside of the select loop, and this is a pretty bad footgun. Really, the underlying future *is* fused, we should just make sure it is reflected in the type system as in #13.

matklad:
To be more precise, the patch in this PR definitely makes the code *work*, but it doesn’t use select/fuse at they are supposed to be used, and I’d rather not do that in the docs.

clemensw:
Thanks for the explanation.

yoshuawuyts:
#208 and #187 add macro support for non-fused streams and futures respectively. I think this will allow us to bypass a lot of the problems with `futures-rs`'s `select` blocks.

Issue #65, opened 2019-08-17
Make `blocking` public?


zkat:
Hi! I was using `async-std` in my [`cacache`](https://crates.io/crates/cacache) library. One of the things that I'm trying to do is implement `AsyncWrite` for it, but it turns out I'm using a tmpfile library that does sync i/o. Because of that, I pretty much copy-pasted the `AsyncWrite` impl for `async-std`'s `File`, and it turns out with the latest version, `task::blocking` is now private, so I can't just... do that. (To clarify, I was using async-std pre-release, when I needed to use async-pool for this, and I just started porting the code over tonight when I ran into this).

For the sake of compatibility, it would be nice to have this available. My code that's doing this is over here, in case there turns out to be a Better Way™ to do what I'm trying to do that hopefully doesn't involve reimplementing tmpfile logic: https://github.com/zkat/cacache-rs/blob/zkat/async/src/content/write.rs#L147-L256

Cheers!

sdroege:
It would also be quite useful for me to have the `task::blocking` module public. It could be re-implemented externally around a thread pool and oneshot channels of course, but doing that everywhere seems a bit suboptimal (performance- and code duplication-wise).

It's `pub(crate)` since the very first commit in this repo, so the rationale why it is not public anymore is lost from history unfortunately.

I assume it's not public anymore because there was not enough certainty that this is a good enough API to be stabilized as 1.0 soonish? Were there specific concerns about it? To me it looks like exactly how you'd expect such an API to look like.

sdroege:
Maybe (also in relation to https://github.com/async-rs/async-std/issues/60), it would make sense to at least give access to some internals and "uncertain-if-good-enough-for-1.0"-APIs via a feature flag.

That way there can be some experimentation with the API and it isn't required for everybody to duplicate the implementation.

yoshuawuyts:
This has come up several times, and we need to come up with a public API for this. This seems something we'll want to do in the medium term for sure; though we *could* cheat it by introducing the current API as "unstable" first (which seems almost guaranteed like it would change).

@stjepang do you have any thoughts?

dvc94ch:
Can you also expose the `net::driver::Watcher`? I'm currently trying to async signal handling to work properly. The `signal-hooks` crate implements `Evented`, so it would be useful to be able to wrap that in the `Watcher`. The other thing I'm trying to get to work is `flock`. Not sure what the best way to do that is. Either implementing `Evented` or just using `blocking`.

[...]

Issue #79, opened 2019-08-18
More flexible reactor/executor API


sdroege:
While this seems like something that shouldn't be considered for 1.0 IMHO, it would be good to start discussions about how an API could look like and what requirements different folks have here.

Also while this is kind of related to https://github.com/async-rs/async-std/issues/60, my main point here is about being able to have control over the lifetime of the reactor/executor, allowing to run multiple and about which would be used when/where. See also https://github.com/rustasync/runtime/issues/42 for a similar issue of mine for the `runtime` crate, on which everything that follows is based.

----

Currently the executor and reactor and thread pools are all global and lazily started when they're first needed, and there's no way to e.g. start them earlier, stop them at some point, run multiple separate ones, etc.

This simplifies the implementation a lot at this point (extremely clean and easy to follow code right now!) and is also potentially more performant than passing around state via thread-local-storage (like in e.g. `tokio`).

It however limits the usability at least in two scenarios where I'd like to make use of `async-std`.

Anyway, reasons why this would be useful to have (I'm going to call the reactor/executor/threadpool combination a runtime for the following):

1. Usage in library crates without interfering with any other futures code other library crates or the application might use. This would potentially also go with specific per-thread configuration inside the library crate, for e.g. setting thread priorities of the runtime in a way that is meaningful for what this specific library is doing. (See also https://github.com/rustasync/runtime/issues/8)
2. Similar to the above, but an extension with more requirements: plugins. For plugins you might want to use a runtime internally, but at some point you might want to be able to unload the plugin again. As Rust generally does static linking at this point, each plugin would have its own version of `async-std`/etc included, so unloading a plugin also requires to be able to shut down the runtime at a specific point and to ensure that none of the code of the plugin is running anymore.
3. Error isolation. While this is probably done even better with separate processes, being able to compartmentalize the application into different parts that don't implicitly share any memory with each other could be useful, also for debuggability.

vertexclique:
Truly, I am into this. I've mentioned this implicitly here:

https://github.com/async-rs/async-std/issues/14#issuecomment-521589882

> ... We should share a part of the thread pool explicitly for polling purposes to our own task mechanisms' management. If we don't do that writing abstractions will be cumbersome over time.

stjepang:
This is a feature we should and will implement, but we'll need to explore the design space first. I think the biggest blocker right now is improving our scheduler. Once the scheduler matures, it should become more obvious how to start/stop runtimes manually instead of always relying on a single static one.

[...]

PR #108, opened 2019-08-24
EMA based statistically adaptive thread pool design

vertexclique:
This is an adaptive thread pool implementation with exponential moving averages(where it doesn't rely on idle time).

Here are the benchmarks:

**This new statistical-EMA-approach:**
- CPU utilization ~97%

```
running 1 test
test blocking ... bench:  39,116,408 ns/iter (+/- 5,686,213)
```

**Current approach**
- CPU utilization ~88%

```
running 1 test
test blocking ... bench: 254,204,382 ns/iter (+/- 214,099,973)
```

This approach enables 5.5x speedup.

On top of that, this PR solves #65 intrinsically 
Just because it is not nice to manually include modules at the benchmark level.

## Documentation
blocking doc comment

### DIFF
--- a/benches/blocking.rs
+++ b/benches/blocking.rs
@@ -9,6 +9,7 @@ use std::thread;
 use std::time::Duration;
 use test::Bencher;
 
+// Benchmark for a 10K burst task spawn
 #[bench]
 fn blocking(b: &mut Bencher) {
     b.iter(|| {
@@ -22,5 +23,16 @@ fn blocking(b: &mut Bencher) {
             .collect::<Vec<JoinHandle<()>>>();
 
         task::block_on(join_all(handles));
+    });
+}
+
+// Benchmark for a single blocking task spawn
+#[bench]
+fn blocking_single(b: &mut Bencher) {
+    b.iter(|| {
+        task::blocking::spawn(async {
+            let duration = Duration::from_millis(1);
+            thread::sleep(duration);
+        })
     });
 }
--- a/benches/blocking.rs
+++ b/benches/blocking.rs
@@ -1,0 +1,26 @@
+#![feature(test)]
+
+extern crate test;
+
+use async_std::task;
+use async_std::task_local;
+use test::{black_box, Bencher};
+use std::thread;
+use std::time::Duration;
+use async_std::task::blocking::JoinHandle;
+use futures::future::{join_all};
+
+
+#[bench]
+fn blocking(b: &mut Bencher) {
+    b.iter(|| {
+        let handles = (0..10_000).map(|_| {
+            task::blocking::spawn(async {
+                let duration = Duration::from_millis(1);
+                thread::sleep(duration);
+            })
+        }).collect::<Vec<JoinHandle<()>>>();
+
+        task::block_on(join_all(handles));
+    });
+}
--- a/benches/blocking.rs
+++ b/benches/blocking.rs
@@ -3,23 +3,23 @@
 extern crate test;
 
 use async_std::task;
-use async_std::task_local;
-use test::{black_box, Bencher};
+use async_std::task::blocking::JoinHandle;
+use futures::future::join_all;
 use std::thread;
 use std::time::Duration;
-use async_std::task::blocking::JoinHandle;
-use futures::future::{join_all};
-
+use test::Bencher;
 
 #[bench]
 fn blocking(b: &mut Bencher) {
     b.iter(|| {
-        let handles = (0..10_000).map(|_| {
-            task::blocking::spawn(async {
-                let duration = Duration::from_millis(1);
-                thread::sleep(duration);
+        let handles = (0..10_000)
+            .map(|_| {
+                task::blocking::spawn(async {
+                    let duration = Duration::from_millis(1);
+                    thread::sleep(duration);
+                })
             })
-        }).collect::<Vec<JoinHandle<()>>>();
+            .collect::<Vec<JoinHandle<()>>>();
 
         task::block_on(join_all(handles));
     });
--- a/src/task/blocking.rs
+++ b/src/task/blocking.rs
@@ -55,6 +55,8 @@ lazy_static! {
                 .expect("cannot start a thread driving blocking tasks");
         }
 
+        // Pool manager to check frequency of task rates
+        // and take action by scaling the pool accordingly.
         thread::Builder::new()
             .name("async-pool-manager".to_string())
             .spawn(|| abort_on_panic(|| {
--- a/src/task/blocking.rs
+++ b/src/task/blocking.rs
@@ -96,9 +96,13 @@ lazy_static! {
 // This value represents the trend of tasks mapped onto the thread pool.
 // Calculation is following:
 //
-// α    :: EMA_COEFFICIENT :: smoothing factor between 0 and 1
-// Yt   :: freq            :: frequency sample at time t
-// St   :: acc             :: EMA at time t
+// +--------+-----------------+----------------------------------+
+// | Symbol |   Identifier    |           Explanation            |
+// +--------+-----------------+----------------------------------+
+// | α      | EMA_COEFFICIENT | smoothing factor between 0 and 1 |
+// | Yt     | freq            | frequency sample at time t       |
+// | St     | acc             | EMA at time t                    |
+// +--------+-----------------+----------------------------------+
 //
 // Under these definitions formula is following:
 // EMA = α * [ Yt + (1 - α)*Yt-1 + ((1 - α)^2)*Yt-2 + ((1 - α)^3)*Yt-3 ... ] + St
@@ -163,7 +167,7 @@ fn scale_pool() {
     {
         // Throughput is low. Allocate more threads to unblock flow.
         // If we fall to this case, scheduler is congested by longhauling tasks.
-        // For unblock the flow we should add up some threads to the pool, but not that much to
+        // For unblock the flow we should add up some threads to the pool, but not that many to
         // stagger the program's operation.
         let scale = LOW_WATERMARK * current_frequency + 1;
 
@@ -221,8 +225,8 @@ fn create_blocking_thread() {
             match err.kind() {
                 ErrorKind::WouldBlock => {
                     // Maximum allowed threads per process is varying from system to system.
-                    // Some systems has it(like MacOS), some doesn't(Linux)
-                    // This case expected to not happen.
+                    // Also, some systems have it(like macOS), and some don't(Linux).
+                    // This case expected not to happen.
                     // But when happened this shouldn't throw a panic.
                     let current_arc = POOL_SIZE.clone();
                     MAX_THREADS.store(*current_arc.lock().unwrap() - 1, Ordering::SeqCst);
@@ -235,7 +239,7 @@ fn create_blocking_thread() {
         });
 }
 
-// Enqueues work, attempting to send to the threadpool in a
+// Enqueues work, attempting to send to the thread pool in a
 // nonblocking way and spinning up needed amount of threads
 // based on the previous statistics without relying on
 // if there is not a thread ready to accept the work or not.
--- a/src/task/blocking.rs
+++ b/src/task/blocking.rs
@@ -21,14 +21,15 @@ const LOW_WATERMARK: u64 = 2;
 
 // Pool managers interval time (milliseconds)
 // This is the actual interval which makes adaptation calculation
-const MANAGER_POLL_INTERVAL: u64 = 50;
-
-// Frequency scale factor for thread adaptation calculation
-const FREQUENCY_SCALE_FACTOR: u64 = 200;
+const MANAGER_POLL_INTERVAL: u64 = 200;
 
 // Frequency histogram's sliding window size
 // Defines how many frequencies will be considered for adaptation.
 const FREQUENCY_QUEUE_SIZE: usize = 10;
+
+// Exponential moving average smoothing coefficient for limited window
+// Smoothing factor is estimated with: 2 / (N + 1) where N is sample size.
+const EMA_COEFFICIENT: f64 = 2_f64 / (FREQUENCY_QUEUE_SIZE as f64 + 1_f64);
 
 // Possible max threads (without OS contract)
 const MAX_THREADS: u64 = 10_000;
@@ -96,6 +97,12 @@ fn get_current_pool_size() -> u64 {
     LOW_WATERMARK.max(current_pool_size)
 }
 
+fn calculate_ema(freq_queue: &VecDeque<u64>) -> f64 {
+    freq_queue.iter().enumerate().fold(0_f64, |acc, (i, freq)| {
+        acc + ((*freq as f64) * ((1_f64 - EMA_COEFFICIENT).powf(i as f64) as f64))
+    }) * EMA_COEFFICIENT as f64
+}
+
 // Adaptive pool scaling function
 //
 // This allows to spawn new threads to make room for incoming task pressure.
@@ -114,39 +121,41 @@ fn scale_pool() {
         freq_queue.push_back(0);
     }
 
-    // Calculate message rate at the given time window
-    // Factor is there for making adaptation linear.
-    // Exponential bursts may create idle threads which won't be utilized.
-    // They may actually cause slowdown.
-    let current_freq_factorized = current_frequency as f64 * FREQUENCY_SCALE_FACTOR as f64;
-    let frequency = (current_freq_factorized / MANAGER_POLL_INTERVAL as f64) as u64;
+    // Calculate message rate for the given time window
+    let frequency = (current_frequency as f64 / MANAGER_POLL_INTERVAL as f64) as u64;
 
     // Adapts the thread count of pool
     //
     // Sliding window of frequencies visited by the pool manager.
     // Select the maximum from the window and check against the current task dispatch frequency.
     // If current frequency is bigger, we will scale up.
-    if let Some(&max_measurement) = freq_queue.iter().max() {
-        if frequency > max_measurement {
-            // Don't spawn more than cores.
-            // Default behaviour of most of the linear adapting thread pools.
-            let scale_by = num_cpus::get().max(LOW_WATERMARK as usize) as u64;
-
-            // Pool size can't reach to max_threads anyway.
-            // Pool manager backpressures itself while visiting message rate frequencies.
-            // You will get an error before hitting to limits by OS.
-            if get_current_pool_size() < MAX_THREADS {
-                (0..scale_by).for_each(|_| {
-                    create_blocking_thread();
-                });
-            }
-        }
-    }
+    let prev_ema_frequency = calculate_ema(&freq_queue);
 
     // Add seen frequency data to the frequency histogram.
     freq_queue.push_back(frequency);
     if freq_queue.len() == FREQUENCY_QUEUE_SIZE + 1 {
         freq_queue.pop_front();
+    }
+
+    let curr_ema_frequency = calculate_ema(&freq_queue);
+
+    if curr_ema_frequency > prev_ema_frequency {
+        let scale_by: f64 = curr_ema_frequency - prev_ema_frequency;
+        let scale = ((LOW_WATERMARK as f64 * scale_by) + LOW_WATERMARK as f64) as u64;
+
+        // Pool size shouldn't reach to max_threads anyway.
+        // Pool manager backpressures itself while visiting message rate frequencies.
+        // You will get an error before hitting to limits by OS.
+        (0..scale).for_each(|_| {
+            create_blocking_thread();
+        });
+    } else if curr_ema_frequency == prev_ema_frequency && current_frequency != 0 {
+        // Throughput is low. Allocate more threads to unblock flow.
+        let scale = LOW_WATERMARK * current_frequency + 1;
+
+        (0..scale).for_each(|_| {
+            create_blocking_thread();
+        });
     }
 
     FREQUENCY.store(0, Ordering::Release);
--- a/src/task/blocking.rs
+++ b/src/task/blocking.rs
@@ -11,6 +11,7 @@ use crossbeam_channel::{bounded, Receiver, Sender};
 use lazy_static::lazy_static;
 
 use crate::future::Future;
+use crate::io::ErrorKind;
 use crate::task::{Context, Poll};
 use crate::utils::abort_on_panic;
 use std::sync::{Arc, Mutex};
@@ -31,12 +32,12 @@ const FREQUENCY_QUEUE_SIZE: usize = 10;
 // Smoothing factor is estimated with: 2 / (N + 1) where N is sample size.
 const EMA_COEFFICIENT: f64 = 2_f64 / (FREQUENCY_QUEUE_SIZE as f64 + 1_f64);
 
-// Possible max threads (without OS contract)
-const MAX_THREADS: u64 = 10_000;
-
 // Pool task frequency variable
 // Holds scheduled tasks onto the thread pool for the calculation window
 static FREQUENCY: AtomicU64 = AtomicU64::new(0);
+
+// Possible max threads (without OS contract)
+static MAX_THREADS: AtomicU64 = AtomicU64::new(10_000);
 
 struct Pool {
     sender: Sender<async_task::Task<()>>,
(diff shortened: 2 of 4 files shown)


yoshuawuyts:
I'm in no means qualified to review this code, but the results are looking very cool! awesome work @vertexclique! :tada:

src/task/blocking.rs
```diff
@@ -73,25 +96,49 @@ fn maybe_create_another_blocking_thread() {
...
+    FREQUENCY.fetch_add(1, Ordering::Relaxed);
+
+    // Calculate the amount of threads needed to spin up
+    // then retry sending while blocking. It doesn't spin if
+    // expected pool size is above the MAX_THREADS (which is a
+    // case won't happen)
+    let pool_size = EXPECTED_POOL_SIZE.load(Ordering::Relaxed);
+    let current_pool_size = CURRENT_POOL_SIZE.load(Ordering::Relaxed);
+    if pool_size > current_pool_size && pool_size <= MAX_THREADS {
+        let needed = pool_size - current_pool_size;
```
spacejam: maybe a saturating subtraction for defense in depth?

vertexclique: Was doing that rn. Just in place comment

src/task/blocking.rs
```diff
@@ -47,18 +58,30 @@ lazy_static! {
...
-    // even if it's random.
-    let workers = DYNAMIC_THREAD_COUNT.load(Ordering::Relaxed);
-    if workers >= MAX_THREADS {
-        return;
+fn calculate_dispatch_frequency() {
+    // Calculate current message processing rate here
+    let previous_freq = FREQUENCY.fetch_sub(1, Ordering::Relaxed);
+    let avr_freq = AVR_FREQUENCY.load(Ordering::Relaxed);
+    let current_pool_size = CURRENT_POOL_SIZE.load(Ordering::Relaxed);
+    let frequency = (avr_freq as f64 + previous_freq as f64 / current_pool_size as f64) as u64;
```
spacejam: Should the avr and previous freq's be collectively divided by the current pool size? It's not totally clear to me how this avoids creating a feedback loop.

vertexclique: Actually, that holds the current one. I made it more clear.

vertexclique: That's how moving average works in general. It won't create a feedback loop since finished threads will decrease the needed count and that saturating thread join will join them over time. It won't spawn new threads, because current pool size is controlling thread creation.

src/task/blocking.rs
```diff
@@ -73,25 +100,53 @@ fn maybe_create_another_blocking_thread() {
...
             let wait_limit = Duration::from_millis(1000 + rand_sleep_ms);
 
-            DYNAMIC_THREAD_COUNT.fetch_add(1, Ordering::Relaxed);
+            CURRENT_POOL_SIZE.fetch_add(1, Ordering::Relaxed);
             while let Ok(task) = POOL.receiver.recv_timeout(wait_limit) {
                 abort_on_panic(|| task.run());
+                calculate_dispatch_frequency();
             }
-            DYNAMIC_THREAD_COUNT.fetch_sub(1, Ordering::Relaxed);
+            CURRENT_POOL_SIZE.fetch_sub(1, Ordering::Relaxed);
```
spacejam: This can't be a Relaxed atomic if we want to guarantee a happens-before relationship with the addition, which is important to guarantee that we never divide by zero in the ema calculation above

spacejam: (can also just add 1 to the divisor in the calculation above to enforce this invariant at the dependency site)

spacejam: but atomics will wrap, so we probably want to use a proper barrier here anyway for defense against wrap-intolerance in other calculation areas

spacejam: Ordering::Release for writes generally and Ordering::Acquire for all loads I think is OK on this code, but honestly due to the timescales involved in blocking codepaths it's fine to use a SeqCst to think even less about wrap issues.

vertexclique: I am going to use SeqCst in both.

spacejam:
Thanks for continuing to boil this down! These are really nice results that address the main runaway thread growth issue in the naive original implementation.

I'm working through some more correctness proofs on paper to try to find more ways to break the large number of Relaxed atomic accesses that are added in this.

vertexclique:
Thanks, Tyler. Very nice to hear that. If you are doing FV please share with me too.

spacejam:
Hehehe the reason I'm using paper for a couple things is because I'm a little uncertain about translating Relaxed stuff into TLA (maybe using sets of all reorderable code and permuting them?)

Basically, anywhere a completely random value impacts runtime correctness is the only thing I'm looking for and desk-checking backwards from.

vertexclique:
More to improve (aka TO-DO):
* Adaptive scale down (we can do this sometime later, it is harder and need more sync points than above)
* Compartmentalize the pool creation into so-called `execution context`'s so they satisfy the needs in the issue #79 . This PR also helps that in an indirect way. (Atomics can be moved for that mentioned issue, no worries)

spacejam:
If more synchronization needs to occur, it probably isn't a bad time to take a step back from the 100% atomics-based approach and either use ROWEX when an adaptation is required (spawning readers work lock-free, writers need to claim an exclusive mutex) or just using a Mutex so that the implementation is obviously correct. Thinking about all of these atomic accesses is quite labor intensive and error-prone.

vertexclique:
I think todos can go to other PRs. For Mutexes (if I understand you correctly you are talking about the scaling down part) it can be useful, but I don't think they should be used for this code since there is no critical section code here, I don't like abusing lock-exclusion in places which are not needed. (But I am not fully against that).

I am rooting for this PR to take it like this for blocking pool like how it is rn. Then we will think on that later in the process of extraction (which will also be for async part, I think).

spacejam:
I guess what I'm mostly getting at is that there are some potentially very subtle information flows that can happen when this math happens to be operating in an entirely concurrent way on global variables via Relaxed atomics (or even SeqCst atomics), and this is expensive overall to reason about as a human (it's also expensive/unreasonable for the formal verification tooling I know about to provide meaningful validations of).

spacejam:
There is almost never a real need for exclusion, but it makes PR's like this understandable without staring at it for hours :P

src/task/blocking.rs
```diff
@@ -47,18 +58,34 @@ lazy_static! {
...
-    // it's just a heuristic, and would not lose correctness
-    // even if it's random.
-    let workers = DYNAMIC_THREAD_COUNT.load(Ordering::Relaxed);
-    if workers >= MAX_THREADS {
-        return;
+fn calculate_dispatch_frequency() {
+    // Calculate current message processing rate here
+    let current_freq = FREQUENCY.fetch_sub(1, Ordering::Relaxed);
+    let avr_freq = AVR_FREQUENCY.load(Ordering::Relaxed);
+    let current_pool_size = CURRENT_POOL_SIZE.load(Ordering::Relaxed);
```
spacejam: For defense in depth and Relaxed raciness protection:
`std::cmp::max(LOW_WATER_MARK, CURRENT_POOL_SIZE.load(Ordering::Relaxed)`

src/task/blocking.rs
```diff
@@ -47,18 +58,34 @@ lazy_static! {
...
-fn maybe_create_another_blocking_thread() {
-    // We use a `Relaxed` atomic operation because
-    // it's just a heuristic, and would not lose correctness
-    // even if it's random.
-    let workers = DYNAMIC_THREAD_COUNT.load(Ordering::Relaxed);
-    if workers >= MAX_THREADS {
-        return;
+fn calculate_dispatch_frequency() {
+    // Calculate current message processing rate here
+    let current_freq = FREQUENCY.fetch_sub(1, Ordering::Relaxed);
```
spacejam: This only happens to be properly serialized due to the SeqCst barrier placed on the CURRENT_POOL_SIZE load in schedule(), but if that changes this is vulnerable to an underwrap. Please make it at least `Ordering::Acquire` for defense.

src/task/blocking.rs
```diff
@@ -73,25 +100,53 @@ fn maybe_create_another_blocking_thread() {
...
+// based on the previous statistics without relying on
+// if there is not a thread ready to accept the work or not.
 fn schedule(t: async_task::Task<()>) {
+    // Add up for every incoming task schedule
+    FREQUENCY.fetch_add(1, Ordering::Relaxed);
+
+    // Calculate the amount of threads needed to spin up
+    // then retry sending while blocking. It doesn't spin if
+    // expected pool size is above the MAX_THREADS (which is a
+    // case won't happen)
```
spacejam: This case comes really close to being able to happen though! Without the SeqCst barrier on CURRENT_POOL_SIZE below, AVR_FREQUENCY and FREQUENCY would be able to underwrap in calculate_dispatch_frequency. To make it really clear that this won't happen, proper barriers should be set on the atomic accesses on FREQUENCY + AVR_FREQUENCY as well.

src/task/blocking.rs
```diff
@@ -73,25 +100,53 @@ fn maybe_create_another_blocking_thread() {
...
 
 // Enqueues work, attempting to send to the threadpool in a
-// nonblocking way and spinning up another worker thread if
-// there is not a thread ready to accept the work.
+// nonblocking way and spinning up needed amount of threads
+// based on the previous statistics without relying on
+// if there is not a thread ready to accept the work or not.
 fn schedule(t: async_task::Task<()>) {
+    // Add up for every incoming task schedule
+    FREQUENCY.fetch_add(1, Ordering::Relaxed);
```
spacejam: Please use an `Ordering::Acquire` or stronger here

src/task/blocking.rs
```diff
@@ -73,25 +100,53 @@ fn maybe_create_another_blocking_thread() {
...
+
+    // Calculate the amount of threads needed to spin up
+    // then retry sending while blocking. It doesn't spin if
+    // expected pool size is above the MAX_THREADS (which is a
+    // case won't happen)
+    let pool_size = EXPECTED_POOL_SIZE.load(Ordering::Relaxed);
+    let current_pool_size = CURRENT_POOL_SIZE.load(Ordering::SeqCst);
+    let reward = (AVR_FREQUENCY.load(Ordering::Relaxed) as f64 / 2.0_f64) as u64;
```
spacejam: `Ordering::Acquire`

spacejam:
In general, please change all of the atomics to use barriers that prevent potentially risky reorderings. Note that in the original, relaxed atomics were used with a clear explanation for why randomness in their use path violated no correctness criteria. In this code, it's much much much more complex, and relaxed atomics make this quite difficult to reason about. 

Mutexes will make this very easy for me to review and merge, and will not impact performance, so I recommend that you replace all of the atomic stuff with them, but I totally get that it's not as fun!

spacejam:
After removing all Relaxed atomics, I have some concerns about long-term maintenance costs of this code due to the large amount of unserialized atomic access, and math without clear documentation. Please document the math more clearly for people like me who are less familiar with it, and please document the atomic access patterns with explanations of what the important correctness criteria are at each dependency site so that future contributors don't introduce bugs. 

Since bugs have been found in review, it would also make me feel better if some concurrency test was added that exercises the interleavings more.

You may think that this is a bit draconian, but lock-free programming is not easy for humans to get right, and this code is not obviously correct to me yet. 

This needs a bit more work before it's safe from a long-term maintenance perspective. The alternative is to wrap all atomic variables other than FREQUENCY in a Mutex so that their complex math is serialized. Please try this out if you're uninterested in adding concurrency tests. Mutex unlocks are less than 16ns, and as long as you're just doing this math on variables inside the Mutex and not doing any IO or running a blocking function with the Mutex held, it will be really really fast.

So, please make this code obviously correct by either using a simple Mutex around all of the variables other than FREQUENCY or by adding concurrent interleaving tests that will cause assumptions we're making to pop out.

benches/blocking.rs
```diff
@@ -0,0 +1,26 @@
...
+use async_std::task::blocking::JoinHandle;
+use futures::future::join_all;
+use std::thread;
+use std::time::Duration;
+use test::Bencher;
+
+#[bench]
+fn blocking(b: &mut Bencher) {
+    b.iter(|| {
+        let handles = (0..10_000)
```
kpp: That's great, how about swapning >MAX_THREADS with thread::sleep(10sec)? Will your patch change the behavior?

src/task/blocking.rs
```diff
@@ -12,10 +13,29 @@ use lazy_static::lazy_static;
 use crate::future::Future;
 use crate::task::{Context, Poll};
 use crate::utils::abort_on_panic;
+use std::sync::{Arc, Mutex};
 
+// Low watermark value, defines the bare minimum of the pool.
+// Spawns initial thread set.
```
killercup: even though the items aren't public these comments could all be doc comments

src/task/blocking.rs
```diff
@@ -45,20 +78,84 @@ lazy_static! {
...
+}
+
+// Adaptive pool scaling function
+//
+// This allows to spawn new threads to make room for incoming task pressure.
+// Works in the background detached from the pool system and scales up the pool based
+// on the request rate.
+//
+// It uses frequency based calculation to define work. Utilizing average processing rate.
+fn scale_pool() {
```
killercup: In another context there was some discussion around adding tracing (using log's `trace!` macro) to async-std. This functions seems like a good place for that, too.

killercup: I just know that the idea was thrown around, at it was probably an opt-in feature, too. @stjepang or @yoshuawuyts would know more :)

vertexclique:
So, to sum up, all the changes that I pushed:
* Values moved around are mutex rn.
* Now I cleanly calculate the values and also documented more explicitly how things are working. Even formula, very explicitly.
* Fixed #126 (OS-specific problem, hard to hit but possible)
* No significant degradation in performance. For the curious:
```
running 2 tests
test blocking        ... bench:  48,959,842 ns/iter (+/- 46,653,507)
test blocking_single ... bench:       3,322 ns/iter (+/- 8,452)
```
* I think I covered most of the cases with tests. Skipped all of them since they are exhaustive. But for the curious cats again here how you can run them:
```
$ cargo test longhauling_task_join -- --ignored --exact --nocapture
$ cargo test slow_join_interrupted -- --ignored --exact --nocapture
$ cargo test slow_join -- --ignored --exact --nocapture
```
* **NOTE**: Benchmarks cover the common cases. Tests are covering the possible edge cases.

src/task/blocking.rs
```diff
@@ -1,5 +1,32 @@
...
+//!
+//! Algorithm is altered and adapted from [A Novel Predictive and Self–Adaptive Dynamic Thread Pool Management](https://doi.org/10.1109/ISPA.2011.61).
+//!
+//! ## Predictive Upscaler
+//! Selects upscaling amount based on estimation or when throughput hogs based on amount of tasks mapped.
+//! Throughput hogs determined by a combination of job in / job out frequency and current scheduler task assignment frequency.
+//!
+//! ## Time-based Downscaler
+//! After dynamic tasks spawned with upscaler they will continue working in between 1 second and 10 seconds.
+//! When tasks are detached from the channels after this amount they join back.
```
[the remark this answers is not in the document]
vertexclique: You got totally right, my mistake of documenting. Will reword this.

src/task/blocking.rs
```diff
@@ -45,20 +133,125 @@ lazy_static! {
...
         Pool { sender, receiver }
     };
+
+    /// Sliding window for pool task frequency calculation
+    static ref FREQ_QUEUE: Arc<Mutex<VecDeque<u64>>> = {
+        Arc::new(Mutex::new(VecDeque::with_capacity(FREQUENCY_QUEUE_SIZE + 1)))
+    };
+
+    /// Dynamic pool thread count variable
+    static ref POOL_SIZE: Arc<Mutex<u64>> = Arc::new(Mutex::new(LOW_WATERMARK));
```
spacejam: All use of arcs in this module is unnecessary, please remove them all.

src/task/blocking.rs
```diff
@@ -45,20 +133,125 @@ lazy_static! {
...
+///
+/// This allows to spawn new threads to make room for incoming task pressure.
+/// Works in the background detached from the pool system and scales up the pool based
+/// on the request rate.
+///
+/// It uses frequency based calculation to define work. Utilizing average processing rate.
+fn scale_pool() {
+    // Fetch current frequency, it does matter that operations are ordered in this approach.
+    let current_frequency = FREQUENCY.load(Ordering::SeqCst);
+    let freq_queue_arc = FREQ_QUEUE.clone();
```
spacejam: there is no need to clone arcs unless you need another reference to the inner member. since there is a unique reference in the lazy_static, no arc clones, or arcs at all, are necessary.

src/task/blocking.rs
```diff
@@ -45,20 +133,125 @@ lazy_static! {
...
+}
+
+/// Creates blocking thread to receive tasks
+/// Dynamic threads will terminate themselves if they don't
+/// receive any work after between one and ten seconds.
+fn create_blocking_thread() {
+    // Check that thread is spawnable.
+    // If it hits to the OS limits don't spawn it.
+    {
+        let current_arc = POOL_SIZE.clone();
```
spacejam: no need to clone or to use an arc at all

src/task/blocking.rs
```diff
@@ -68,29 +261,53 @@ fn maybe_create_another_blocking_thread() {
...
         .name("async-blocking-driver-dynamic".to_string())
         .spawn(move || {
             let wait_limit = Duration::from_millis(1000 + rand_sleep_ms);
 
-            DYNAMIC_THREAD_COUNT.fetch_add(1, Ordering::Relaxed);
+            // Adjust the pool size counter before and after spawn
+            {
+                let current_arc = POOL_SIZE.clone();
+                *current_arc.lock().unwrap() += 1;
+            }
```
spacejam: cloning this arc does nothing, please remove the arc, its clone, and the subscope

src/task/blocking.rs
```diff
@@ -68,29 +261,53 @@ fn maybe_create_another_blocking_thread() {
...
             while let Ok(task) = POOL.receiver.recv_timeout(wait_limit) {
                 abort_on_panic(|| task.run());
             }
-            DYNAMIC_THREAD_COUNT.fetch_sub(1, Ordering::Relaxed);
+            {
+                let current_arc = POOL_SIZE.clone();
+                *current_arc.lock().unwrap() -= 1;
+            }
```
spacejam: cloning this arc does nothing, please remove the arc, its clone, and the subscope

src/task/blocking.rs
```diff
@@ -68,29 +261,53 @@ fn maybe_create_another_blocking_thread() {
...
-        .expect("cannot start a dynamic thread driving blocking tasks");
+        .map_err(|err| {
+            match err.kind() {
+                ErrorKind::WouldBlock => {
+                    // Maximum allowed threads per process is varying from system to system.
+                    // Also, some systems have it(like macOS), and some don't(Linux).
+                    // This case expected not to happen.
+                    // But when happened this shouldn't throw a panic.
+                    let current_arc = POOL_SIZE.clone();
+                    MAX_THREADS.store(*current_arc.lock().unwrap() - 1, Ordering::SeqCst);
```
spacejam: cloning this arc does nothing, please remove the arc, its clone, and the subscope

tests/thread_pool.rs
```diff
@@ -0,0 +1,141 @@
...
+    let elapsed = start.elapsed().as_millis() - thread_join_time_max as u128;
+    println!(
+        "Long-hauling task join. Monotonic exec time: {:?} ns",
+        elapsed
+    );
+
+    // Should be less than 200_000 ns
+    // Previous implementation will panic when this test is running.
+    assert_eq!(elapsed < 200_000, true);
+}
```
spacejam: these tests are generally non-reproducible, and it's unclear if they are valuable. they tend to fail on my fairly modern 2018 i7 mbp

spacejam: when I said concurrency tests before, I meant adding randomized delays to operations that involve cross-thread coordination via atomics (or higher level primitives that invoke them internally). These might be OK as a benchmark so you can compare results on your own machine before and after changes, but I don't think they are useful as tests.

src/task/blocking.rs
```diff
@@ -45,20 +133,125 @@ lazy_static! {
...
+/// Adaptive pool scaling function
+///
+/// This allows to spawn new threads to make room for incoming task pressure.
+/// Works in the background detached from the pool system and scales up the pool based
+/// on the request rate.
+///
+/// It uses frequency based calculation to define work. Utilizing average processing rate.
+fn scale_pool() {
+    // Fetch current frequency, it does matter that operations are ordered in this approach.
+    let current_frequency = FREQUENCY.load(Ordering::SeqCst);
```
spacejam: It feels kind of weird that you're loading here and later storing 0, instead of swapping 0 in here and not ignoring items in the interval. would that be more correct?

src/task/blocking.rs
```diff
@@ -10,21 +62,44 @@ use crossbeam_channel::{bounded, Receiver, Sender};
 use lazy_static::lazy_static;
 
 use crate::future::Future;
+use crate::io::ErrorKind;
 use crate::task::{Context, Poll};
 use crate::utils::abort_on_panic;
+use std::sync::{Arc, Mutex};
```
spacejam: `Arc` unnecessary

tests/thread_pool.rs
vertexclique: Please try running the tests with master's `blocking.rs`. They will panic.

src/task/blocking.rs
vertexclique: People tend to share this between threads. It is unnecessary because pool manager doesn't use it. If someone starts to share this across threads we will see the use of it.

src/task/blocking.rs
yoshuawuyts: `log` now has kv support, which we're already using in other places too. `tracing` can pick these up too, so using ` log` is probably the best choice here overall (:

davidbarsky: I think we should get the `kv` support in `tracing-log` soon-ish? However I don't believe that `log` supports the `span`-like macros, so async-std might need to have its own, internal `span`-like macro and dispatch conditionally to async-log or tracing.

 I haven't considered how the differences between [async-log's `span`](https://docs.rs/async-log/2.0.0/async_log/macro.span.html) and [`tracing`'s span](https://docs.rs/tracing/0.1.7/tracing/macro.span.html) can be bridged, but I think it's feasible if `tracing'`s `span!().enter()` is used to mimic async-log's nested `span` block.

yoshuawuyts: @davidbarsky we could use [`async-log::span`](https://docs.rs/async-log/2.0.0/async_log/macro.span.html) for this, probably. Though there's still a few open issues on the repo, the interface wouldn't change.

davidbarsky: @yoshuawuyts Yep! I think we're on the same page about wanting to bridge the two libraries. 

I'm spitballing _one_ mechanism on bridging that gap, which consists of using mutually exclusive feature flags in a `build.rs` to dispatch to either `async-log` or `tracing`. This would entail using an `async-std`-specific `span` macro until the gaps—if any exist—between `async-std`'s `span` and `tracing`'s `span` are closed.

I'm sorry if what I communicated wasn't clear!

tests/thread_pool.rs
spacejam: They panic on this branch on a fairly modern machine, so I'm not sure they are useful on machines other than yours. In general, timing-based performance tests are an antipattern, as they have big reproducibility issues. If you want to do this kind of testing, it's better to use something like cachegrind to count CPU instructions executed.

src/task/blocking.rs
```diff
@@ -45,52 +133,175 @@ lazy_static! {
...
+/// +--------+-----------------+----------------------------------+
+/// ```
+/// Under these definitions formula is following:
+/// ```text
+/// EMA = α * [ Yt + (1 - α)*Yt-1 + ((1 - α)^2)*Yt-2 + ((1 - α)^3)*Yt-3 ... ] + St
+/// ```
+/// # Arguments
+///
+/// * `freq_queue` - Sliding window of frequency samples
+#[inline]
```
spacejam: in general, humans are bad at guessing whether something should be inlined. Usually in rust it's better to leave these attributes out unless come code is going to be executed in a super hot path and you've done work to ensure the improvement is measurable

tests/thread_pool.rs
spacejam: Because they don't actually assert on anything now (which was what caused them to fail on my laptop), I think they are better as benchmarks, right? I don't think it makes sense to have tests that can't fail.

src/task/blocking.rs
vertexclique: This comment is not contributing to the overall discussion + review. I’ve found it quite a bit `out of context` and also from the `book`. I've made the benchmarks already with and without and decided to leave it inlined. You can see below my last run again after your comment (run shouldn't be needed because of how compiler inlines and unrolls):

Mean of 10 consecutive runs of `blocking` benchmark:
```
43057207.9 ns without inlining
41816626.1 ns with inlining
```

That is also better for users of this library indirectly.

tests/thread_pool.rs
vertexclique: > I don't think it makes sense to have tests that can't fail.

Please read my comments carefully. You can see what it tests in action:
https://asciinema.org/a/R6b3UcFo0xfNo1sCPOPJrB04T

> when I said concurrency tests before, I meant adding randomized delays to operations that involve cross-thread coordination via atomics (or higher level primitives that invoke them internally).

Unexposed architecture shouldn't be tested with whitebox fashion.

spacejam: Where are the assertions that fail? The asciicinema is too long to watch. 

Possible concurrent interleavings are exposed architecture, and this kind of test has resulted in a huge number of bugs being discovered in various lock-free projects I've been involved in. Concurrency is really hard to get right sometimes, and the original PR with more unserialized atomic accesses would have had bugs jump out pretty quickly with this approach.

src/task/blocking.rs
spacejam: I'm a bit skeptical that change is related to the inline here because this is a function that gets called 5 times per second, and that measurement skew is 2 milliseconds.

vertexclique:
Please track the progress on https://github.com/async-rs/async-std/pull/181

PR #117, opened 2019-08-27
Update Future description

### LINKED ISSUE #103: Book chapter 1.2 futures::future::Future description
First, thanks a lot for this great library and it's accompanying documentation!

This description of `std::future::Future` form the book sounds not quite correct: 

> In some sense, the std::future::Future can be seen as a minimal subset of futures::future::Future

https://book.async.rs/overview/std-and-library-futures.html

Actually both traits are the same. It's just a reexport:

https://github.com/rust-lang-nursery/futures-rs/blob/cde791c00b8b9c4fd14a594855038a1bc4b6323e/futures-core/src/future/mod.rs#L7

JohnTitor:
Fixes #103


stjepang:
I kind of feel it would be best to just delete that sentence :) wdyt, @skade?

JohnTitor:
Hmm, what I should do? Should improve the sentence or remove it?

stjepang:
Hmm, perhaps we should do another pass over the whole section? It seems to imply a few things that are potentially confusing or misleading:

1. It says `Future` is a type, but it really is a trait.
2. It talks about differences between `futures::future::Future` and `std::future::Future`, but they are the same trait.

yoshuawuyts:
ping @skade

joodicator:
While reading https://book.async.rs/overview/std-and-library-futures.html, which at time of writing currently still incorrectly states that `std::future::Future` and `futures::future::Future` are distinct, I was confused, and it took me about half an hour to establish what was actually going on. It would be good to have this changed, since, as I can attest, the misinformation is having real-world consequences. Aside from that, I appreciate the information about asynchronous Rust.

joshtriplett:
https://github.com/async-rs/async-std/pull/1024

joshtriplett:
Fixed by https://github.com/async-rs/async-std/pull/1024

PR #118, opened 2019-08-27
Pretty Travis cfg

kpp:
See https://travis-ci.com/async-rs/async-std/builds/124829949 for nicely separated build jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,44 @@
 language: rust
 
+env: RUSTFLAGS="-D warnings"
+
 matrix:
   fast_finish: true
   include:
     - rust: nightly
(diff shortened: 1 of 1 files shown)


kpp:
> I'd still prefer environment config to be per matrix 

Since I eliminated env vars `BUILD_DOCS` and `BUILD_BOOK` you don't need to set up env per matrix. Every matrix build get pure `RUSTFLAGS="-D warnings"`. I can put it down explicitly if you want but there is no need to.

kpp:
@skade ping?

kpp:
@yoshuawuyts @stjepang ?

PR #119, opened 2019-08-28
Refactor TcpStream::connect into resolving loop and TcpStream::connect_to

kpp:
Now TcpStream::connect is a small function that iterates over addrs and tries find first good connection.

### DIFF
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -78,61 +78,10 @@ impl TcpStream {
     /// # Ok(()) }) }
     /// ```
     pub async fn connect<A: ToSocketAddrs>(addrs: A) -> io::Result<TcpStream> {
-        enum State {
-            Waiting(TcpStream),
-            Error(io::Error),
-            Done,
-        }
-
         let mut last_err = None;
 
         for addr in addrs.to_socket_addrs()? {
-            let mut state = {
-                match mio::net::TcpStream::connect(&addr) {
-                    Ok(mio_stream) => {
-                        #[cfg(unix)]
-                        let stream = TcpStream {
-                            raw_fd: mio_stream.as_raw_fd(),
-                            io_handle: IoHandle::new(mio_stream),
-                        };
-
-                        #[cfg(windows)]
-                        let stream = TcpStream {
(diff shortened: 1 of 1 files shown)


src/net/tcp/stream.rs
```diff
@@ -148,6 +97,78 @@ impl TcpStream {
...
+            stream
+        });
+
+        enum State {
+            Waiting(TcpStream),
+            Error(io::Error),
+            Done,
+        }
+        let mut state = match stream {
+            // TODO replace with .map_or_else(State::Error, State::Waiting)
```
stjepang: Can we delete this TODO? I actually prefer the `match` statement since it's easier to see what it does at a glance, despite it being a bit more verbose than `map_or_else`. :) It takes me a moment of pause to figure out how `State::Error` is constructed in this case and how it differs from `State::Waiting`.

src/net/tcp/stream.rs
```diff
@@ -148,6 +97,78 @@ impl TcpStream {
...
+    /// # fn main() -> std::io::Result<()> { async_std::task::block_on(async {
+    /// #
+    /// use async_std::net::TcpStream;
+    ///
+    /// let addr = "127.0.0.1".parse().unwrap();
+    /// let stream = TcpStream::connect_to(addr).await?;
+    /// #
+    /// # Ok(()) }) }
+    /// ```
+    pub async fn connect_to(addr: SocketAddr) -> io::Result<TcpStream> {
```
stjepang: I'd prefer not making this method public since the standard library doesn't have its equivalent. That does not mean I'm against having additional methods per se, but rather that it'd be easier to punt those decisions to post-1.0 release (Sep 26, 2019).

kpp: Sure

src/net/tcp/stream.rs
kpp: Ok

src/net/tcp/stream.rs
kpp: Fixed

src/net/tcp/stream.rs
kpp: Fixed

PR #120, opened 2019-08-28
Refactor TcpListener::incoming to use stream::poll_fn instread of hand-…

kpp:
…written impl futures::Stream for Incoming

### DIFF
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -28,7 +28,7 @@
 //! # }) }
 //! ```
 
-pub use tcp::{Incoming, TcpListener, TcpStream};
+pub use tcp::{TcpListener, TcpStream};
 pub use udp::UdpSocket;
 
 pub(crate) mod driver;
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -1,5 +1,4 @@
 use std::net::{self, SocketAddr, ToSocketAddrs};
-use std::pin::Pin;
 
 use cfg_if::cfg_if;
 use futures::future;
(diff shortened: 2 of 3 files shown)


stjepang:
Could you perhaps say a bit about the rationale for this PR?

The reason why we have `Incoming` is because the standard library also has [`std::net::Incoming`](https://doc.rust-lang.org/nightly/std/net/struct.Incoming.html), and our stream is just an async version of that iterator. We are trying really hard to not deviate from the standard library (except in a few special cases), no matter whether the standard library's API is good or bad.

kpp:
> Could you perhaps say a bit about the rationale for this PR?

I am a little bit against hand written `impl futures::Stream { fn poll_next() }` with `pin_utils::pin_mut` inside when we can avoid it.

> The reason why we have Incoming is because the standard library also has std::net::Incoming, and our stream is just an async version of that iterator.

I am not sure how to do this without implementing `futures::Stream` for a custom struct because impl Trait aliases are unstable. Is there another way?

stjepang:
> I am a little bit against hand-written `impl futures::Stream { fn poll_next() }` with `pin_utils::pin_mut` inside when we can avoid it.

I'm curious - why are you against that?

> I am not sure how to do this without implementing `futures::Stream` for a custom struct because impl Trait aliases are unstable. Is there another way?

Unfortunately, it is not possible until we get impl Trait aliases :(

kpp:
> I'm curious - why are you against that?

Because every time somebody needs a stream/future they implement them manually instead of using combinators from `futures-rs`. This is fine for a low-level stuff but it's not okay for a regular everyday simple code.

yoshuawuyts:
This doesn't seem like a patch we desire at this time. Thanks!

PR #123, opened 2019-08-28
Add future::poll_fn

stjepang:


kpp:
Why not `pub use futures`?

stjepang:
@kpp It's because we want to provide stronger stability guarantees, a more faithful feel of `std` than what `futures` has to offer, and more polished documentation that fits better into `async-std`.

We're still compatible with traits `Stream`, `AsyncRead`, `AsyncWrite`, and `AsyncSeek` from the `futures` crate, but don't think the extension traits like `FutureExt` or `StreamExt` have been polished enough to be worth exporting here. After all, the 0.3 version of `futures` has been unstable for a very long time and we want to be commit to stability in less than a month (on Sep 26).

berkus:
Committing to stability of unstable features may be not the right idea. I understand that std should feel stable, but it's as stable as its constituent parts - and async-std has at least half of it pretty unstable still.

yoshuawuyts:
I'm not sure why folks are downvoting the clarification from @stjepang, `async-std`'s principal designer. Being able to provide a consistent API surface is literally this crate's reason for existing. Defining most of our own types allows us to control our stability guarantees and user experience. It's a fundamental design principle that underpins _all_ of this crate.

We're committed to providing an ergonomic async counterpart to stdlib that's fully compatible with the core traits and async ecosystem underpinned by `futures-rs`. All solutions around how to use `futures-rs` have tradeoffs, and we feel comfortable with the path we've chosen.

So I'd like to take a moment here to provide some clarity: we're happy to answer any questions about design in earnest, but __we're not going to change the design direction of this crate__. If this is something you disagree with, that's okay -- but it might be good to realize this crate is not for you.

I hope this makes sense; I prefer being clear on these kinds of matters for the benefit of everyone involved.

---

@stjepang lol I just wrote https://github.com/async-rs/async-std/pull/124 with the exact same API. Seems like we had the same idea at the same time!

stjepang:
Closing in favor of #124

PR #124, opened 2019-08-29
add future::poll_fn

yoshuawuyts:
This adds `futures::poll_fn`, which is fairly common and useful when defining custom futures. This seems like a generally useful, and easy addition to the futures submodule. Thanks!

## Ref
- https://docs.rs/futures-preview/0.3.0-alpha.18/futures/future/fn.poll_fn.html

### DIFF
--- a/src/future/poll_fn.rs
+++ b/src/future/poll_fn.rs
@@ -1,6 +1,5 @@
 //! Definition of the `PollFn` adapter combinator
 
-use core::fmt;
 use core::pin::Pin;
 use std::future::Future;
 use std::task::{Context, Poll};
@@ -30,19 +29,13 @@ impl<F> Unpin for PollFn<F> {}
 /// }
 ///
 /// let read_future = poll_fn(read_line);
-/// assert_eq!(read_future.await, "Hello, World!".to_owned());
+/// assert_eq!(read_future.await, "Hello, World!");
 /// #
 /// # }) }
 /// ```
 pub async fn poll_fn<T>(f: impl FnMut(&mut Context<'_>) -> Poll<T>) -> T {
     let fut = PollFn { f };
     fut.await
-}
-
-impl<F> fmt::Debug for PollFn<F> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("PollFn").finish()
-    }
 }
 
 impl<T, F> Future for PollFn<F>
--- a/src/future/poll_fn.rs
+++ b/src/future/poll_fn.rs
@@ -1,0 +1,57 @@
+//! Definition of the `PollFn` adapter combinator
+
+use core::fmt;
+use core::pin::Pin;
+use std::future::Future;
+use std::task::{Context, Poll};
+
+/// Future for the [`poll_fn`] function.
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+struct PollFn<F> {
+    f: F,
+}
+
+impl<F> Unpin for PollFn<F> {}
+
+/// Creates a new future wrapping around a function returning `Poll`.
+///
+/// Polling the returned future delegates to the wrapped function.
+///
+/// # Examples
+///
+/// ```
+/// # fn main() { async_std::task::block_on(async {
+/// #
+/// use async_std::future::poll_fn;
+/// use async_std::task::{Context, Poll};
+///
+/// fn read_line(_cx: &mut Context<'_>) -> Poll<String> {
+///     Poll::Ready("Hello, World!".into())
+/// }
+///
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
Ooops, just noticed this is a dupe of #123. @stjepang I kind of like this version better if I'm honest. What do you think?

src/future/poll_fn.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+impl<T, F> Future for PollFn<F>
+where
+    F: FnMut(&mut Context<'_>) -> Poll<T>,
+{
+    type Output = T;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<T> {
```
kpp: Is it a Not Invented Here stance? Providing a stable API over hiding unstable futures-rs seems better than copy-pasting every function from `futures-rs`.

```
pub async fn poll_fn<F, T>(f: F) -> T
where
    F: FnMut(&mut Context<'_>) -> Poll<T>,
{
    futures::future::poll_fn(f).await
}
```

Feels much better.

jamesmunns:
> By the way why don't you close this PR in favor of #123?

Hey @kpp, @yoshuawuyts has requested review from @stjepang on the merits of #123 vs #124. As far as I have seen, @stjepang hasn't had a chance to address these points yet. You're involved with both conversations, so you should have visibility on all open discussions.

Additionally, regarding this:

> Is it a Not Invented Here stance? Providing a stable API over hiding unstable futures-rs seems better than copy-pasting every function from futures-rs.

Stjepan has provided context around this decision here: https://github.com/async-rs/async-std/pull/123#issuecomment-525947333, and Yosh has also given context and reasoning around this decision here: https://github.com/async-rs/async-std/pull/123#issuecomment-526120427. You may certainly have a different opinion, however reiterating it across multiple threads is unlikely to promote reasonable discussions on the topic.

You definitely seem passionate about this project, and that is a good thing! However the design decisions of this crate lie with @stjepang, and it would be appreciated if you presented differing opinions in a constructive way, and when there have been statements from the maintainers here pointing in a different direction, for that to be accepted.

Additionally, this is a project worked on in the free time of the maintainers, many of whom are still on vacation after RustConf. Comments like https://github.com/async-rs/async-std/pull/118#issuecomment-525906612 and https://github.com/async-rs/async-std/pull/118#issuecomment-526133788 in short succession requesting re-reviewhttps://github.com/async-rs/async-std/pull/118#issuecomment-526133788s are not necessary, and reduce the signal to noise ratio for notifications to the maintainers. Please allow a reasonable amount of time (at least a few days) before repinging. There are not so many open issues and PRs, and they are being worked through as the maintainers have available time.

We definitely appreciate the effort you are putting towards the numerous issues, PRs, and comments in this repo, we just ask that you allow time for design and other decisions to be made, and to collaborate productively with us to improve the project.

Please feel free to reach out to me privately via email (on my profile) if you have an questions or would like to discuss.

kpp:
What a neat response! Thanks

src/future/poll_fn.rs
```diff
@@ -0,0 +1,57 @@
...
+/// #
+/// use async_std::future::poll_fn;
+/// use async_std::task::{Context, Poll};
+///
+/// fn read_line(_cx: &mut Context<'_>) -> Poll<String> {
+///     Poll::Ready("Hello, World!".into())
+/// }
+///
+/// let read_future = poll_fn(read_line);
+/// assert_eq!(read_future.await, "Hello, World!".to_owned());
```
stjepang: I think we can delete the `to_owned()` part here.

src/future/poll_fn.rs
```diff
@@ -0,0 +1,57 @@
+//! Definition of the `PollFn` adapter combinator
+
+use core::fmt;
+use core::pin::Pin;
+use std::future::Future;
+use std::task::{Context, Poll};
+
+/// Future for the [`poll_fn`] function.
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+struct PollFn<F> {
```
stjepang: So this struct is private, which means it's debug impl, must_use, and unpin impl do nothing (dead code).

What do you think would be the end goal for these structs? Should we eventually make them public or keep functions like `poll_fn` just `async fn`s?

yoshuawuyts: I think `poll_fn` should stay an `async fn`; it looks nice, and I don't see much value in having the actual struct be public.

Also fwiw `Unpin` *does* actually do something here for us right now, so I'll keep it (else it doesn't compile).

yoshuawuyts:
`poll_fn` is already part of `async-std`; going to go ahead and close this.

PR #125, opened 2019-08-29
from/into stream

yoshuawuyts:
This adds `Stream` counterparts to `FromIterator`, `IntoIterator` and `Iterator::collect`, allowing to use the same patterns that are common in streams. Thanks!

## Tasks
- [x]  `FromStream`
- [x] `IntoStream`
- [x] `Stream::collect`

## Screenshot
Screenshot_2019-08-29 async_std stream - Rust

### DIFF
--- a/src/vec/from_stream.rs
+++ b/src/vec/from_stream.rs
@@ -1,0 +1,25 @@
+use crate::stream::{FromStream, IntoStream, Stream};
+
+use std::pin::Pin;
+
+impl<T> FromStream<T> for Vec<T> {
+    #[inline]
+    fn from_stream<'a, S: IntoStream<Item = T>>(
+        stream: S,
+    ) -> Pin<Box<dyn core::future::Future<Output = Self> + 'a>>
+    where
+        <S as IntoStream>::IntoStream: 'a,
+    {
+        let stream = stream.into_stream();
+
+        Pin::from(Box::new(async move {
+            pin_utils::pin_mut!(stream);
+
+            let mut out = vec![];
+            while let Some(item) = stream.next().await {
+                out.push(item);
+            }
+            out
+        }))
+    }
+}
--- a/src/vec/from_stream.rs
+++ b/src/vec/from_stream.rs
@@ -2,13 +2,13 @@ use crate::stream::{FromStream, IntoStream, Stream};
 
 use std::pin::Pin;
 
-impl<T> FromStream<T> for Vec<T> {
+impl<T: Send> FromStream<T> for Vec<T> {
     #[inline]
     fn from_stream<'a, S: IntoStream<Item = T>>(
         stream: S,
-    ) -> Pin<Box<dyn core::future::Future<Output = Self> + 'a>>
+    ) -> Pin<Box<dyn core::future::Future<Output = Self> + Send + 'a>>
     where
-        <S as IntoStream>::IntoStream: 'a,
+        <S as IntoStream>::IntoStream: Send + 'a,
     {
         let stream = stream.into_stream();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,6 +51,7 @@ pub mod prelude;
 pub mod stream;
 pub mod sync;
 pub mod task;
+mod vec;
 
 #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
 #[cfg(feature = "unstable")]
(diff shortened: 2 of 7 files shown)


yoshuawuyts:
Okay, so interesting fact here: if we actually want to provide implementations for `collect` here we'll have to add at least the `vec`, and probably also the `collections` submodules to hold the various `IntoStream` types (e.g. like `vec`'s https://doc.rust-lang.org/std/vec/struct.IntoIter.html).

I won't add it as part of this patch, but it seems like something that we should enumerate if we land this, and probably would make for a good issue to onboard folks with!

yoshuawuyts:
Noticed another thing:

- futures-rs implements `Stream` for `VecDeque` directly: https://docs.rs/futures-preview/0.3.0-alpha.18/futures/stream/trait.Stream.html#foreign-impls
- `Iterator` isn't available on `VecDeque`, instead `IntoIterator` is implemented for it: https://doc.rust-lang.org/std/collections/struct.VecDeque.html#implementations

This not only applies to `VecDeque`, but `Vec` as well (implements `Stream` directly). I'm uhhh, feeling like we're going in the right direction here.

Ideally we could also expose `stream` / `stream_mut` methods on the collections, but I think that might be more trouble than is worth it.

yoshuawuyts:
I've gone ahead and also implemented `Stream::collect`. Usage example:
```rust
use async_std::prelude::*;
use async_std::stream;

let s = stream::repeat(9u8).take(3);
let buf: Vec<u8> = s.collect().await;

assert_eq!(buf, vec![9; 3]);
```

I couldn't figure out the right trait bounds using static futures, so I just did a box bound on it. It shows up as `DynFuture` in the doc output. Hope that's okay!

src/stream/stream.rs
```diff
@@ -111,6 +127,48 @@ pub trait Stream {
...
+    /// let buf: Vec<u8> = s.collect().await;
+    ///
+    /// assert_eq!(buf, vec![9; 3]);
+    /// #
+    /// # }) }
+    /// ```
+    ///
+    /// [`stream`]: trait.Stream.html#tymethod.next
+    #[must_use = "if you really need to exhaust the iterator, consider `.for_each(drop)` instead (TODO)"]
+    fn collect<'a, B: FromStream<<Self as Stream>::Item>>(self) -> dyn_ret!('a, B)
```
stjepang: Any reason why the return type is `dyn Future` rather than `impl Future`? Shouldn't the `collect` method be an `async fn`?

yoshuawuyts: As far as I can see this is part of a trait rather than a concrete impl where `async fn` is allowed yet. Or am I missing something?

yoshuawuyts: Oh wait, I misread. I think you're right tho!

yoshuawuyts: @stjepang ahh, soooooooo -- this is the part I was talking about earlier; I tried *real* hard but couldn't get this to work using static futures. I gave up after a day of trying, and settled on obfuscated futures being "good enough". But if you think we really should have static futures here, maybe you could give it a shot?

yoshuawuyts:
@stjepang could you push your changes from last Tuesday if you have time?

stjepang:
@yoshuawuyts I have pushed the changes into `stjepang/async-std` (branch `into_stream_trait`) since I don't have permissions to push into your fork.

yoshuawuyts:
I had restore all the `Send` bounds that were removed in `stjepang/async-std` because the tests were failing (even if the code did compile). We've been able to keep the `Unpin` bounds out though. This patch should be all good now. Thanks!

src/vec/from_stream.rs
```diff
@@ -0,0 +1,25 @@
+use crate::stream::{FromStream, IntoStream, Stream};
+
+use std::pin::Pin;
+
+impl<T: Send> FromStream<T> for Vec<T> {
```
stjepang: Instead of adding this impl for `Vec<T>`, I wonder if it'd make more sense to make a blanket impl of this type:

```rust
impl<T, I> FromStream<T> for I
where
    I: FromIterator<Item = T>,
{
    // ...
}
```

We could make a similar blanket impl with `IntoStream`, too.

yoshuawuyts: Hmmm, yeah that's not a bad idea! Would substantially save us on the impls (e.g. half the list of #129) -- I'll give this a shot!

yoshuawuyts:
Also note from private chat: we should introduce this feature behind the "unstable" flag.

src/vec/from_stream.rs
yoshuawuyts: Talked in person about this, and the current approach is similar to Rayon which seems like a good indicator that it's worth doing.

yoshuawuyts:
All feedback implemented; talked about this and we're happy to try this out as "unstable" :tada:

bors r+

Issue #129, opened 2019-08-30
[tracking issue] streams


yoshuawuyts:
With #125 out, it's probably worth looking at which other parts of [`std::iter`](https://doc.rust-lang.org/std/iter/index.html) we can port to [`async_std::stream`](https://docs.rs/async-std/0.99.4/async_std/stream/index.html). This issue is intended to track what's left for us to port.

## Missing free functions
- [ ] `from_fn`
- [ ] `repeat_with`
- [ ] `successors`
- [ ] `once_with`

## Missing traits
- [ ]  `DoubleEndedStream`
- [ ] `ExactSizeStream`
- [ ] `Extend`
- [ ] `FusedStream`
- [ ] `Product`
- [ ] `Sum`

## Missing stream methods
- [ ] `Stream::all`
- [ ] `Stream::any`
- [ ] `Stream::by_ref`
- [ ] `Stream::chain`
- [ ] `Stream::cloned`
- [ ] `Stream::cmp`
- [x] `Stream::collect`
- [ ] `Stream::copied`
- [ ] `Stream::count`
- [ ] `Stream::cycle`
- [ ] `Stream::enumerate`
- [ ] `Stream::eq`
- [ ] `Stream::filter`
- [ ] `Stream::filter_map`
- [ ] `Stream::find`
- [ ] `Stream::find_map`
- [ ] `Stream::flat_map`
- [ ] `Stream::flatten`
- [ ] `Stream::fold`
- [ ] `Stream::for_each`
- [ ] `Stream::fuse`
- [ ] `Stream::ge`
- [ ] `Stream::gt`
- [ ] `Stream::inspect`
- [ ] `Stream::is_sorted`
- [ ] `Stream::is_sorted_by`
- [ ] `Stream::is_sorted_by_key`
- [ ] `Stream::last`
- [ ] `Stream::le`
- [ ] `Stream::lt`
- [ ] `Stream::map`
- [ ] `Stream::max`
- [ ] `Stream::max_by`
- [ ] `Stream::max_by_key`
- [ ] `Stream::min`
- [ ] `Stream::min_by`
- [ ] `Stream::min_by_key`
- [ ] `Stream::ne`
- [ ] `Stream::nth`
- [ ] `Stream::partial_cmp`
- [ ] `Stream::partition`
- [ ] `Stream::peekable`
- [ ] `Stream::position`
- [ ] `Stream::product`
- [ ] `Stream::rev`
- [ ] `Stream::rposition`
[...]

PR #130, opened 2019-08-30
add stream::from_fn

yoshuawuyts:
Implements `stream::poll_fn`. Ref #129.

Got one compile error around pin that I can't quite figure out; if someone could take a look to help out it'd be much appreciated. Thanks!

### DIFF
--- a/src/stream/from_fn.rs
+++ b/src/stream/from_fn.rs
@@ -3,6 +3,7 @@ use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
+use futures::ready;
 use futures::stream::Stream;
 
 /// Creates a new stream where each iteration calls the provided closure.
@@ -27,7 +28,6 @@ pub fn from_fn<T, F>(f: F) -> FromFn<F, T>
 /// See its documentation for more.
 ///
 /// [`stream::from_fn`]: fn.from_fn.html
-#[derive(Clone)]
 pub struct FromFn<F, T> {
     closure: F,
     fut: Option<Box<dyn Future<Output = Option<T>>>>,
@@ -44,8 +44,11 @@ impl<T, F: Unpin> Stream for FromFn<F, T>
             self.fut = Some((self.closure)());
         }
         
-        let pinned = unsafe { Pin::new_unchecked(&mut self.fut.unwrap()) };
-        pinned.poll(cx)
+        let pinned = unsafe { Pin::new_unchecked(&mut self.fut.as_mut().unwrap()) };
+        let out = ready!(pinned.poll(cx));
+
+        self.fut = None;
+        Poll::Ready(out)
     }
 }
 
--- a/src/stream/from_fn.rs
+++ b/src/stream/from_fn.rs
@@ -1,81 +1,100 @@
+use futures::stream::Stream;
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+
 use std::fmt;
 use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-use futures::ready;
-use futures::stream::Stream;
-
-/// Creates a new stream where each iteration calls the provided closure.
+/// Creates a new stream where each iteration calls the provided closure
+/// `F: FnMut() -> Option<T>`.
 ///
-/// This allows creating a custom iterator with any behavior
+/// This allows creating a custom stream with any behavior
 /// without using the more verbose syntax of creating a dedicated type
-/// and implementing the `Iterator` trait for it.
+/// and implementing the `Stream` trait for it.
+///
+/// Note that the `FromFn` stream doesn’t make assumptions about the behavior of the closure,
+/// and therefore conservatively does not implement [`FusedStream`](futures_core::stream::FusedStream).
+///
+/// The closure can use captures and its environment to track state across iterations. Depending on
+/// how the stream is used, this may require specifying the `move` keyword on the closure.
 ///
 /// # Examples
-/// 
+///
 /// ```
-/// # fn main() { async_std::task::block_on(async {
-/// #
-/// let mut count = 0;
-/// let counter = async_std::stream::from_fn(|| async move {
+/// # futures::executor::block_on(async {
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
Also depends on landing #125 first to make the example work`

laizy:
this fix the compile error: https://github.com/laizy/async-std/commit/5f6e5261c38b08f7848430bc23b4b8be982651dd

taiki-e:
@yoshuawuyts There is a higher-performance version: https://github.com/rust-lang-nursery/futures-rs/pull/1842 
Free feel to copy :)

src/stream/from_fn.rs
```diff
@@ -0,0 +1,104 @@
...
+/// let stream = stream::from_fn(|| {
+///     // Increment our count. This is why we started at zero.
+///
+///     let count = COUNT.fetch_add(1, Ordering::SeqCst);
+///
+///     // Check to see if we've finished counting or not.
+///     if count < 6 {
+///         future::ready(Some(count))
+///     } else {
+///         future::ready(None)
```
yoshuawuyts: We should make this an async closure (closure of async block) instead

taiki-e: Unfortunately, probably the equivalent of this cannot be written by async/await at this time. (see https://github.com/rust-lang-nursery/futures-rs/pull/1842#issuecomment-526666577)

taiki-e: (This is one of the unuseful features I mentioned in https://github.com/async-rs/async-std/issues/129#issuecomment-526833077.)

yoshuawuyts: My understanding is that it can't close over futures, but it *can* access `static`s, which may make it more useful?

stjepang:
I still think `stream::from_fn` should have a different signature :)

Consider the following types of constructors:

```rust
// We already have this that matches `iter::once`.
fn once<T>(t: T) -> impl Stream<T>;

// What I want and matches `iter::from_fn`.
fn from_fn<T>(f: impl FnOnce() -> T) -> impl Stream<T>;

// What this PR introduces, but names it `from_fn`.
fn from_async_fn<T>(f: impl FnOnce() -> impl Future<Output = T>) -> impl Stream<T>;

// We don't have this yet, but perhaps should.
fn from_future<T>(f: impl Future<Output = T>) -> impl Stream<T>;
```

I find `from_future` the most powerful constructor because all others can be trivially implemented using it:

```rust
fn once<T>(t: T) -> impl Stream<T> {
    from_future(async move { t) })
}

fn from_fn<T>(f: impl FnOnce() -> T) -> impl Stream<T> {
    from_future(async move { f() })
}

fn from_async_fn<T>(f: impl FnOnce() -> impl Future<Output = T>) -> impl Stream<T> {
    from_future(async move { f().await })
}
```

The main hazard of the `from_async_fn`-style constructor is in that it might have to be changed once the standard library introduces `AsyncFnOnce` trait. In that case, our constructor would change to:

```rust
fn from_async_fn<T>(f: impl AsyncFnOnce() -> T) -> impl Stream<T>;
```

I feel confident about having `once` and `from_fn` because they look like exact copies of `iter::once` and `iter::from_fn`.

But I don't feel confident about `from_async_fn` until the lang team decides what to do about the `AsyncFnOnce` trait.

Constructor `from_future` has no precedent in `std::iter` but I don't think we'd ever change our mind if we added it so that one is fine.

yoshuawuyts:
[`std::iter::from_fn`](https://doc.rust-lang.org/std/iter/fn.from_fn.html) has the following signature:

```rust
pub fn from_fn<T, F>(f: F) -> FromFn<F> where
    F: FnMut() -> Option<T>, 
```

The method I'm proposing here is an async version of that:

```rust
pub fn from_fn<T, F>(f: F) -> FromFn<F> where
    F: FnMut() -> impl Future<Output = Option<T>>,
```

@stjepang this would be equivalent to what you refer to as `from_async_fn`. In all fairness I don't see much value in adding `sync -> async` implementations because they could be implemented in an `async -> async` method with the same cost.

You do raise an interesting point about forwards compatibility. Given our signature there wouldn't be a conflict with `AsyncFnOnce` but with a possible `AsyncFnMut`. I'm not sure what this would look like, but wouldn't it be fair to assume it might be similar to how `async move || {}` and `|| async move {}` are equivalent for all intents and purposes?

---

However concluding; I agree there's no much inherent value to this one method. But I do like the idea of filling in all the blanks so we can point to stdlib to things we're missing (such as async variants of the closures perhaps?) But if we feel uncomfortable with this proposal I suggest we shelve it for now, and focus on things we feel more confident in -- overall this function doesn't seem that important, and we probably shouldn't spend too much time on it.

stjepang:
> The method I'm proposing here is an async version of that:

I guess my disagreement is around that - an async version of `iter::from_fn` should take a normal closure and not an async closure, just like `Stream::map` takes a normal closure and not an async closure.

This also got me thinking - an async closure taking no arguments is really not much different from a normal future.

Closures with no arguments are typically used to express laziness, for example:

```rust
// The "else value" gets computed only if `my_option` is `None`.
my_option.unwrap_or_else(|| compute_something_expensive());
```

But futures are already lazy by their nature, so a closure with no arguments returning a future is two layers of laziness. I find it difficult to come up with a situation where `async || {}` is more useful than `async {}`.

yoshuawuyts:
> But futures are already lazy by their nature, so a closure with no arguments returning a future is two layers of laziness.

Hmm, these are good questions. But I think the differentiator is not in whether they're lazy, but how they work. Consider this scenario:

```rust
let a = async {};
let b = async || {};

a.await;
a.await; // panic

b().await;
b().await; // ok
```

The difference is that `async {}` is at-most-once, while `async || {}` can be called repeatedly. The way I'm thinking about `from_fn` is that it's a placeholder for generators. In `std::iterator` it yields an "anonymous" iterator from sync calls. The way I view it in `std::stream` it should return an "anonymous" stream for async calls.

yoshuawuyts:
Let's revisit this at a later point as it's not that useful right now anyway. Thanks!

PR #132, opened 2019-08-30
Stream::all implementation

shekohex:
- [x] Adds stream::all implementation.
- [x] update method docs
- [x] an example for stream `all` method.
- [x] tests

PR #135, opened 2019-08-31
Stream::any implementation

shekohex:
- [x] Adds stream::any implementation.
- [x] update method docs
- [x] an example for stream `all` method.
- [x] tests

> this PR is just copy/pasted from #132 with just small changes in the logic.
is there a way that we can reuse the same code here?
I think it's fine.

### DIFF
--- a/src/stream/stream.rs
+++ b/src/stream/stream.rs
@@ -164,7 +164,64 @@ pub trait Stream {
     {
         AllFuture {
             stream: self,
-            result: true,
+            result: true, // the default if the empty stream
+            __item: PhantomData,
+            f,
+        }
+    }
+
+    /// Tests if any element of the stream matches a predicate.
+    ///
+    /// `any()` takes a closure that returns `true` or `false`. It applies
+    /// this closure to each element of the stream, and if any of them return
+    /// `true`, then so does `any()`. If they all return `false`, it
+    /// returns `false`.
+    ///
+    /// `any()` is short-circuiting; in other words, it will stop processing
+    /// as soon as it finds a `true`, given that no matter what else happens,
+    /// the result will also be `true`.
+    ///
+    /// An empty stream returns `false`.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    /// # fn main() { async_std::task::block_on(async {
+    /// #
+    /// use async_std::prelude::*;
+    /// use async_std::stream;
+    ///
+    /// let mut s = stream::repeat::<u32>(42).take(3);
+    /// assert!(s.any(|x| x ==  42).await);
+    ///
+    /// #
+    /// # }) }
+    /// ```
+    ///
+    /// Empty stream:
+    ///
+    /// ```
+    /// # fn main() { async_std::task::block_on(async {
+    /// #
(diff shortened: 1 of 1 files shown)


src/stream/stream.rs
```diff
@@ -273,3 +330,50 @@ where
         }
     }
 }
+
+#[derive(Debug)]
+pub struct AnyFuture<'a, S, F, T>
```
shekohex: i'm bad at naming things xD
```rust
type AnyFuture = Future<Output = Any>
```
sorry :joy_cat:

shekohex:
@jamesmunns could you please review this quickly and merge it so I can work on other methods?

src/stream/stream.rs
yoshuawuyts: I'm thinking we should probably hide this Future using the `impl/ImplFuture` macro

shekohex: is there any example, where is that macro?

yoshuawuyts: On a phone right now, but look for `macro_rules` in `Stream`. There's a macro that returns `ImplFuture` defined.

src/stream/stream.rs
```diff
@@ -273,3 +330,50 @@ where
...
+{
+    type Output = bool;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        use futures::Stream;
+        let next = futures::ready!(self.as_mut().stream().poll_next(cx));
+        match next {
+            Some(v) => {
+                let result = (self.as_mut().f())(v);
+                *self.as_mut().result() = result;
```
laizy: why we need a `result` field in `AnyFuture` and update every time here, instead:
```rust
        match next {
            Some(v) => {
                let result = (self.as_mut().f())(v);
                if result {
                    Poll::Ready(true)
                } else {
                    // don't forget to wake this task again to pull the next item from stream
                    cx.waker().wake_by_ref();
                    Poll::Pending
                }
            }
            None => Poll::Ready(false),
        }
```

shekohex: We should save the last result somewhere since the stream would yield `None` in two cases if it's already an empty stream or we get the end of the stream.

in the first case, the default value of `result` is `fasle` .
in the second case, we should return the last `result` when we get `None`.

in your example, the result of `stream::any` method would always return `false` whatever the stream is empty or have elements on it.

src/stream/stream.rs
shekohex: ok, I will try to do so.

shekohex:
Hi @yoshuawuyts could you please review the new changes?

src/stream/stream.rs
laizy: > in the second case, we should return the last result when we get None.

if the last `result` is `true`, does not it return `Poll::Ready(true)` directly in line 369? so if it reaches the `None` case, `self.result` should be `false`.

shekohex: If it returned with `Poll::Ready` then we are done, the Future is resolved and there is no calls to the `poll` method again.
So no, it can't reach the None case here, since the `any` method is short-circuit it will stop once it got `true`.

PR #138, opened 2019-09-02
Verify documentation links on travis

### LINKED ISSUE #82: Verify documentation links on Travis
Let's use [`cargo-deadlinks`](https://github.com/deadlinks/cargo-deadlinks) with the following command on Travis:

```rust
cargo deadlinks --check-http
```

This currently fails with some errors which are due to re-exports from `std`:

```
Found invalid urls in /home/stjepang/work/async-std/target/doc/async_std/io/type.Result.html:
        Linked file at path /home/stjepang/work/async-std/target/doc/async_std/result/enum.Result.html does not exist!
Found invalid urls in /home/stjepang/work/async-std/target/doc/async_std/io/struct.Error.html:
        Linked file at path /home/stjepang/work/async-std/target/doc/std/io/struct.Error.html does not exist!
        Linked file at path /home/stjepang/work/async-std/target/doc/std/io/enum.ErrorKind.html does not exist!
        Linked file at path /home/stjepang/work/async-std/target/doc/async_std/ffi/struct.NulError.html does not exist!
```

The way we resolve these errors is by writing shim docs for re-exports from `std`, similarly to how we did that here:

https://github.com/async-rs/async-std/blob/addda3937fe62c892e28747dd5aa33ce339bb7fa/src/os/unix/net/mod.rs#L14

The idea is that under the `docs` feature flag we generate "fake" docs linking to `async-std`'s types, but otherwise re-export real types from `std`.

freebroccolo:
Fixes #82

### DIFF
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -21,7 +21,7 @@
 //! ```
 
 #[doc(inline)]
-pub use std::io::{Error, ErrorKind, Result, SeekFrom};
+pub use std::io::SeekFrom;
 
 pub use buf_read::{BufRead, Lines};
 pub use buf_reader::BufReader;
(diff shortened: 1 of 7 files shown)


src/io/mod.rs
```diff
@@ -48,3 +47,357 @@ mod stdin;
...
+            /// ```
+            fn write_all(&mut self, mut buf: &[u8]) -> Result<()> {
+                unimplemented!()
+            }
+
+            /// Writes a formatted string into this writer, returning any error
+            /// encountered.
+            ///
+            /// This method is primarily used to interface with the
+            /// [`format_args!`] macro, but it is rare that this should
```
freebroccolo: I couldn't quite figure out how to get this approach to work for macros but I might be missing something simple.

For reference, in the docs for `io::Write` the macros are linked like this:

```rust
    /// This method is primarily used to interface with the
    /// [`format_args!`][formatargs] macro, but it is rare that this should
    /// explicitly be called. The [`write!`][write] macro should be favored to
    /// invoke this method instead.
    ///
    /// [formatargs]: ../macro.format_args.html
    /// [write]: ../macro.write.html
```

However, when I try to define those macros locally (either within `src/io/mod.rs` or within `src/lib.rs`) the `cargo deadlinks` still complains.

freebroccolo:
@stjepang when you have a moment would you please take a look at this and let me know if it's what you had in mind for resolving #82. I think this should fix all the current errors produced by `cargo deadlinks --check-http`.

If it looks okay, I'll squash the commits and mark it ready for review.

One thing that will probably need to be changed is the installation step for `cargo-deadlinks`. Should I just create a `ci/install-cargo-deadlinks.sh` and follow the same template as `ci/install-mdbook.sh`?

src/io/mod.rs
```diff
@@ -48,3 +47,357 @@ mod stdin;
...
+        /// }
+        /// ```
+        pub type Result<T> = std::result::Result<T, std::io::Error>;
+    } else {
+        #[doc(inline)]
+        pub use std::io::Result;
+    }
+}
+
+cfg_if! {
```
stjepang: We don't need to add this section for the `Write` trait because our `Write` trait is different from the one in `std` :)

src/net/mod.rs
```diff
@@ -28,9 +28,103 @@
...
+pub use tcp::Incoming;
 pub use udp::UdpSocket;
 
 pub(crate) mod driver;
 mod tcp;
 mod udp;
+
+use cfg_if::cfg_if;
+
+cfg_if! {
```
stjepang: No need to add this section for `TcpListener` because ours differs from the one in `std`

stjepang: The same applies to `TcpStream`

src/os/unix/fs.rs
```diff
@@ -40,14 +40,25 @@ cfg_if! {
...
         /// Unix-specific extension methods for `DirEntry`.
         pub trait DirEntryExt {
             /// Returns the underlying `d_ino` field in the contained `dirent`
             /// structure.
             fn ino(&self) -> u64;
         }
+    }
+}
 
+cfg_if! {
```
stjepang: Let's also delete this section.

src/os/unix/fs.rs
```diff
@@ -40,14 +40,25 @@ cfg_if! {
             /// `0o777`.
             fn mode(&mut self, mode: u32) -> &mut Self;
         }
+    } else {
```
stjepang: Let's also delete this `else` block.

src/os/unix/fs.rs
```diff
@@ -70,6 +81,68 @@ cfg_if! {
...
+            ///
+            /// # Examples
+            ///
+            /// ```
+            /// use std::fs::Permissions;
+            /// use std::os::unix::fs::PermissionsExt;
+            ///
+            /// // Read/write for owner and read for others.
+            /// let permissions = Permissions::from_mode(0o644);
+            /// assert_eq!(permissions.mode(), 0o644);
```
stjepang: It'd also be good to update these examples to use `async_std` instead of `std`, but we don't have to do that in this PR.

stjepang:
> One thing that will probably need to be changed is the installation step for `cargo-deadlinks`. Should I just create a `ci/install-cargo-deadlinks.sh` and follow the same template as `ci/install-mdbook.sh`?

Yes, that sounds reasonable to me!

src/task/mod.rs
```diff
@@ -35,3 +35,43 @@ mod sleep;
...
+
+cfg_if! {
+    if #[cfg(feature = "docs")] {
+        /// A `Waker` is a handle for waking up a task by notifying its executor that it
+        /// is ready to be run.
+        ///
+        /// This handle encapsulates a [`RawWaker`] instance, which defines the
+        /// executor-specific wakeup behavior.
+        ///
+        /// Implements [`Clone`], [`trait@Send`], and [`trait@Sync`].
```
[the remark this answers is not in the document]
freebroccolo: Yeah, the links do actually work there. They currently point to the docs upstream.

rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2019-09-01
```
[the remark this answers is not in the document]
freebroccolo: Not necessary, I've just been pinning the toolchain on these PRs because I develop on Windows sometimes and the build has been broken with current nightly. I'll remove it before readying for review.

src/os/unix/fs.rs
freebroccolo: If I delete these doc shims in `src/os/unix/fs.rs`, then I start to get errors like this:

```
Found invalid urls in async-std/target/doc/async_std/os/unix/fs/index.html:
        Linked file at path async-std/target/doc/std/fs/struct.OpenOptions.html does not exist!
```

I tried to trace back to where all of these are being referenced but wasn't successful. Do you still want to try and remove them?

src/os/unix/fs.rs
freebroccolo: Okay, I will create a separate issue for that.

freebroccolo:
@stjepang Thanks for the feedback!

Responding to a few specific points you raised:

1) I've been pinning the toolchain since I work on Windows sometimes and the build is (was?) broken on current nightly.

2) The `trait@Send` links do indeed work and they are currently pointed to the docs upstream.

3) I've been unable to remove some of the shims in `os/unix/fs.rs`; see comments above.

Finally, installing `cargo-deadlinks` following the same template as [ci/install-mdbook.sh](https://github.com/async-rs/async-std/blob/master/ci/install-mdbook.sh) isn't going to work. The reason is because the script at https://japaric.github.io/trust/install.sh has the release download url hard-coded for a specific tag format that the [deadlinks/cargo-deadlinks](https://github.com/deadlinks/cargo-deadlinks) repository does not use.

Given that, I'm not sure how you want to handle the installation. Thoughts? (cc @jamesmunns)

freebroccolo:
Going to close this. If someone would like to resume the work, feel free to ping me.

PR #140, opened 2019-09-03
Reduce dependency on futures crate

stjepang:
This PR reduces our dependency on the futures crate by only keeping `futures-channel-preview`, `futures-core-preview`, and `futures-io-preview` as dependencies.

The `futures-timer` crate still pulls the whole `futures-preview` crate as a transitive dependency, so this PR is not yet *the* solution, but is a first step towards it.

In the end, we'd like to depend only on `futures-core-preview` and `futures-io-preview`, only using traits `Stream`, `AsyncRead`, `AsyncWrite`, and nothing else.


skade:
This is fails on travis is failing imports.

src/future/mod.rs
```diff
@@ -4,7 +4,9 @@
 pub use std::future::Future;
 
 pub use pending::pending;
+pub use poll_fn::poll_fn;
```
yoshuawuyts: This seems to depend on #124.

PR #144, opened 2019-09-04
move a-chat tutorial's code to this repo

matklad:
After this, I propose archiving https://github.com/async-rs/a-chat/: it seems bad that folks need to send PR to to repositories.


matklad:
We can do that if we merge client and server into a single binary, i’ll Do that

matklad:
should be better now!

stjepang:
Perfect, thanks! Can you run `cargo fmt` and then merge this PR?

PR #146, opened 2019-09-05
add stream::min_by method

montekki:
Implements `Stream::min_by`.

Not sure if:
1. Should live in a separate file, but it looks like adding all those methods to `stream.rs` will bloat at some point.
2. It can be effectively reused for `Stream::min`
3. Other `min*` and `max*` should be added to this PR.

### DIFF
--- a/src/stream/min_by.rs
+++ b/src/stream/min_by.rs
@@ -1,0 +1,54 @@
+use std::cmp::Ordering;
+use std::pin::Pin;
+
+use super::stream::Stream;
+use crate::future::Future;
+use crate::task::{Context, Poll};
+
+/// A future that yields the minimum item in a stream by a given comparison function.
+#[derive(Clone, Debug)]
+pub struct MinBy<S: Stream, F> {
+    stream: S,
+    compare: F,
+    min: Option<S::Item>,
+}
+
+impl<S: Stream + Unpin, F> Unpin for MinBy<S, F> {}
+
+impl<S: Stream + Unpin, F> MinBy<S, F> {
+    pub(super) fn new(stream: S, compare: F) -> Self {
+        MinBy {
+            stream,
+            compare,
+            min: None,
+        }
+    }
+}
+
+impl<S, F> Future for MinBy<S, F>
+where
+    S: futures_core::stream::Stream + Unpin,
+    S::Item: Copy,
+    F: FnMut(&S::Item, &S::Item) -> Ordering,
+{
+    type Output = Option<S::Item>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let next = futures_core::ready!(Pin::new(&mut self.stream).poll_next(cx));
+
+        match next {
+            Some(new) => {
+                cx.waker().wake_by_ref();
+                match self.as_mut().min.take() {
+                    None => self.as_mut().min = Some(new),
+                    Some(old) => match (&mut self.as_mut().compare)(&new, &old) {
+                        Ordering::Less => self.as_mut().min = Some(new),
+                        _ => self.as_mut().min = Some(old),
+                    },
+                }
+                Poll::Pending
+            }
+            None => Poll::Ready(self.min),
+        }
+    }
+}
--- a/src/stream/stream.rs
+++ b/src/stream/stream.rs
@@ -122,7 +122,7 @@ pub trait Stream {
 
     /// Returns the element that gives the minimum value with respect to the
     /// specified comparison function. If several elements are equally minimum,
-    /// the first element is returned. If the stream is empty, None is returned.
+    /// the first element is returned. If the stream is empty, `None` is returned.
     ///
     /// # Examples
     ///
--- a/src/stream/stream.rs
+++ b/src/stream/stream.rs
@@ -21,10 +21,12 @@
 //! # }) }
 //! ```
 
+use std::cmp::Ordering;
 use std::pin::Pin;
 
 use cfg_if::cfg_if;
 
+use super::min_by::MinBy;
 use crate::future::Future;
 use crate::task::{Context, Poll};
 use std::marker::PhantomData;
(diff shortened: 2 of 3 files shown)


yoshuawuyts:
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.min_by

Ref: #129

---

1. I agree. We should do this in a separate PR probably though.
2. Yay!
3. Yayay! Though doing these PRs one at the time is also welcome; it allows us to review and land them at a faster pace which might be nice for everyone!

src/stream/stream.rs
```diff
@@ -118,6 +120,39 @@ pub trait Stream {
         }
     }
 
+    /// Returns the element that gives the minimum value with respect to the
+    /// specified comparison function. If several elements are equally minimum,
+    /// the first element is returned. If the stream is empty, None is returned.
```
yoshuawuyts:
```suggestion
    /// the first element is returned. If the stream is empty, `None` is returned.
```

src/stream/min_by.rs
```diff
@@ -0,0 +1,54 @@
...
+            stream,
+            compare,
+            min: None,
+        }
+    }
+}
+
+impl<S, F> Future for MinBy<S, F>
+where
+    S: futures_core::stream::Stream + Unpin,
```
yoshuawuyts: Ah, this bound may need to be updated to just be `Stream` as #140 was merged.

montekki: hmmm, you probably mean #145 ?

montekki:
>     1. I agree. We should do this in a separate PR probably though.
ok
>     2. Yay!

Those were questions prefixed by "Not sure if" actually :) So at the point I have no idea about this point

>     3. Yayay! Though doing these PRs one at the time is also welcome; it allows us to review and land them at a faster pace which might be nice for everyone!
yep, sure

src/stream/min_by.rs
yoshuawuyts: Ah yes indeed

src/stream/stream.rs
```diff
@@ -118,6 +120,39 @@ pub trait Stream {
...
+    /// assert_eq!(min, Some(3));
+    ///
+    /// let min = Stream::min_by(VecDeque::<usize>::new(), |x, y| x.cmp(y)).await;
+    /// assert_eq!(min, None);
+    /// #
+    /// # }) }
+    /// ```
+    fn min_by<F>(self, compare: F) -> MinBy<Self, F>
+    where
+        Self: Sized + Unpin,
```
stjepang: The `Unpin` bound here is not necessary because we're taking ownership of `self` in this method.

src/stream/min_by.rs
```diff
@@ -0,0 +1,54 @@
+use std::cmp::Ordering;
+use std::pin::Pin;
+
+use super::stream::Stream;
+use crate::future::Future;
+use crate::task::{Context, Poll};
+
+/// A future that yields the minimum item in a stream by a given comparison function.
+#[derive(Clone, Debug)]
+pub struct MinBy<S: Stream, F> {
```
stjepang: Since `MinBy` is here in a separate file, it'd be a good idea to move `Take` and others into their own files, too.

src/stream/min_by.rs
stjepang: Is this still unresolved?

src/stream/min_by.rs
stjepang: But we can do that in a follow-up PR :)

src/stream/min_by.rs
```diff
@@ -0,0 +1,54 @@
...
+    stream: S,
+    compare: F,
+    min: Option<S::Item>,
+}
+
+impl<S: Stream + Unpin, F> Unpin for MinBy<S, F> {}
+
+impl<S: Stream + Unpin, F> MinBy<S, F> {
```
stjepang: Instead of requiring `Unpin`, we should use `unsafe_pinner` like for the `Take` struct.

src/stream/min_by.rs
```diff
@@ -0,0 +1,54 @@
...
+    F: FnMut(&S::Item, &S::Item) -> Ordering,
+{
+    type Output = Option<S::Item>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let next = futures_core::ready!(Pin::new(&mut self.stream).poll_next(cx));
+
+        match next {
+            Some(new) => {
+                cx.waker().wake_by_ref();
```
stjepang: Is this line needed at all? I believe it can be deleted.

src/stream/min_by.rs
stjepang: Since this is a future, it should be named `MinByFuture` just like `AllFuture` and `AnyFuture`.

src/stream/stream.rs
```diff
@@ -118,6 +120,39 @@ pub trait Stream {
...
+    ///
+    /// let min = Stream::min_by(s, |x, y| y.cmp(x)).await;
+    /// assert_eq!(min, Some(3));
+    ///
+    /// let min = Stream::min_by(VecDeque::<usize>::new(), |x, y| x.cmp(y)).await;
+    /// assert_eq!(min, None);
+    /// #
+    /// # }) }
+    /// ```
+    fn min_by<F>(self, compare: F) -> MinBy<Self, F>
```
stjepang: The return type here should use `ret!` like in `fn any`, for example. This function would ideally be an `async fn` if Rust supported that syntax in traits today.

src/stream/stream.rs
```diff
@@ -21,10 +21,12 @@
 //! # }) }
 //! ```
 
+use std::cmp::Ordering;
 use std::pin::Pin;
 
 use cfg_if::cfg_if;
 
+use super::min_by::MinBy;
```
stjepang: Since `min_by` should ideally be an `async fn`, let's make `MinBy` a hidden type and not even implement `Debug` (because `async fn` desugaring doesn't implement `Debug` for its future).

stjepang:
Sorry for taking so long to review this! :(

Since there's no way to revert and re-open this PR, could you please address my comment in a follow-up PR?

montekki:
yep, sure, Ok to do it in a followup to the #150?

stjepang:
yeah, sounds good! :)

src/stream/stream.rs
montekki: How is `ret!` supposed to work for
1. Futures that take `self` instead of `&mut self` and thus should have no lifetime parameters
2. Streams
?

src/stream/min_by.rs
yoshuawuyts: See https://github.com/async-rs/async-std/pull/150!

PR #148, opened 2019-09-05
Make sure into_raw_fd doesn't close the file descriptor

### LINKED ISSUE #147: socket closed after UnixStream::into_raw_fd()
UnixStream holds a `IoHandle<mio_uds::UnixStream>` which is not correctly disposed of in `into_raw_fd`, resulting in the returned file descriptor being closed when `into_raw_fd` returns, making the function useless.

yshui:
Closes #147

### DIFF
--- a/src/net/driver/mod.rs
+++ b/src/net/driver/mod.rs
@@ -279,9 +279,9 @@ impl<T: Evented> IoHandle<T> {
         Ok(())
     }
 
-    /// Deregister and return the I/O source
+    /// Deregisters and returns the inner I/O source.
     ///
-    /// This method is to support IntoRawFd in struct that uses IoHandle
+    /// This method is typically used to convert `IoHandle`s to raw file descriptors/handles.
     pub fn into_inner(mut self) -> T {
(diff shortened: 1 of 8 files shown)


src/net/driver/mod.rs
```diff
@@ -278,13 +278,26 @@ impl<T: Evented> IoHandle<T> {
 
         Ok(())
     }
+
+    /// Deregister and return the I/O source
```
stjepang:
```suggestion
    /// Deregisters and returns the inner I/O source.
```

src/net/driver/mod.rs
```diff
@@ -278,13 +278,26 @@ impl<T: Evented> IoHandle<T> {
...
+    ///
+    /// This method is to support IntoRawFd in struct that uses IoHandle
```
stjepang:
```suggestion
    /// This method is typically used to convert `IoHandle`s to raw file descriptors/handles.
```

yshui:
@stjepang Thanks. Applied

PR #149, opened 2019-09-06
update deps

yoshuawuyts:
Updates all deps. Thanks!


yoshuawuyts:
- cargo fmt fails because it can't be installed: https://travis-ci.com/async-rs/async-std/jobs/231791179
- mdbook is failing because it can't find `futures::channel` https://travis-ci.com/async-rs/async-std/jobs/231791181 -- I think we should investigate this

PR #150, opened 2019-09-06
split stream into multiple files

yoshuawuyts:
This splits `stream/mod.rs`'s combinators into multiple files, making it easier to contribute combinators. Additionally we've renamed `MinBy` to `MinByFuture` to make it the same as the other private futures. Ref #146 #129. Thanks!


yoshuawuyts:
`cargo fmt` is failing because it's broken on nightly https://travis-ci.com/async-rs/async-std/jobs/231806192 -- this patch should be good to merge tho!

PR #152, opened 2019-09-06
fixes docs for io::buf_read::read_until

montekki:
Docs are a bit misleading, no use reading into a buffer that is already full.


skade:
Correct. A+ Would be an additional example that shows the append behaviour. We mention it in the docs, but there's possibly some room for improvement for people that learn by example.

PR #155, opened 2019-09-06
Use the latest toolchain with rustfmt available if rustfmt is unavailable on the latest nightly

### LINKED ISSUE #154: Fix nightly rustfmt
We should implement the patch done by @taiki-e in https://github.com/rustasync/runtime/pull/104 here too to fix our build failures. Thanks!

Atul9:
closes #154

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,14 @@ matrix:
     - name: fmt
       rust: nightly
       os: linux
-      before_script:
-        - rustup component add rustfmt
+      before_script: |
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,7 +19,13 @@ matrix:
       rust: nightly
       os: linux
       before_script:
```
taiki-e:
```suggestion
       before_script: |
```

taiki-e: Nope, it may not be necessary.

PR #157, opened 2019-09-07
More robust file implementation

stjepang:
This is a reimplementation of the `File`s state machine.

The previous implementation was simple and a bit naive. It was not fundamentally wrong but had surprises in some corner cases. For example, if an async read operation was started but we timed out on it, the file cursor would move even though we didn't complete the operation. The new implementation will move the cursor only when read/write operations complete successfully.

There was also a deadlock hazard in the case where multiple tasks were concurrently reading or writing to the same file, in which case some task wakeups would be lost. This PR fixes the problem.

A nice consequence of this PR: `futures-channel` is now unused, so we can remove it from the dependency list.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,10 +27,9 @@ unstable = []
 async-task = "1.0.0"
 cfg-if = "0.1.9"
 crossbeam-channel = "0.3.9"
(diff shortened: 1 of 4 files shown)


yoshuawuyts:
Rebased on master to enable tests to pass.

Issue #158, opened 2019-09-07
mdbook is failing to build


yoshuawuyts:
Example https://travis-ci.com/async-rs/async-std/jobs/232070246; this has been happening to all recent PRs.

PR #159, opened 2019-09-07
disable mdbook to allow tests to pass again

yoshuawuyts:
Hotfix for #158. This is not a permanent solution, and we should investigate what's the root cause of this. However the more pressing issue is that we can continue to land patches with feedback from CI on the actual implementation. Thanks!


yoshuawuyts:
Going to go ahead and merge this, but keep #159 in order to make sure we fix it.

PR #162, opened 2019-09-08
fixes to stream::min_by

montekki:
1. `min_by` should not have a `Self: Unpin` bound.
2. uses `unsafe_pinned` instead of an `Unpin` bound in the `impl`s.
3. leaves `waker().wake_by_ref()` line in place since here we are returning `Poll::Pending` upon receiving new elements from the `inner` stream and need to re-schedule ourselves to receive new elements (as in `AllFuture`)
4. It's not clear at the moment how `ret!` macro should be used for returning types without explicit lifetimes. That's the case for `Take` and `MinByFuture`.


yoshuawuyts:
Ref #146.

PR #163, opened 2019-09-08
adds stream::filter_map combinator

montekki:
Implements a `flat_map` combinator. Currently the same about `ret!` as in #162 .

Also the naming should probably be `FilterMapStream`, but in that case `Take` stream should also change it's name i guess.

Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map
Ref: #129

### DIFF
--- a/src/stream/stream/mod.rs
+++ b/src/stream/stream/mod.rs
@@ -23,6 +23,7 @@
 
 mod all;
 mod any;
+mod filter_map;
 mod min_by;
 mod next;
 mod take;
@@ -31,6 +32,7 @@ pub use take::Take;
 
 use all::AllFuture;
 use any::AnyFuture;
+use filter_map::FilterMap;
 use min_by::MinByFuture;
 use next::NextFuture;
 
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
> Also the naming should probably be `FilterMapStream`.

We've made a deliberate choice to name the stream types the same as in stdlib. So `FilterMap` and `Take` are as intended!

src/stream/stream/mod.rs
```diff
@@ -128,6 +130,45 @@ pub trait Stream {
...
     }
 
+    /// Both filters and maps a stream.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```
+    ///
```
```suggestion
```

src/stream/stream/mod.rs
```diff
@@ -128,6 +130,45 @@ pub trait Stream {
...
+    /// let three = parsed.next().await;
+    /// assert_eq!(three, Some(3));
+    ///
+    /// let five = parsed.next().await;
+    /// assert_eq!(five, Some(5));
+    ///
+    /// let end = parsed.next().await;
+    /// assert_eq!(end, None);
+    ///
```
```suggestion
```

Issue #131, opened 2019-08-30
[tracking] io


yoshuawuyts:
Similar to #129 for streams, this issue tracks what's left to port from [`std::io`](https://doc.rust-lang.org/std/io/index.html) to [`async_std::io`](https://docs.rs/async-std/0.99.4/async_std/io/index.html).

## Top-level exports
- `prelude`

## Free functions
- [x] `copy`
- [x] `empty`
- [x] `repeat`
- [x] `sink`
- [x] `stderr`
- [x] `stdin`
- [x] `stdout`

## Structs
- [x] `BufReader`
- [x] `BufWriter`
- [x] `Bytes`
- [x] `Chain`
- [x] `Cursor`
- [x] `Empty`
- [x] `Error`
- [x] `IntoInnerError`
- [x] `IoSlice`
- [x] `IoSliceMut`
- [ ] `LineWriter`
- [x] `Lines`
- [x] `Repeat`
- [x] `Sink`
- [x] `Split`
- [x] `Stderr`
- [x] `StderrLock`
- [x] `Stdin`
- [x] `StdinLock`
- [x] `Stdout`
- [x] `StdoutLock`
- [x] `Take`

## `Read` methods
- [x] `Read::by_ref`
- [x] `Read::bytes`
- [x] `Read::chain`
- [x] `Read::read_exact`
- [x] `Read::read_to_end`
- [x] `Read::read_to_string`
- [x] `Read::read_vectored`
- [x] `Read::take`

## `Write` methods
- [ ] `Write::by_ref`
- [x] `Write::write_all`
- [x] `Write::write_fmt`
- [x] `Write::write_vectored`

## `BufRead` methods
- [x] `BufRead::buffer`
- [x] `BufRead::consume`
- [x] `BufRead::lines`
- [x] `BufRead::read_line`
- [x] `BufRead::read_until`
- [x] `BufRead::split`

## `BufWriter` methods
- [x] `BufWriter::buffer`
- [x] `BufWriter::get_mut`
- [x] `BufWriter::get_ref`
- [x] `BufWriter::into_inner`
- [x] `BufWriter::new`
- [x] `BufWriter::with_capacity`

## `BufReader` methods
- [x] `BufReader::fill_buf`
- [x] `BufReader::get_mut`
- [x] `BufReader::get_ref`
- [x] `BufReader::into_inner`
- [x] `BufReader::new`
- [x] `BufReader::with_capacity`

taiki-e:
[...]

PR #167, opened 2019-09-08
add io::cursor

yoshuawuyts:
Adds `io::Cursor` and makes it so `io::prelude::*` behaves the way it does in std (so it can actually be implemented - though this might just have been a bug on my side??).

Ref #131. Thanks!


stjepang:
Why is the `AsyncBufRead` impl commented out? Is there something we're still missing that would make it compile?

yoshuawuyts:
@stjepang yeah, not sure -- I can't get the borrow checker to allow it. If you want to take a look that'd definitely be helpful.

2019-09-10-134945_1920x1080

taiki-e:
@yoshuawuyts it needs to use Pin::get_mut
```rust
Poll::Ready(io::BufRead::fill_buf(&mut self.get_mut().inner))
```
https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-io/src/lib.rs#L599

yoshuawuyts:
@taiki-e that did the trick! -- thanks so much!

PR #174, opened 2019-09-10
adds stream::find_map combinator

montekki:
Adds a `stream::find_map` combinator
---
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map
Ref: #129


yoshuawuyts:
Bors r+

PR #176, opened 2019-09-10
Add BufRead::fill_buf

tirr-c:
Tracking issue: #131. Docs from `std`.

**Note** that there is a use of `unsafe` in `fill_buf.rs` that transmutes the lifetime of output buffer.

### DIFF
--- a/src/io/buf_read/fill_buf.rs
+++ b/src/io/buf_read/fill_buf.rs
@@ -12,8 +12,8 @@ pub struct FillBufFuture<'a, R: ?Sized> {
     reader: &'a mut R,
 }
 
-impl<'a, R: ?Sized> From<&'a mut R> for FillBufFuture<'a, R> {
-    fn from(reader: &'a mut R) -> Self {
+impl<'a, R: ?Sized> FillBufFuture<'a, R> {
+    pub(crate) fn new(reader: &'a mut R) -> Self {
         Self { reader }
     }
 }
--- a/src/io/buf_read/fill_buf.rs
+++ b/src/io/buf_read/fill_buf.rs
@@ -1,0 +1,32 @@
+use std::pin::Pin;
+
+use futures_io::AsyncBufRead;
+
+use crate::future::Future;
+use crate::io;
+use crate::task::{Context, Poll};
+
+#[doc(hidden)]
+#[allow(missing_debug_implementations)]
+pub struct FillBufFuture<'a, R: ?Sized> {
+    reader: &'a mut R,
+}
+
+impl<'a, R: ?Sized> From<&'a mut R> for FillBufFuture<'a, R> {
+    fn from(reader: &'a mut R) -> Self {
+        Self { reader }
+    }
+}
+
+impl<'a, R: AsyncBufRead + Unpin + ?Sized> Future for FillBufFuture<'a, R> {
+    type Output = io::Result<&'a [u8]>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<&'a [u8]>> {
+        let Self { reader } = &mut *self;
+        let result = Pin::new(reader).poll_fill_buf(cx);
+        // This is safe because:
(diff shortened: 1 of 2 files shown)


src/io/buf_read/fill_buf.rs
```diff
@@ -0,0 +1,32 @@
...
+#[allow(missing_debug_implementations)]
+pub struct FillBufFuture<'a, R: ?Sized> {
+    reader: &'a mut R,
+}
+
+impl<'a, R: ?Sized> From<&'a mut R> for FillBufFuture<'a, R> {
+    fn from(reader: &'a mut R) -> Self {
+        Self { reader }
+    }
+}
```
yoshuawuyts: Given there is only a single creator of this future, is this `From` impl necessary? If it is, could we perhaps add a comment to explain why?

tirr-c: I wanted to make `reader` field private, but yeah, it could be an associated function, or just make `reader` `pub(crate)`.

stjepang: @tirr-c This is not a big issue, but given a choice I'd prefer a `pub(crate)` field and a function in order to not leak any unintended APIs publicly. This `impl` is currently public in the sense that the user could do `.into()` to convert a buffer into the future.

src/io/buf_read/fill_buf.rs
```diff
@@ -0,0 +1,32 @@
...
+impl<'a, R: AsyncBufRead + Unpin + ?Sized> Future for FillBufFuture<'a, R> {
+    type Output = io::Result<&'a [u8]>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<&'a [u8]>> {
+        let Self { reader } = &mut *self;
+        let result = Pin::new(reader).poll_fill_buf(cx);
+        // This is safe because:
+        // 1. The buffer is valid for the lifetime of the reader.
+        // 2. Output is unrelated to the wrapper (Self).
+        result.map_ok(|buf| unsafe { std::mem::transmute::<_, &'a [u8]>(buf) })
```
stjepang: I'm always a bit anxious about mixing type inference and `transmute`, so could you perhaps make types more explicit like this?

```suggestion
        result.map_ok(|buf| unsafe { std::mem::transmute::<&'_ [u8], &'a [u8]>(buf) })
```

stjepang:
A logical follow-up now would be to add the `consume` method. I'm okay with either adding it in this PR or in a new PR.

tirr-c:
@stjepang Review addressed!

taiki-e:
This implementation is unsound (this already reverted in https://github.com/async-rs/async-std/commit/edfa2358a42872af35024dd83e78ea80e8fe6a4a without explanation, but I leave this comment to prevent someone from accidentally using this implementation, like #763).

Specifically, this implementation allows you to write code like this (this is because `FillBufFuture::poll` is changing the lifetime of the output):

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b8a8bfeba15e480844c8ac2a1df35c5b)

```rust
let mut r = R([0]);
let mut r = FillBufFuture::new(&mut r);
let mut r = Pin::new(&mut r);
let w = noop_waker();
let mut cx = Context::from_waker(&w);

let x = r.as_mut().poll(&mut cx);
println!("{:?}", x); // Ready(Ok([1]))
// `x` has not been dropped yet, but we can call `poll` again.
// This is not correct.
let _y = r.as_mut().poll(&mut cx);
println!("{:?}", x); // Ready(Ok([2]))

struct R([u8; 1]);
impl AsyncBufRead for R {
    fn poll_fill_buf(mut self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<io::Result<&[u8]>> {
        self.0[0] += 1;
        Poll::Ready(Ok(&self.into_ref().get_ref().0))
    }
    fn consume(self: Pin<&mut Self>, _: usize) {
        unimplemented!()
    }
}

```

PR #177, opened 2019-09-10
implement DoubleEndedStream

yoshuawuyts:
Ref #129. This is the most basic version of the `DoubleEndedStream` trait. Because there is no counterpart in `futures-rs` we allow this to be implementable (not sure if we should though?).

This is not a high-priority trait to implement, with probably the most useful addition being the blanket impl over [`std::iter::Fuse`](https://doc.rust-lang.org/std/iter/struct.Fuse.html) (where we should have a `Fuse` counterpart for `Stream` also).

So I'm taking this one step at the time, and this PR introduces just the bare minimum to get things working. Thanks!

r? @stjepang @taiki-e

## Refs
- https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html
- #129

### DIFF
--- a/src/stream/double_ended_stream.rs
+++ b/src/stream/double_ended_stream.rs
@@ -1,0 +1,22 @@
+use crate::stream::Stream;
+
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+/// An stream able to yield elements from both ends.
+///
+/// Something that implements `DoubleEndedStream` has one extra capability
+/// over something that implements [`Stream`]: the ability to also take
+/// `Item`s from the back, as well as the front.
+///
+/// [`Stream`]: trait.Stream.html
+pub trait DoubleEndedStream: Stream {
+    /// Removes and returns an element from the end of the stream.
+    ///
+    /// Returns `None` when there are no more elements.
+    ///
+    /// The [trait-level] docs contain more details.
+    ///
+    /// [trait-level]: trait.DoubleEndedStream.html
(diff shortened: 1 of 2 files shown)


src/stream/double_ended_stream.rs
```diff
@@ -0,0 +1,22 @@
...
+pub trait DoubleEndedStream: Stream {
+    /// Removes and returns an element from the end of the stream.
+    ///
+    /// Returns `None` when there are no more elements.
+    ///
+    /// The [trait-level] docs contain more details.
+    ///
+    /// [trait-level]: trait.DoubleEndedStream.html
+    fn poll_next_back(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>>;
+}
```
stjepang: I think we'd also want to have `async fn next_back`, just like we have `Stream::poll_next` and `Stream::next`.

yoshuawuyts: For sure, but was thinking we make that part of the follow-up?

yoshuawuyts:
> By the way, I wonder -- do you perhaps have an example use case for double-ended streams?

Not anything concretely; but it generally seems like a nice thing to have.

One could imagine having a stream of tables queried from a database. In which case sometimes we want to go forward, but sometimes also backwards. This would allow us to implement that conveniently.

src/stream/double_ended_stream.rs
stjepang: Sure!

yoshuawuyts:
Implemented the feedback and marked this API as unstable so we get some wiggle room to change things even after 1.0.

PR #178, opened 2019-09-10
adds stream::enumerate combinator

montekki:
enumerate might be handy.
---
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate
Ref: #129

### DIFF
--- a/src/stream/stream/enumerate.rs
+++ b/src/stream/stream/enumerate.rs
@@ -21,7 +21,7 @@ impl<S> Enumerate<S> {
 
 impl<S> futures_core::stream::Stream for Enumerate<S>
 where
-    S: Stream + Unpin + Sized,
+    S: Stream,
 {
     type Item = (usize, S::Item);
 
--- a/src/stream/stream/enumerate.rs
+++ b/src/stream/stream/enumerate.rs
@@ -1,6 +1,8 @@
 use crate::task::{Context, Poll};
 use std::pin::Pin;
 
+use crate::stream::Stream;
+
 #[doc(hidden)]
 #[allow(missing_debug_implementations)]
 pub struct Enumerate<S> {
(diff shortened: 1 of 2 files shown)


src/stream/stream/mod.rs
```diff
@@ -136,6 +138,36 @@ pub trait Stream {
...
+    /// let mut s = stream::repeat(9).take(4).enumerate();
+    /// let mut c: usize = 0;
+    ///
+    /// while let Some((i, v)) = s.next().await {
+    ///     assert_eq!(c, i);
+    ///     assert_eq!(v, 9);
+    ///     c += 1;
+    /// }
+    /// #
+    /// # }) }
```
yoshuawuyts: This is a great test! But I feel like the example from `Iterator::enumerate` was a little clearer in its intent:
```rust
let a = ['a', 'b', 'c'];

let mut iter = a.iter().enumerate();

assert_eq!(iter.next(), Some((0, &'a')));
assert_eq!(iter.next(), Some((1, &'b')));
assert_eq!(iter.next(), Some((2, &'c')));
assert_eq!(iter.next(), None);
```

Perhaps manually calling `next().await` might make it easier to follow?

montekki: yep, makes sense

src/stream/stream/enumerate.rs
```diff
@@ -0,0 +1,40 @@
...
+    pin_utils::unsafe_unpinned!(i: usize);
+
+    pub(super) fn new(stream: S) -> Self {
+        Enumerate { stream, i: 0 }
+    }
+}
+
+impl<S> futures_core::stream::Stream for Enumerate<S>
+where
+    S: Stream + Unpin + Sized,
```
stjepang:
```suggestion
    S: Stream,
```

PR #180, opened 2019-09-11
adds stream::fold combinator

montekki:
Fold. Kind of clumsy around the part with the option and moving out of the shared context.
___
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold
Ref: #129

### DIFF
--- a/src/stream/stream/fold.rs
+++ b/src/stream/stream/fold.rs
@@ -1,0 +1,61 @@
+use std::marker::PhantomData;
+use std::pin::Pin;
+
+use crate::future::Future;
+use crate::task::{Context, Poll};
+
+#[doc(hidden)]
+#[allow(missing_debug_implementations)]
+pub struct FoldFuture<S, F, T, B> {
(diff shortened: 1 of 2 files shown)


src/stream/stream/fold.rs
```diff
@@ -0,0 +1,53 @@
...
+    F: FnMut(B, S::Item) -> B,
+{
+    type Output = B;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let next = futures_core::ready!(self.as_mut().stream().poll_next(cx));
+
+        match next {
+            Some(v) => {
+                cx.waker().wake_by_ref();
```
stjepang: Instead of waking ourselves on every step, let's turn this into a loop instead and implement `poll()` like here:

https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-util/src/stream/fold.rs

Issue #175, opened 2019-09-10
[tracking] futures concurrency macros


yoshuawuyts:
As per [my blog post](https://blog.yoshuawuyts.com/futures-concurrency/), I'd like us to try out futures concurrency macros. These should be introduced as "experimental" because they're mostly new, but I think this would make for a good starting point.

I propose these macros would live under `future::*` similar to `task::ready` in #133.

__update: these macros are now available as unstable since the 0.99.7 release. Please post any feedback about it in this issue.__

## Macros
- [x] `future::join`
- [x] `future::try_join`
- [x] `future::select`
- [x] `future::try_select`

yoshuawuyts:
We now have these, and I think it's okay to close this. Thanks!

PR #187, opened 2019-09-12
add future::{join,try_join,select,try_select} macros

### LINKED ISSUE #14: Add Future::join
In the [2016 futures announcement post](https://aturon.github.io/blog/2016/08/11/futures/), the `join` combinator was shown as something that would make choosing between several futures easy.

In https://github.com/rust-lang-nursery/futures-rs/issues/1215 and beyond this was changed to several methods: `join`, `join1`, `join2`, `join3`, and `join!`, `try_join` macros (proc macros since `0.3.0-alpha.18` so statements can be written inline).

## `Future::join`
It still seems incredibly useful to be able to join multiple futures together, and having a single combinator to do that seems like the simplest API, even if resulting code might not look completely symmetrical. I propose we add [`Future::join`](https://docs.rs/futures-preview/0.3.0-alpha.17/futures/future/fn.join.html):

```rust
use async_std::future;

let a = future::ready(1);
let b = future::ready(2);
let pair = a.join(b);

assert_eq!(pair.await, (1, 2));
```
## `Future::try_join`
The `futures-preview` library also exposes a [`try_join`](https://docs.rs/futures-preview/0.3.0-alpha.17/futures/future/fn.try_join.html) method. This is useful when you want to unwrap two results. Internally it uses `TryFuture` as a reference, which means this method should only exist on futures where `Output = Result<T, E>`, and I'm not entirely sure if that's feasible. However if it is it might be convenient to also expose:

```rust
use async_std::future;

let a = future::ready(Ok::<i32, i32>(1));
let b = future::ready(Ok::<i32, i32>(2));
let pair = a.try_join(b);

assert_eq!(pair.await, Ok((1, 2)));
```

## `Future::join_all`

The third join combinator present is [`Future::join_all`](https://docs.rs/futures-preview/0.3.0-alpha.17/futures/future/fn.join_all.html). The docs don't make a big sell on them (inefficient, set can't be modified after polling started, prefer `futures_unordered`), but it's probably still worth mentioning. I don't think we should add this combinator, but instead point people to use `fold` instead:

__don't do this__
```rust
use async_std::future::join_all;

async fn foo(i: u32) -> u32 { i }
let futures = vec![foo(1), foo(2), foo(3)];

assert_eq!(join_all(futures).await, [1, 2, 3]);
```

__do this instead__
```rust
let futures = vec![foo(1), foo(2), foo(3)];
let futures = futures.fold(|p, n| p.join(n))
assert_eq!(futures.await, [1, 2, 3]);
```
_note: not tested this, but in general I don't think we need to worry about this case too much as handling the unordered case seems much more important and would cover this too._

yoshuawuyts:
Adds the `future::{join,try_join}` macros. Depends on publishing a first release of https://github.com/async-rs/async-macros first (@stjepang could I get publish access to crates.io for `async-macros`?). Thanks!

Closes #14, #175.

## Screenshot

This is what the docs look like:

Screenshot_2019-09-12 async_std future - Rust

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ docs = []
 unstable = []
 
 [dependencies]
-async-macros = { path = "../async-macros" }
+async-macros = "1.0.0"
 async-task = "1.0.0"
 cfg-if = "0.1.9"
 crossbeam-channel = "0.3.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ docs = []
 unstable = []
 
 [dependencies]
+async-macros = { path = "../async-macros" }
 async-task = "1.0.0"
 cfg-if = "0.1.9"
 crossbeam-channel = "0.3.9"
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -3,6 +3,10 @@
 #[doc(inline)]
 pub use std::future::Future;
 
+#[doc(inline)]
+#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
+pub use async_macros::{join, try_join};
+
 use cfg_if::cfg_if;
 
 pub use pending::pending;
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -5,7 +5,7 @@ pub use std::future::Future;
 
 #[doc(inline)]
 #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
-pub use async_macros::{join, try_join};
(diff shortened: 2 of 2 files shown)


yoshuawuyts:
Updated with `select` and `try_select`!

Screenshot_2019-09-14 async_std future - Rust

yoshuawuyts:
This PR is ready for review now that `async-macros` has been published! :tada:

yoshuawuyts:
We now have `stream::join` as module-level macro and marked as unstable. I think it should be okay to merge this to let people try this out in the next release. Thanks!

PR #192, opened 2019-09-14
Add Stream::scan

tirr-c:
Ref #129. The mapper function `f` is synchronous and returns bare `Option<B>`.

Asynchronous `f` seems tricky to implement right. It requires the wrapper to be self-referential, as a reference to internal state may be captured by the returned future.

### DIFF
--- a/src/stream/stream/mod.rs
+++ b/src/stream/stream/mod.rs
@@ -29,8 +29,10 @@ mod find_map;
 mod min_by;
 mod next;
 mod nth;
+mod scan;
 mod take;
 
+pub use scan::Scan;
 pub use take::Take;
 
 use all::AllFuture;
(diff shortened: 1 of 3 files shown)


src/stream/stream/mod.rs
```diff
@@ -501,6 +503,49 @@ pub trait Stream {
...
+    /// assert_eq!(s.next().await, None);
+    /// #
+    /// # }) }
+    /// ```
+    #[inline]
+    fn scan<St, B, F>(self, initial_state: St, f: F) -> Scan<Self, St, F>
+    where
+        Self: Sized,
+        St: Unpin,
+        F: Unpin + FnMut(&mut St, Self::Item) -> Option<B>,
```
stjepang: We don't need `St` nor `F` to be `Unpin` because we're never pinning them.

stjepang:
Thanks!

bors r+

PR #194, opened 2019-09-14
Add doc comment for JoinHandle drop behavior

### LINKED ISSUE #143: Document behaviour of dropping JoinHandle
It appears that dropping an `async_std::task::JoinHandle` will currently detach the task and allow it to keep running. Presumably this is intended behaviour in order to match `std::thread::JoinHandle`, it should probably be documented with a paragraph like that from `std`:

> A `JoinHandle` _detaches_ the associated thread when it is dropped, which means that there is no longer any handle to thread and no way to `join` on it.

nonnontrivial:
fixes #143

### DIFF
--- a/src/task/task.rs
+++ b/src/task/task.rs
@@ -46,6 +46,9 @@ impl fmt::Debug for Task {
 
 /// A handle that awaits the result of a task.
 ///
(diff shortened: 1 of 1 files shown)


stjepang:
Thanks!

bors r+

PR #198, opened 2019-09-16
Remove Unpin bound in `impl Stream for T`

tirr-c:
I guess this is not needed anymore since we have `poll_next` that receives `Pin` now. The original bound is moved to `Stream::next`.


stjepang:
That's a good catch, thanks!

bors r+

PR #203, opened 2019-09-16
expose std::pin

yoshuawuyts:
This is important when defining / calling futures, so it makes sense for us to also export this.

But also given recent user feedback on the confusion on pinning, I'd like to open up the possibility to experiment with providing better pinning facilities  such as [`pin-project`](https://github.com/taiki-e/pin-project) or [`pin_mut`](https://docs.rs/pin-utils/0.1.0-alpha.4/pin_utils/macro.pin_mut.html) behind flags. I'm not sure if we could, or even should. But I want to allow us to have that conversation and test things out (even if it's just in floating patches.)

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,6 +53,7 @@ pub mod sync;
 pub mod task;
 
 #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
+#[cfg(feature = "unstable")]
 pub mod pin;
 
 pub(crate) mod utils;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,5 +51,6 @@ pub mod prelude;
 pub mod stream;
 pub mod sync;
 pub mod task;
+pub mod pin;
 
 pub(crate) mod utils;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,10 +47,12 @@ pub mod future;
 pub mod io;
 pub mod net;
 pub mod os;
-pub mod pin;
 pub mod prelude;
 pub mod stream;
 pub mod sync;
 pub mod task;
+
+#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
+pub mod pin;
 
 pub(crate) mod utils;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,10 +47,10 @@ pub mod future;
 pub mod io;
 pub mod net;
 pub mod os;
+pub mod pin;
 pub mod prelude;
 pub mod stream;
 pub mod sync;
 pub mod task;
-pub mod pin;
 
 pub(crate) mod utils;
--- a/src/pin.rs
+++ b/src/pin.rs
@@ -1,0 +1,4 @@
+//! Types that pin data to its location in memory.
+
+#[doc(inline)]
+pub use std::pin::Pin;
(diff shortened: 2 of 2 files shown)


stjepang:
The module is not behind the `unstable` flag :)

yoshuawuyts:
> The module is not behind the unstable flag :)

Oh yeah accurate; I was thinking `pin::Pin` is fine exactly the way it is because it's part of std and we already use it. Any addition to this module would be behind an `unstable` flag.

stjepang:
I'm not that sure we should expose `std::pin` yet and would still prefer having the whole module behind `unstable`.

> This is important when defining / calling futures, so it makes sense for us to also export this.

Defining and calling futures is something users should rarely do (ideally, never!) - after all, the promise of async/await is that we don't write futures by hand anymore. And even if we still do write them by hand from time to time today, I expect that to be less and less the case over time.

Another point is that pinning is not that much related to async programming, and is an orthogonal feature that has more to do with generators. But I admit that, practically speaking, pinning is very closely tied to futures right now.

There are also `std::marker::Unpin` and `std::marker::PhantomPinned`, without which the "pinning story" would be incomplete. Imagine someone having the following imports in their code:

```rust
use std::marker::PhantomPinned;
use async_std::pin::Pin;
```

Why is `Pin` here in `async_std` but `PhantomPinned` is not? I worry that might be confusing.

yoshuawuyts:
@stjepang those are all valid points, and I agree. I'll put the whole module behind an unstable flag!

src/lib.rs
```diff
@@ -52,4 +52,7 @@ pub mod stream;
 pub mod sync;
 pub mod task;
 
+#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
```
taiki-e: This probably appears properly in the documentation, but it seems that you can actually access this module without the unstable feature. (it needs to add `#[cfg(feature = "unstable")]`)

src/lib.rs
```diff
@@ -52,4 +52,8 @@ pub mod stream;
...
+#[cfg(feature = "unstable")]
+pub mod pin;
```
taiki-e: Also, I think it's better to re-export the `std::pin` module itself instead of defining our own module. `std::pin` module documentation is great (and proper use of Pin API without `std::pin` module documentation is difficult).

yoshuawuyts: I've added a link to `std::pin`'s docs. I don't think we should directly re-export `std::pin` in full because we may want to consider adding more things to it in the submodule in the future that currently aren't part of std. This should hopefully make for a good enough middle ground.

src/lib.rs
yoshuawuyts: Done!

src/lib.rs
taiki-e: Sounds good to me.

yoshuawuyts:
all feedback implemented. bors r+

taiki-e:
@yoshuawuyts: It seems bors could not recognize your comment as a command.

https://bors.tech/documentation/
> Also, the command will be recognized if, and only if, the word “bors” is at the beginning of a line.

stjepang:
Would it make sense to experiment with `pin_mut!` or `#[pin_project]` in `async-macros`/`async-attributes`? That feels to like a more appropriate place to me than `async-std` since those macros are not really "async version of `Pin`".

I'd love to collect all kinds of nice stuff there like `#[test]`, `#[bench]`, `#[main]`, `#[pin_project]`, `#[throws]`, `#[async_trait]`, etc. And we would be able to iterate quickly without worrying about stability - breaking changes are perfectly fine because these attributes are just syntax sugar.

yoshuawuyts:
@stjepang Hell yes!

PR #204, opened 2019-09-17
Add Stream::zip

tirr-c:


src/stream/stream/zip.rs
```diff
@@ -0,0 +1,54 @@
...
+impl<A: fmt::Debug + Stream, B: fmt::Debug> fmt::Debug for Zip<A, B> {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt.debug_struct("Zip")
+            .field("first", &self.first)
+            .field("second", &self.second)
+            .finish()
+    }
+}
+
+impl<A: Unpin + Stream, B: Unpin> Unpin for Zip<A, B> {}
```
stjepang:
```suggestion
impl<A: Unpin, B: Unpin> Unpin for Zip<A, B> {}
```

src/stream/stream/mod.rs
```diff
@@ -545,6 +547,49 @@ pub trait Stream {
...
+    ///
+    /// ```
+    /// # fn main() { async_std::task::block_on(async {
+    /// #
+    /// use std::collections::VecDeque;
+    /// use async_std::stream::Stream;
+    ///
+    /// let l: VecDeque<isize> = vec![1, 2, 3].into_iter().collect();
+    /// let r: VecDeque<isize> = vec![4, 5, 6, 7].into_iter().collect();
+    /// let mut s = l.zip(r);
```
stjepang: I'm a bit confused - how does this work? Are `l` and `r` streams or `VecDeque`s here?

tirr-c: `VecDeque`s implement `futures::stream::Stream` so they work as streams.

src/stream/stream/zip.rs
tirr-c: It doesn't work because of `A: Stream` bound on `struct Zip`.

yoshuawuyts:
Ref #129

PR #205, opened 2019-09-17
Implement simple work stealing

stjepang:
This is our first version of a work-stealing scheduler. We won't stop here, there is still lots of room for improvement.

### DIFF
--- a/src/task/builder.rs
+++ b/src/task/builder.rs
@@ -1,0 +1,32 @@
+use super::pool;
+use super::JoinHandle;
+use crate::future::Future;
+use crate::io;
+
+/// Task builder that configures the settings of a new task.
+#[derive(Debug)]
+pub struct Builder {
+    pub(crate) name: Option<String>,
+}
+
+impl Builder {
+    /// Creates a new builder.
+    pub fn new() -> Builder {
+        Builder { name: None }
+    }
+
+    /// Configures the name of the task.
+    pub fn name(mut self, name: String) -> Builder {
+        self.name = Some(name);
+        self
+    }
+
+    /// Spawns a task with the configured settings.
+    pub fn spawn<F, T>(self, future: F) -> io::Result<JoinHandle<T>>
+    where
+        F: Future<Output = T> + Send + 'static,
+        T: Send + 'static,
+    {
+        Ok(pool::get().spawn(future, self))
+    }
+}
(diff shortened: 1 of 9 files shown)


src/task/worker.rs
```diff
@@ -0,0 +1,110 @@
...
+        Ok(None) | Err(_) => None,
+    }
+}
+
+thread_local! {
+    static IS_WORKER: Cell<bool> = Cell::new(false);
+    static QUEUE: Cell<Option<Worker<task::Runnable>>> = Cell::new(None);
+}
+
+pub fn is_worker() -> bool {
```
yoshuawuyts:
```suggestion
pub(crate) fn is_worker() -> bool {
```

This does not seem to be re-exported as pub in the top-level.

src/task/pool.rs
```diff
@@ -64,130 +39,100 @@ where
...
 }
 
-/// Task builder that configures the settings of a new task.
-#[derive(Debug)]
-pub struct Builder {
-    pub(crate) name: Option<String>,
+pub(crate) struct Pool {
+    pub injector: Injector<task::Runnable>,
+    pub stealers: Vec<Stealer<task::Runnable>>,
+    pub sleepers: Sleepers,
```
yoshuawuyts: These fields should probably be marked `pub(crate)`.

src/task/builder.rs
```diff
@@ -0,0 +1,32 @@
+use super::pool;
+use super::JoinHandle;
+use crate::future::Future;
+use crate::io;
+
+/// Task builder that configures the settings of a new task.
+#[derive(Debug)]
+pub struct Builder {
+    pub(crate) name: Option<String>,
```
vertexclique: Will you add lifecycle callbacks here? like `after_start`, `before_stop`?

stjepang: Do you mean lifecycles callbacks for tasks or for worker threads?

stjepang: Do you mean lifecycles callbacks for tasks or for worker threads?

vertexclique: I think tokio has them for worker threads. What about doing it for `tasks` here?

stjepang: Yes, Tokio has them for worker threads. It doesn't make much sense for tasks because you can do:

```rust
let future = async move {
    // do stuff before
    let res = future.await;
    // do stuff after
    res
};
```

src/task/pool.rs
yoshuawuyts: We can always do this later.

PR #211, opened 2019-09-18
Add stream::Extend

tirr-c:


stjepang:
Thank you for another PR!

While this is a very logical addition, I'm feeling kind of unsure of the utility of this trait. The `std::iter::Extend` trait is implemented for synchronous collections only and is not used for I/O in the standard library at all. So I wonder whether we should have an async equivalent or not...

Another (minor!) concern is the name of the trait. It's a bit unfortunate that the trait name is `Extend`, while the method name is `extend_with_stream`. I wish the names matched, but the `ExtendWithStream` trait name does not feel amazing either. :) Maybe it's okay though, idk!

@yoshuawuyts Do you have thoughts on this?

yoshuawuyts:
> While this is a very logical addition, I'm feeling kind of unsure of the utility of this trait.

It's worth noting that Rayon has [`ParallelExtend`](https://docs.rs/rayon/1.2.0/rayon/iter/trait.ParallelExtend.html) trait too. I see this as a useful addition to `FromStream`.

In terms of uses outside of `std::collections`, you could imagine having a kv database that once opened can be "extended" by taking a stream of key-value tuples. There's probably some really interesting API designs that could be done, and I think accepting this PR will allow people to experiment with them.

---

> It's a bit unfortunate that the trait name is `Extend`, while the method name is `extend_with_stream`.

Rayon calls their method `par_extend`. I think following their lead and calling it `stream_extend` might be a suitable fit. Alternatively we keep calling it `extend` and let people calling it deal with specifying *which* extend they want to use. I don't love this, but I think it's acceptable.

[playground example of `Extend` conflicts](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c16c4702543df2e905129baaaacd0976)

src/stream/extend.rs
```diff
@@ -0,0 +1,50 @@
...
+///
+/// ```
+/// # fn main() { async_std::task::block_on(async {
+/// #
+/// use async_std::prelude::*;
+/// use async_std::stream::{self, Extend};
+///
+/// let mut v: Vec<usize> = vec![1, 2];
+/// let s = stream::repeat(3usize).take(3);
+/// v.extend_with_stream(s).await;
```
yoshuawuyts:
```suggestion
/// v.stream_extend(s).await;
```

src/stream/extend.rs
```diff
@@ -0,0 +1,50 @@
...
+///
+/// assert_eq!(v, vec![1, 2, 3, 3, 3]);
+/// #
+/// # }) }
+/// ```
+#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
+pub trait Extend<A> {
+    /// Extends a collection with the contents of a stream.
+    fn extend_with_stream<'a, T: IntoStream<Item = A> + 'a>(
```
yoshuawuyts:
```suggestion
    fn stream_extend<'a, T: IntoStream<Item = A> + 'a>(
```

src/stream/extend.rs
```diff
@@ -0,0 +1,50 @@
...
+        &'a mut self,
+        stream: T,
+    ) -> Pin<Box<dyn Future<Output = ()> + 'a>>;
+}
+
+impl Extend<()> for () {
+    fn extend_with_stream<'a, T: IntoStream<Item = ()> + 'a>(
```
yoshuawuyts:
```suggestion
    fn stream_extend<'a, T: IntoStream<Item = ()> + 'a>(
```

src/vec/extend.rs
```diff
@@ -0,0 +1,19 @@
+use std::pin::Pin;
+
+use crate::future::Future;
+use crate::stream::{Extend, IntoStream, Stream};
+
+impl<T> Extend<T> for Vec<T> {
+    fn extend_with_stream<'a, S: IntoStream<Item = T> + 'a>(
```
yoshuawuyts:
```suggestion
    fn stream_extend<'a, S: IntoStream<Item = T> + 'a>(
```

yoshuawuyts:
Ref #129

tirr-c:
@stjepang I decided to add `Extend` since `Iterator::unzip` in std requires `Extend`, but I now think `unzip` could be implemented with `FromStream` bound. I'll open up a PR soon!

stjepang:
Note that `Extend` is re-exported in `std`'s prelude: https://doc.rust-lang.org/nightly/std/prelude/v1/index.html

Should we likewise re-export our `Extend` in `async_std`'s prelude? And if so, should we rename the trait to `StreamExtend` in order not to conflict with `std`'s `Extend` and to align the name more closely with the method name `stream_extend()`?

yoshuawuyts:
> Should we likewise re-export our Extend in async_std's prelude? And if so, should we rename the trait to StreamExtend in order not to conflict with std's Extend and to align the name more closely with the method name stream_extend()?

I think yes on the former, and no on the latter. The name conflicting only becomes a problem when wanting to implement it, which I don't think this is for (as long as we export it as `_`). It's more so people can use it, and the method name is unique already.

PR #213, opened 2019-09-18
IntoStream for vec

yoshuawuyts:
This WIP PR implements `IntoStream` for `Vec`, adding three new stream types for vec: `vec::Stream`, `vec::IntoStream`, `vec::StreamMut`. We may want to extend this patch to also cover `slice`, because if we'd follow std, vec should use `Stream` and `StreamMut` types from `slice::*`.

Ref #129.

cc/ @stjepang @sunjay 

## Error
I'm currently blocked on a problem with our bounds. Namely we do:

```rust
impl<I: Stream + Send> IntoStream for I {
    type Item = I::Item;
    type IntoStream = I;

    #[inline]
    fn into_stream(self) -> I {
        self
    }
}
```

Which is causing the following error to show up:

2019-09-18-234517_1920x1080

I'm confused why this is happening, as `std::iter` has [a similar impl](https://doc.rust-lang.org/src/core/iter/traits/collect.rs.html#241-248). I'm worried this is some kind of orphan rule thing, but I don't quite get it. Getting help with this would be greatly appreciated!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,7 +52,7 @@ mod result;
 pub mod stream;
 pub mod sync;
 pub mod task;
-mod vec;
+pub mod vec;
 
 #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
 #[cfg(feature = "unstable")]
--- a/src/stream/into_stream.rs
+++ b/src/stream/into_stream.rs
@@ -25,12 +25,12 @@ pub trait IntoStream {
     fn into_stream(self) -> Self::IntoStream;
 }
 
-// impl<I: Stream + Send> IntoStream for I {
-//     type Item = I::Item;
-//     type IntoStream = I;
+impl<I: Stream + Send> IntoStream for I {
+    type Item = I::Item;
+    type IntoStream = I;
 
-//     #[inline]
-//     fn into_stream(self) -> I {
-//         self
-//     }
-// }
+    #[inline]
+    fn into_stream(self) -> I {
+        self
+    }
+}
--- a/src/stream/into_stream.rs
+++ b/src/stream/into_stream.rs
@@ -25,12 +25,12 @@ pub trait IntoStream {
     fn into_stream(self) -> Self::IntoStream;
 }
 
-impl<I: Stream + Send> IntoStream for I {
-    type Item = I::Item;
-    type IntoStream = I;
+// impl<I: Stream + Send> IntoStream for I {
+//     type Item = I::Item;
+//     type IntoStream = I;
 
-    #[inline]
-    fn into_stream(self) -> I {
-        self
-    }
-}
+//     #[inline]
+//     fn into_stream(self) -> I {
+//         self
+//     }
+// }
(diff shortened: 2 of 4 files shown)


jamesmunns:
@yoshuawuyts is this a specialization problem? e.g. https://github.com/rust-lang/rust/issues/31844 ?

Is this because `Vec<T>` could be `Stream`, so there would be two implementations, which means both of these impls would match (for example):

```
  --> src/vec/into_stream.rs:10:1
   |
10 | impl<T: Send> crate::stream::IntoStream for Vec<T> {
```

```
  ::: src/stream/into_stream.rs:28:1
   |
28 | impl<I: Stream + Send> IntoStream for I {
```

yoshuawuyts:
okay, I'm very confused by this. Rayon has [a similar impl as we do](https://docs.rs/rayon/1.2.0/src/rayon/iter/mod.rs.html#2071-2078), and they're not having the same problem. It seems like this may be something rather nuanced.

yoshuawuyts:
From conversations over chat, the next thing we should test is if it matters that we're using a foreign trait as part of the trait bound. Rayon doesn't do this, and we should check if only to cross it off.

sunjay:
The thing that confuses me about this is that the following code [compiles](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=37ed52a8234bd7717845fdd8ccb08a22):

```rust
trait IntoStream {}
trait Stream {}

impl<T: Send> IntoStream for Vec<T> {}
impl<T: Stream + Send> IntoStream for T {}
```

These are *exactly* the same impls as what you seem to have. This is also exactly the same as what `rayon` has and it works for them.

This is probably not the issue at all, but have you tried removing the `crate::stream` prefixes from `IntoStream`? I'm getting into the "it's a compiler bug" mindset and wondering if that will fix it for some odd reason. If that doesn't work (and it shouldn't!) I can probably look into this more in a day or so and see if I can get it to work.

taiki-e:
This behavior is not a bug. For example, if `Vec` comes to implement `futures_core::stream::Stream`, there is a conflict, so it is correct that this is a compilation error.

Refs: https://github.com/rust-lang/rust/pull/43426

sunjay:
Oh I didn't realize that Stream isn't defined in this crate. That's my bad then. Yes this isn't a bug.

yoshuawuyts:
Even when working around this by making the bound our own `Stream` type in #214, this rule still seems to be triggered (which is good!) but bad news for us. @taiki-e is entirely right, and if we want to keep compat out of the box between `async-std` and `futures-rs` then the only way to add this feature is by moving the implementation to `futures-rs`; possibly even `futures-core` because that is where the `Stream` trait is defined.

Conversation about this has been happening in https://github.com/rust-lang-nursery/futures-rs/issues/1879, but perhaps we should have a dedicated issue for this?

yoshuawuyts:
Closing this PR because it isn't possible for all the reasons @taiki-e has listed. I've posted a summary of this PR in the streams tracking issue: https://github.com/async-rs/async-std/issues/129#issuecomment-533093186.

My hope is that we can add `IntoStream` to `futures-core` so this can be possible. Thanks all!

PR #214, opened 2019-09-19
add crate::stream::Stream bound

yoshuawuyts:
This is an attempt to fix the problems in #213. Put out as a separate branch because we're doing some rather gnarly things here and we may want to try a few approaches. Thanks!


yoshuawuyts:
Looks like https://github.com/async-rs/async-std/pull/214/commits/f5451c2e097a0091afcb70c4d4f4b0ac4986b5ef breaks our bounds again, because a generic `futures_rs::stream::Stream` -> `async_std::stream::Stream` bound makes this impossible again.

Good of the trait system, but this isn't great for us.

PR #221, opened 2019-09-21
adds stream::filter combinator

montekki:
Ref: #129 
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter

### DIFF
--- a/src/stream/stream/mod.rs
+++ b/src/stream/stream/mod.rs
@@ -37,6 +37,7 @@ mod scan;
 mod take;
 mod zip;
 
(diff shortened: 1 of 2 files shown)


src/stream/stream/mod.rs
```diff
@@ -44,6 +45,7 @@ pub use zip::Zip;
 use all::AllFuture;
 use any::AnyFuture;
 use enumerate::Enumerate;
+use filter::Filter;
```
stjepang: Let's export this type :)

PR #222, opened 2019-09-21
adds stream::skip combinator

montekki:
Ref: #129 
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip


src/stream/stream/mod.rs
```diff
@@ -51,6 +52,7 @@ use fold::FoldFuture;
 use min_by::MinByFuture;
 use next::NextFuture;
 use nth::NthFuture;
+use skip::Skip;
```
stjepang: Let's export this type :)

PR #223, opened 2019-09-21
adds stream::skip_while combinator

montekki:
Ref: #129 
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip_while

### DIFF
--- a/src/stream/stream/mod.rs
+++ b/src/stream/stream/mod.rs
@@ -33,6 +33,7 @@ mod min_by;
 mod next;
 mod nth;
 mod scan;
+mod skip_while;
 mod take;
 mod zip;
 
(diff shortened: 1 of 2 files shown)


src/stream/stream/mod.rs
```diff
@@ -51,6 +52,7 @@ use fold::FoldFuture;
 use min_by::MinByFuture;
 use next::NextFuture;
 use nth::NthFuture;
+use skip_while::SkipWhile;
```
stjepang: Let's export `SkipWhile` (see lines 40-43 above).

The general policy is that we don't re-export `Future` types because functions that return them will eventually become `async fn`s, while we do export `Stream` types.

montekki: ah, i see now, makes sense.

PR #226, opened 2019-09-21
adds stream::step_by combinator

montekki:
Ref: #129 
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.step_by


stjepang:
We need to resolve merge conflicts :)

PR #231, opened 2019-09-22
Fix error[E0008]: cannot bind by-move into a pattern guard

froody:
Fixes the following error:

Error: error[E0008]: cannot bind by-move into a pattern guard
   --> src/net/driver/mod.rs:207:17
    |
207 |             Err(err) if err.kind() == io::ErrorKind::WouldBlock => {}
    |                 ^^^ moves value into pattern guard
    |
    = help: add `#![feature(bind_by_move_pattern_guards)]` to the crate attributes to enable

### DIFF
--- a/src/net/driver/mod.rs
+++ b/src/net/driver/mod.rs
@@ -204,7 +204,7 @@ impl<T: Evented> Watcher<T> {
     {
         // If the operation isn't blocked, return its result.
         match f(self.source.as_ref().unwrap()) {
-            Err(err) if err.kind() == io::ErrorKind::WouldBlock => {}
+            Err(ref err) if err.kind() == io::ErrorKind::WouldBlock => {}
             res => return Poll::Ready(res),
         }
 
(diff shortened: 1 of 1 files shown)


taiki-e:
As `bind_by_move_pattern_guards ` stabilized about two weeks ago (https://github.com/rust-lang/rust/pull/63118), I believe updating the version of rustc should fix this error.

stjepang:
@froody what is your rustc version?

froody:
@stjepang 1.38, updating to 1.39 isn't really feasible right now

yoshuawuyts:
We haven't quite talked about this internally, but given nobody on the team has really moved to act on this PR I think there's a silent consensus that we're not targeting any versions of the compiler under 1.39.

We probably should still formulate our support strategy in the face of the upcoming stabilization of async/await. I suspect it'll be something like: "always work on stable; only conditionally support new nightly features".

Either way, I don't want to keep this PR open indefinitely so I'm going to go ahead and close this. Thanks heaps!

Issue #212, opened 2019-09-18
Design of async channels


stjepang:
It's time to port `crossbeam-channel` to futures.

Previous discussions:

- https://github.com/crossbeam-rs/crossbeam/issues/314
- https://github.com/crossbeam-rs/crossbeam-channel/issues/39
- https://github.com/crossbeam-rs/crossbeam-channel/issues/61

cc @matklad @BurntSushi @glaebhoerl

Instead of copying `crossbeam-channel`'s API directly, I'm thinking perhaps we should design async channels a bit differently.

In our previous discussions, we figured that dropped receivers should disconnect the channel and make send operations fail for the following reason. If a receiver thread panics, the sending side needs a way to stop producing messages and terminate. If dropped receivers disconnect the channel, the sending side will usually panic due to an attempt of sending a message into the disconnected channel.

In Go, sending into a channel is not a fallible operation even if there are no more receivers. That is because Go only uses bounded channels so they will eventually fill up and the sending side will then block on the channel, attempting to send another message into the channel while it's full. Fortunately, Go's scheduler has a deadlock detection mechanism so it will realize the sending side is deadlocked and will thus make the goroutine fail.

In `async-std`, we could implement a similar kind of deadlock detection: a task is deadlocked if it's sleeping and there are no more wakers associated with it, or if all tasks are suddenly put to sleep. Therefore, channel disconnection from the receiver side is not such a crucial feature and can simplify the API a lot.

If we were to have only bounded channels and infallible send operations, the API could look like this:

```rust
fn new(cap: usize) -> (Sender<T>, Receiver<T>);

struct Sender<T>;
struct Receiver<T>;

impl<T> Sender<T> {
    async fn send(&self, msg: T);
}

impl<T> Receiver<T> {
    fn try_recv(&self) -> Option<T>;
    async fn recv(&self) -> Option<T>;
}

impl<T> Clone for Sender<T> {}
impl<T> Clone for Receiver<T> {}

impl<T> Stream for Receiver<T> {
    type Item = Option<T>;
}
```

This is a very simple and ergonomic API that is easy to learn.

[...]

PR #232, opened 2019-09-22
Add Stream::unzip

tirr-c:
`Iterator::unzip` collects elements in `Default + Extend` containers, but here I used `FromStream` on bound channels with no capacity.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,8 @@ features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
-docs = []
-unstable = []
+docs = ["futures-channel-preview"]
+unstable = ["futures-channel-preview"]
 
 [dependencies]
 async-macros = "1.0.0"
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
Ref #129.

@tirr-c Thanks so much for starting this! Could you say more about why you chose to change the trait bounds? We've just merged #211, which makes it possible to use in the bounds.

Heh also def seems like this would pair up nicely with the channel design in #212! (though we shouldn't block ourselves on it I think).

tirr-c:
@yoshuawuyts I first thought that `Default + Extend` bounds meant it'll make a new container and extend from some iterator (or stream), so I used `FromStream` instead.

I've just checked out the implementation of `Iterator::unzip`. Surprisingly it was using `Extend` as an abstraction of element addition! So we can even use `Default + std::iter::Extend` for the container bounds.

yoshuawuyts:
> So we can even use `Default + std::iter::Extend` for the container bounds.

Not *necessarily* -- if `Extend` is implemented for some kind of database handle then `std::iter::Extend` would not be suitable. I think if we want people to be able to experiment with building networked streaming patterns we keep it as `Default + Stream::Extend`.

## Refs
- [`std::iter::Iterator::unzip` impl](https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#2184)

tirr-c:
If we're not given any existing database handle, we need to build a new handle from its `default` impl; that isn't what we want, I think. How about `unzip` returning two streams?

tirr-c:
This PR is superseded by #543.

PR #238, opened 2019-09-24
re-export async-attributes

### LINKED ISSUE #237: Expose async-attributes macros behind feature flag
For applications it's generally useful to have access to the attributes as defined in [`async-attributes`](https://docs.rs/async-attributes/1.0.0/async_attributes/). Being able to install a single dependency seems quite desirable. 

## Implementation
I propose we expose these attributes from the root of the crate, relying on a feature flag. As for the flag name, there's a few options. But the one I think I prefer is `"bin"` to indicate this should be used for binaries only.

```rust
#[cfg(feature = "bin")]
pub use async_attributes::*;
```

yoshuawuyts:
Closes #237. The docs don't render properly on this, but I think that's okay. Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ unstable = ["broadcaster"]
 attributes = ["async-attributes"]
 
 [dependencies]
-async-attributes = { version = "1.0.0", optional = true }
+async-attributes = { version = "1.1.0", optional = true }
 async-macros = "1.0.0"
 async-task = "1.0.0"
 broadcaster = { version = "0.2.6", optional = true, default-features = false, features = ["default-channels"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,9 @@ rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
 default = []
-docs = ["unstable"]
+docs = ["unstable", "attributes"]
 unstable = ["broadcaster"]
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
This PR and https://github.com/async-rs/async-attributes/pull/8 have been updated to now use the `attributes` feature as suggested in #438.

I think merging this would actually be really nice, as we enable people to write some nicer-looking applications the way they could with Runtime, but we don't lose any performance in the base case. Coupled with the `runtime` feature I think `async-std` will be a *really* well rounded option for almost every use!

stjepang:
Are these now stable attributes? I assume so and fine with that, but just making sure it's not an oversight. :)

There's not much we can do wrong with these attributes - they're pretty straightforward, right?

yoshuawuyts:
> There's not much we can do wrong with these attributes - they're pretty straightforward, right?

Not necessarily, and I think we should be cautious. https://github.com/async-rs/async-attributes/pull/11 and https://github.com/async-rs/async-attributes/issues/9 actually should be resolved first.

yoshuawuyts:
Updated the PR to be exactly that. Thanks!

stjepang:
Agree with everything you said 👍

stjepang:
ping for @skade

yoshuawuyts:
@stjepang I don't want to miss trying this out in the next release; I'd like to propose we merge this.

PR #239, opened 2019-09-25
v0.99.7

yoshuawuyts:
Prepares the 0.99.7 release. Intended to release tomorrow. Thanks!


yoshuawuyts:
Only rustdoc is broken (because nightly), which means we should be okay to merge. Going to cut a release now!

PR #240, opened 2019-09-25
feat: implement sync::Barrier

dignifiedquire:
~~Based on the implementation in https://github.com/tokio-rs/tokio/pull/1571~~

Based on the implementation in the std lib 

Things to discuss

- `async` or `std` `Arc` in examples and tests
- ~~`async` or `std` `Mutex` in the implementation~~ can't use std, as it needs to be shared
- use of `broadcaster` dependency to implement, well broadcast

To fix
- [x] missing debug implementation, once broadcaster has it, or it is dropped

## Docs

Screenshot 2019-09-25 at 21 07 25

### DIFF
--- a/src/sync/barrier.rs
+++ b/src/sync/barrier.rs
@@ -1,0 +1,174 @@
+use broadcaster::BroadcastChannel;
+
+use crate::sync::Mutex;
+
+/// A barrier enables multiple tasks to synchronize the beginning
+/// of some computation.
+///
+/// ```
+/// # fn main() { async_std::task::block_on(async {
+/// #
+/// use std::sync::Arc;
+/// use async_std::sync::Barrier;
+/// use async_std::task;
+///
+/// let mut handles = Vec::with_capacity(10);
+/// let barrier = Arc::new(Barrier::new(10));
+/// for _ in 0..10 {
+///     let c = barrier.clone();
+///     // The same messages will be printed together.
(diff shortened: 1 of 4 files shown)


src/sync/barrier.rs
```diff
@@ -0,0 +1,130 @@
...
+///     // You will NOT see any interleaving.
+///     handles.push(async move {
+///         println!("before wait");
+///         let wr = c.wait().await;
+///         println!("after wait");
+///         wr
+///     });
+/// }
+/// // Will not resolve until all "before wait" messages have been printed
+/// let wrs = join_all(handles).await;
```
yoshuawuyts: We make this example even more similar to [stdlib's](https://doc.rust-lang.org/std/sync/struct.Barrier.html#examples) by doing:

```rust
for handle in handles {
    handle.await;
}
```

This allows us to not have to use the `join_all` macro. It works especially well if we remove the `wr` result part from the example as well.

dignifiedquire: fixed

yoshuawuyts:
It looks like it's Cargo that's failing CI on MacOS. All other tests' pass, so I think this should be fine. Let's see if Bors accepts this.

bors r+

yoshuawuyts:
Looks like on this nightly `cargo doc` broke. Think this is okay to merge manually.

PR #241, opened 2019-09-26
Simplify extension traits using a macro

### LINKED ISSUE #235: stream doc tests are now disabled
Looks like they are no longer run since 73d7fea9 but they probably should be. At least in [CI](https://travis-ci.com/async-rs/async-std/jobs/237994862) they are gone.

stjepang:
This PR would fix #235 

Async methods in our extension traits are now written in the simpler `-> impl Future<Output = T> [ConcreteFuture<Self>]` style. At the same time, doc tests are used even when the `docs` feature is not enabled.

### DIFF
--- a/src/stream/stream/mod.rs
+++ b/src/stream/stream/mod.rs
@@ -503,7 +503,7 @@ extension_trait! {
         fn min_by<F>(
             self,
             compare: F,
-        )-> impl Future<Output = Option<Self::Item>> [MinByFuture<Self, F, Self::Item>]
+        ) -> impl Future<Output = Option<Self::Item>> [MinByFuture<Self, F, Self::Item>]
         where
             Self: Sized,
             F: FnMut(&Self::Item, &Self::Item) -> Ordering,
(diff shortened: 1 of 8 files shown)


src/stream/stream/mod.rs
```diff
@@ -78,1089 +88,952 @@ cfg_if! {
...
+    ///
+    /// [`std::iter::Iterator`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html
+    /// [`futures::stream::Stream`]:
+    /// https://docs.rs/futures-preview/0.3.0-alpha.17/futures/stream/trait.Stream.html
+    /// [provided methods]: #provided-methods
+    pub trait Stream [StreamExt: futures_core::stream::Stream] {
+        /// The type of items yielded by this stream.
+        type Item;
+
+        #[doc = r#"
```
stjepang: We have to write docs in `Stream` in this style because the macro will otherwise exceed the recursion limit and dramatically slow compilation down. That is because each line starting with `///` is a token itself, but with `#[doc = "..."]` we can put the whole docs for a function into a single token tree, meaning the macro makes progress faster - instead of consuming docs line by line it consumes them in a single bite.

PR #245, opened 2019-09-27
feat: missing Read and Write methods

dignifiedquire:
Ref: #131 

- [x] Read::by_ref
- [x] Read::bytes
- [x] Read::chain
- [x] Read::take
- [ ] Write::by_ref
- [ ] ~~Write::write_fmt~~ postponed until https://github.com/async-rs/async-std/issues/247 is solved

Needs fixing:

- [x] `BufRead` for `Take`
- [x] `BufRead` for `Chain`
- [ ] `by_ref` conflict between `Read` and `Write`, unable to add both, as they conflict, and the current state of things does not allow to differentiate between the two.

### DIFF
--- a/src/io/read/bytes.rs
+++ b/src/io/read/bytes.rs
@@ -1,0 +1,60 @@
+use std::pin::Pin;
+
+use crate::io::{self, Read};
+use crate::stream::stream::Stream;
+use crate::task::{Context, Poll};
+
+/// An iterator over `u8` values of a reader.
+///
+/// This struct is generally created by calling [`bytes`] on a reader.
+/// Please see the documentation of [`bytes`] for more details.
+///
+/// [`bytes`]: trait.Read.html#method.bytes
+#[derive(Debug)]
+pub struct Bytes<T> {
+    pub(crate) inner: T,
+}
+
+impl<T: Read + Unpin> Stream for Bytes<T> {
+    type Item = io::Result<u8>;
+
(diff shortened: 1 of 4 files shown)


dignifiedquire:
marking as ready, this can be reviewed and merged, and `Write::by_ref` figured out later

src/io/read/bytes.rs
```diff
@@ -0,0 +1,60 @@
+use std::pin::Pin;
+
+use crate::io::{self, Read};
+use crate::stream::stream::Stream;
+use crate::task::{Context, Poll};
+
+/// An iterator over `u8` values of a reader.
```
yoshuawuyts:
```suggestion
/// A stream of `u8` values from a reader.
```

src/io/read/mod.rs
```diff
@@ -261,6 +264,157 @@ extension_trait! {
...
+
+            [`File`]: ../fs/struct.File.html
+            [`Ok(0)`]: ../../std/result/enum.Result.html#variant.Ok
+            [`read()`]: tymethod.read
+
+            ```no_run
+            use async_std::io::prelude::*;
+            use async_std::fs::File;
+
+            fn main() -> std::io::Result<()> {
```
yoshuawuyts: Generally we've been hiding the `fn main` and closing blocks from the examples. Could you maybe go over them and make sure those lines are prefixed with `#`?

dignifiedquire: fixed

PR #251, opened 2019-09-27
add an unstable `task::blocking` function

### LINKED ISSUE #65: Make `blocking` public?

### LINKED ISSUE #219: Set up task::Tag in blocking tasks
Blocking tasks should behave similarly to other tasks - in particular, they should be nameable, have an id, and task-locals.

yoshuawuyts:
This add an unstable `task::spawn_blocking` function to unblock people (lol) waiting for #65. I'm not sure if this is the right API, but we've talked about adding `spawn_local` at some point in the future too, so it doesn't seem too wrong either.

Regardless, I think it's time to provide people with at least "unstable" access to this. Thanks!

Closes #219, closes #65 

cc/ @zkat 

## Screenshot

Screenshot_2019-09-28 async_std task - Rust

### DIFF
--- a/src/fs/read_dir.rs
+++ b/src/fs/read_dir.rs
@@ -5,7 +5,7 @@ use crate::fs::DirEntry;
 use crate::future::Future;
 use crate::io;
 use crate::stream::Stream;
-use crate::task::{blocking, Context, Poll};
+use crate::task::{blocking, Context, Poll, JoinHandle};
 
 /// Returns a stream of entries in a directory.
 ///
@@ -71,7 +71,7 @@ pub struct ReadDir(State);
 #[derive(Debug)]
 enum State {
     Idle(Option<std::fs::ReadDir>),
-    Busy(blocking::JoinHandle<(std::fs::ReadDir, Option<io::Result<std::fs::DirEntry>>)>),
+    Busy(JoinHandle<(std::fs::ReadDir, Option<io::Result<std::fs::DirEntry>>)>),
 }
 
 impl ReadDir {
--- a/src/fs/read_dir.rs
+++ b/src/fs/read_dir.rs
@@ -5,7 +5,7 @@ use crate::fs::DirEntry;
 use crate::future::Future;
 use crate::io;
 use crate::stream::Stream;
-use crate::task::{blocking, Context, Poll, JoinHandle};
+use crate::task::{blocking, Context, JoinHandle, Poll};
 
 /// Returns a stream of entries in a directory.
 ///
--- a/src/io/stderr.rs
+++ b/src/io/stderr.rs
@@ -5,7 +5,7 @@ use cfg_if::cfg_if;
 
 use crate::future::Future;
 use crate::io::{self, Write};
-use crate::task::{blocking, Context, Poll};
+use crate::task::{blocking, Context, Poll, JoinHandle};
 
 /// Constructs a new handle to the standard error of the current process.
 ///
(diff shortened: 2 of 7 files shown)


yoshuawuyts:
I still think we should merge this, but I'm almost wondering if the mapping of blocking -> future shouldn't be part of the thread submodule.

There's something about blocking tasks that makes me feel like it should be part of that. Not sure tho; feel like we may need to talk it through sometime.

skade:
I see the issue that this function will be _effectively_ stable. The proliferation of reaching for an `unstable` flag to expose things people are asking for will either force people to choose the unstable flag all the time and not caring about it and taking the ability from us to iterate on the solution.

We should instead spend more time on figuring out what a stable blocking interface should be.

dvc94ch:
Having to make some changes after running `cargo update` is better than having to maintain a fork, that is out of date. And isn't semver supposed to alleviate this? I don't understand all this commitment to find a solution that will be supported indefinitly.

dvc94ch:
After some reflection I see the problem. Whenever a crate exports a trait the crate can only occur once in the dependency tree.

zkat:
Yes please! Is this moving forward? :D

yoshuawuyts:
@skade We've been talking about this for _months_ now, and we haven't made any tangible progress on an API design. Currently there's quite a few other things in-progress, and several other things marked as priority before we could get around to designing this. Where I estimate it'll probably be at least another month or two before this can be prioritized, which imo is too long.

I think the higher order bit would be to unblock people who have been waiting on this feature for a long time already, so they can move forward and build projects on top of async-std. We can provide detailed upgrade instructions if we decide to stabilize a different version of the API.

cc/ @stjepang I'd be curious for your input here also.

stjepang:
I'm comfortable with stabilizing this function, provided we do two changes:

1. Rename to something simpler, perhaps `task::blocking()`.
2. Return a regular `task::JoinHandle<T>` from the function.

We should probably make blocking tasks behave just like normal tasks spawned using `task::spawn()`, except `task::blocking()` would give a hint to the scheduler that this task might block. That means inside blocking tasks we should be able to use task-locals, `task::current()` and so on.

yoshuawuyts:
@stjepang and I had a chat yesterday, and we're on the same page about what to do. I agree with what he's said in https://github.com/async-rs/async-std/pull/251#issuecomment-538843574.

yoshuawuyts:
Alright, this PR now exposes `task::blocking`, and implements #219 as requested. This should be good to review now!

yoshuawuyts:
I've addressed all feedback, and things are compiling + tests passing so I think this should be good to merge!

Issue #246, opened 2019-09-27
Use GitHub Actions


yoshuawuyts:
[actions-rs](https://github.com/actions-rs) by @svartalf has been looking fantastic, and I think we should consider using it instead of Travis.

For example today [actions-rs/clippy](https://github.com/actions-rs/clippy-check) was released, which places clippy checks inline in the PR review:

![screenshot](https://user-images.githubusercontent.com/2467194/65774660-ff5cbb00-e13e-11e9-8c1d-13a819039bf6.png)

I think some really good work is being done as part of `actions-rs`, and by using it we'll probably enjoy faster build times thanks to GitHub actions providing more cycles, and a chance to help Rust tooling development by being able to provide feedback / patches back to a very promising project.

The only open question I have is how this would integrate with Bors, if at all.

taiki-e:
> The only open question I have is how this would integrate with Bors, if at all.

Related: https://github.com/bors-ng/bors-ng/issues/730 (Doesn't seem to be officially supported)

yoshuawuyts:
Example of the clippy checks in action: https://github.com/actions-rs/example/pull/2/files.

k-nasa:
Isn't it possible to use it with such a change?
https://github.com/samueldple/date_time/pull/45/files#diff-2f8c9db9253b64f9a332c09f492cd54e

yoshuawuyts:
@k-nasa oh yeah, probably! -- is this something you'd like to try spinning up for our repo?

k-nasa:
I'll try it! Can you assigne me?

freebroccolo:
> Example of the clippy checks in action: https://github.com/actions-rs/example/pull/2/files.

[Here](https://github.com/freebroccolo/interfaces/blob/master/.github/workflows/main.yml) is an example of a configuration I'm using that might be useful.

The tarpaulin step takes the longest but there is a [PR](https://github.com/actions-rs/tarpaulin) at [actions-rs/tarpaulin](https://github.com/actions-rs/tarpaulin) that downloads the binary instead of building it and uses the tools cache. Should be much faster.

svartalf:
It is a closed issue, but I would like to add a late comment: as soon as Github will add proper caching for Actions (AFAIK, mid-November), we try to add caching for installed binaries too; so the first one `cargo install` will be slow as usual, but the next installations of the same version will use the pre-compiled binary from cache. That would reduce time for all builds, which are installing something.

taiki-e:
[...]

PR #253, opened 2019-09-28
Add github actions workflows

k-nasa:
## background
close: https://github.com/async-rs/async-std/issues/246

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,57 @@
+on: [push]
+
+jobs:
+  build_and_test:
+    name: Build adn test on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Install nightly
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: nightly
+        override: true
+
+    - name: check
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --all --benches --bins --examples --tests
+
+    - name: check unstable
+      uses: actions-rs/cargo@v1
+      with:
+        command:  check
+        args: --features unstable --all --benches --bins --examples --tests
+
+    - name: tests
+      uses: actions-rs/cargo@v1
+      with:
+        command: test
+        args: --all --doc --features unstable
+
+  check_fmt_and_docs:
+    name: Checking fmt and docs
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+
+    - name: setup
+      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
+on: [push]
+
+jobs:
+  build_and_test:
+    name: Build adn test on ${{ matrix.os }}
```
freebroccolo: Thanks for working on this! Just noticed a little typo here: the `adn` should be `and`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+
+    - name: setup
+      run: |
+        rustup default nightly
+        rustup component add rustfmt
+        cargo install mdbook
+        rustc --version
```
freebroccolo: Did you see the [install-mdbook](https://github.com/async-rs/async-std/blob/master/ci/install-mdbook.sh) script? That way saves quite a bit of time compared to the `cargo install mdbook`.

However, the script is fragile because it ultimately relies on a specific release filename pattern and it would be nice if we could have a more general solution. (This is currently an issue for another CI related PR [here](https://github.com/async-rs/async-std/pull/138)).

I think a better way might be to create a docker image based on the official image (or slim or docs.rs) with all of the tools we need (deadlinks, mdbook, tarpaulin?) pre-installed. Then we could use the [container.image setting](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idcontainerimage) to set the environment so that stuff is available without having to rebuild it.

Another github action could handle docker image related tasks.

.github/workflows/ci.yml
yoshuawuyts:
```suggestion
    name: Build and test on ${{ matrix.os }}
```

.github/workflows/ci.yml
yoshuawuyts: I agree this would be really nice, and it would probably worth doing this in a follow-up!

Perhaps it's worth opening an issue for this on https://github.com/actions-rs/meta/issues also; I agree that having a general solution for this would probably be very valuable to a lot of people.

skade:
Bors seems to hang.

.github/workflows/ci.yml
skade: `cargo install mdbook` should be avoided, it adds ~8 min of build time.

k-nasa: Wow, long.  I will fix it tonight, so take a moment

.github/workflows/ci.yml
k-nasa: Thanks 😘

k-nasa:
I fixed install mdbook. Please re-review 🙇

skade:
Bors breaks on this. I'll merge it in good faith to test.

PR #255, opened 2019-09-28
Update futures-preview to 0.3.0-alpha.19

taiki-e:
Depends on https://github.com/rustasync/surf/pull/74

Refs:
* https://github.com/rust-lang-nursery/futures-rs/issues/1761
* https://github.com/rust-lang-nursery/futures-rs/pull/1845

cc @skade @stjepang @yoshuawuyts

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,13 +46,10 @@ broadcaster = { version = "0.2.6", optional = true, default-features = false, fe
 
 [dev-dependencies]
 femme = "1.2.0"
-surf = "1.0.2"
+# surf = "1.0.2"
 tempdir = "0.3.7"
 futures-preview = { version = "=0.3.0-alpha.19", features = ["async-await"] }
 
 # These are used by the book for examples
 futures-channel-preview = "=0.3.0-alpha.19"
 futures-util-preview = "=0.3.0-alpha.19"
-
-[patch.crates-io]
-surf = { git = "https://github.com/taiki-e/surf", branch = "futures" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,8 @@ async-task = "1.0.0"
 cfg-if = "0.1.9"
 crossbeam-channel = "0.3.9"
 crossbeam-deque = "0.7.1"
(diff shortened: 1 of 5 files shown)


tomaka:
That would now be a breaking change and require a new "major" version of async-std, I guess?

taiki-e:
@tomaka Since the major version of async-std is 0, we will actually need to increase the minor version (if we follow semver).

Also, I believe https://github.com/async-rs/async-std/pull/239/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR39-R40 was a breaking change.
So my understanding is that async-std will not follow semver until 1.0.0 is released (Disclaimer: I'm not involved in these release processes, so I'm not sure about the exact thing).
(This actually occurs even in crate managed by rustasync such as surf.)

Cargo.toml
```diff
@@ -48,11 +48,11 @@ broadcaster = { version = "0.2.6", optional = true, default-features = false, fe
 femme = "1.2.0"
 surf = "1.0.2"
 tempdir = "0.3.7"
+futures-preview = { version = "=0.3.0-alpha.19", features = ["async-await"] }
```
skade: We can also drop the `async-await` features, we don't need it anymore.

taiki-e: I think examples/a-chat is still using it.
https://github.com/async-rs/async-std/blob/5881d9addef991e6d261a073f81c961ee3eb4bad/examples/a-chat/client.rs#L1

skade: Right, in this case, we should keep it.

skade:
@taiki-e I can't quite figure out why this is a draft, is there anything missing?

taiki-e:
@skade This PR is marked as a draft because it depends on other  PR (rustasync/surf#74, But, I'm not sure if this is the correct usage of the draft). Everything else is ready.

skade:
In this case, I think we should deactivate the surf example temporarily in this case and merge it.

taiki-e:
@skade Done.

taiki-e:
Oops, sorry, I forgot to remove the patch.

taiki-e:
Fixed and CI passed.

PR #258, opened 2019-09-28
Expose pin project

yoshuawuyts:
re-exports [`pin-project`](https://docs.rs/pin-project/0.4.1/pin_project/) traits from our unstable `pin` submodule. We should eventually pair this up with some good docs on what pinning is, why pin projection is useful, and how to do it using these attributes.

From internal discussions we've been considering adopting `pin-project` to remove our use of `unsafe` from our pin projections. It'd be great if we could write down our experiences porting from unsafe projections to pin-project, and use that to guide others to do the same. Thanks!

Related #203 

cc/ @taiki-e 

## Screenshot

Screenshot_2019-09-28 async_std pin - Rust

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://book.async.rs/overview
 
 ## [Unreleased]
 
+## Added
+
+- Added `pin::pin_project` attribute macro as unstable.
+- Added `pin::pinned_drop` attribute macro as unstable.
+- Added `pin::project` attribute macro as unstable.
+- Added `pin::project_ref` attribute macro as unstable.
+- Added `pin::UnsafeUnpin` trait as unstable.
+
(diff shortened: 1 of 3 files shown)


taiki-e:
I don't think re-exports will work at this time [because current pin-project requires that the crate name is exactly `::pin_project`.](https://github.com/taiki-e/pin-project/blob/ff227fed7927306c4d443ad98d08c0fb32bd9a30/pin-project-internal/src/pinned_drop.rs#L20)

I know [how to make this work well](https://github.com/taiki-e/futures-enum/blob/b1914fffbb0af92764bc603dea9352d430bbc227/src/lib.rs#L26) :), but that means blessing async-std, so I don't know if I really want to do it.

skade:
I would also hold back on this, to avoid accidental blessing.

yoshuawuyts:
Closing because we can't merge this now either way. I'm sympathetic to what's been said in this thread. Thanks!

PR #259, opened 2019-09-28
add IntoFuture

yoshuawuyts:
Adds the `IntoFuture` trait, wihch @nemo157 [pointed out](https://www.reddit.com/r/rust/comments/d7acl0/async_finalizers/f0ytyyj/) would be useful for implementing the "async builder" pattern we used when building Surf.

In the [RFC2394 async_await](https://github.com/rust-lang/rfcs/blob/master/text/2394-async_await.md#the-await-compiler-built-in) it's also stated that `await` should call `IntoFuture` under the hood.

> A builtin called await! is added to the compiler. `await!` can be used to "pause" the computation of the future, yielding control back to the caller. __`await!` takes any expression which implements IntoFuture__, and evaluates to a value of the item type that that future has.

This PR doesn't add that functionality to `await`, but does have the right shape it should take, which allows people to start implementing `IntoFuture` so that later it will "just work".

---

cc/ @withoutboats, is what I've said so far about `IntoFuture` still accurate? I didn't see any mentions that this had been changed in [the stabilization report](https://github.com/rust-lang/rust/issues/62149), so does that mean that it means this simply has not been implemented yet?

## Screenshot

Screenshot_2019-09-28 async_std future - Rust

### DIFF
--- a/src/future/into_future.rs
+++ b/src/future/into_future.rs
@@ -11,3 +11,13 @@ pub trait IntoFuture {
     /// Create a future from a value
     fn into_future(self) -> Self::Future;
 }
+
+impl<T: Future> IntoFuture for T {
+    type Output = T::Output;
+
+    type Future = T;
+
+    fn into_future(self) -> Self::Future {
+        self
+    }
+}
--- a/src/future/into_future.rs
+++ b/src/future/into_future.rs
@@ -1,6 +1,35 @@
 use crate::future::Future;
 
 /// Convert a type into a `Future`.
+///
+/// # Examples
+///
+/// ```
+/// use async_std::future::{Future, IntoFuture};
+/// use async_std::io;
+/// use async_std::pin::Pin;
+///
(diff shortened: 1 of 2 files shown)


stjepang:
Do you think it would make sense to open this PR in the rust-lang/rust repository? :) It seems to me the intention was for this trait to exist in the standard library, but nobody has implemented it there yet.

yoshuawuyts:
@stjepang yeah I think it does. I do think there's value in still merging the PR here though. I'll check with the lang team what they think and if they're okay with it I'll follow up with the PR to std.

withoutboats:
@stjepang @yoshuawuyts I think `IntoFuture` (and making `await` work with that) would be a great extension, but probably one that would require a (small) RFC.

yoshuawuyts:
There's now a PR to std for this also: https://github.com/rust-lang/rust/pull/65244.

PR #262, opened 2019-09-30
Async combinators

### LINKED ISSUE #244: async combinators for Stream
Our [stream combinators](https://docs.rs/async-std/0.99.7/async_std/stream/trait.Stream.html) operate on asynchronous data, but are themselves synchronous.

This means it's not possible to for example filter or map a stream based on other async operations. Which seems limiting, as it's possible to call sync functions from async functions but not the other way around.

I'd like to propose we update our Stream combinators that take closures to be async functions.

cc/ @montekki 
ref #129

montekki:
Partially fixes #244 (Probably that issue needs to be broken down in two: 1) for what can be done now 2) what can be done in the future like async closures and taking by ref.

Also unimplements `Unpin` for `all` and `any` combinators becasue now we have a future cached that may not be unpin.


yoshuawuyts:
This is very exciting; thanks for putting in the work!

montekki:
@yoshuawuyts can we r+ #263 so this one can be updated and move on from draft?

src/stream/stream/mod.rs
```diff
@@ -484,7 +486,12 @@ extension_trait! {
 
             let s: VecDeque<&str> = vec!["1", "lol", "3", "NaN", "5"].into_iter().collect();
 
-            let mut parsed = s.filter_map(|a| a.parse::<u32>().ok());
+            let mut parsed = s.filter_map(|a|
+                async move {
+                    a.parse::<u32>().ok()
+                });
+
+            let mut parsed = unsafe { std::pin::Pin::new_unchecked(&mut parsed) };
```
yoshuawuyts: Hm is there something we could do about the `unsafe` block here? Even using `pin_utils` might be better than what we currently have, but I'm curious why we need it?

montekki: yep, that's must be some leftover

src/stream/stream/mod.rs
```diff
@@ -797,27 +816,28 @@ extension_trait! {
...
-            let s: VecDeque<usize> = vec![1, 2, 3].into_iter().collect();
-            let sum = s.for_each(move |x| tx.clone().send(x).unwrap()).await;
+            let mut x = 0;
 
-            let v: Vec<_> = rx.iter().collect();
+            let s: VecDeque<_> = vec![1u8, 2, 3].into_iter().collect();
+            s.for_each(|item| {
+                x += item;
+                futures::future::ready(())
+            }).await;
```
yoshuawuyts: Could we perhaps use an async block here?

montekki: capturing in async closures doesn't look that well now, but I will cook an example with Arc and Mutex

montekki:
@yoshuawuyts fixed, but the `for_each` example is now cringe. Idk how to dodge this kind of capturing trouble with the support from the language we have at hand r.n.

src/stream/stream/mod.rs
yoshuawuyts: @montekki oh dang, yeah you're right -- I think what we had before is def better indeed. Perhaps we should revert it?

This is def yet another argument for having async closures!

montekki: done

yoshuawuyts:
Dang, it seems that CI went and broke between the last commit and the most recent one. Currently trying to fix parts of it through #280, but I'm thinking I should probably just manually run the tests on this PR to confirm things are good and then merge (:

yoshuawuyts:
@montekki could you try rebasing on master?

yoshuawuyts:
@montekki could you try rebasing again?

dignifiedquire:
@yoshuawuyts @montekki what is the state of this?

montekki:
@dignifiedquire this still needed? merged with master, going to do required cleanups a bit later

taiki-e:
As far as I know, this was postponed to the next major version: https://github.com/async-rs/async-std/pull/496#issuecomment-552377963 (perhaps because async closure is incomplete)

montekki:
Ok no rush then.

montekki:
Ok, at this point I think that I'd rather re-implement it from scratch when time comes than continue attempts to update this PR so closing.

PR #263, opened 2019-09-30
Adds stream map combinator

montekki:
Ref: #129 
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map


yoshuawuyts:
This looks good! Merging manually because nightly is broken.

PR #264, opened 2019-09-30
Adds for_each stream combinator

montekki:
Ref: #129 
Stdlib: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each

### DIFF
--- a/src/stream/stream/mod.rs
+++ b/src/stream/stream/mod.rs
@@ -30,6 +30,7 @@ mod filter_map;
 mod find;
 mod find_map;
 mod fold;
+mod for_each;
 mod fuse;
 mod inspect;
 mod min_by;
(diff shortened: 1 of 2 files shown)


src/stream/stream/mod.rs
```diff
@@ -750,6 +752,41 @@ extension_trait! {
...
+            # fn main() { async_std::task::block_on(async {
+            #
+            use async_std::prelude::*;
+            use std::collections::VecDeque;
+            use std::sync::mpsc::channel;
+
+            let (tx, rx) = channel();
+
+            let s: VecDeque<usize> = vec![1, 2, 3].into_iter().collect();
+            let sum = s.for_each(move |x| tx.clone().send(x).unwrap()).await;
```
yoshuawuyts: Could this method maybe take an async closure as well, similar to #262? Thanks!

montekki: Yup, like i said i intend to update #262 as soon as this is merged. Or we could do this the other way around: merge #262 and update this.

yoshuawuyts: oh I missed that -- yeah let's go ahead and merge this first!

PR #267, opened 2019-10-01
Add pull_request event to github actions hooks

k-nasa:
https://github.com/async-rs/async-std/pull/253

I forgot to add pull_request event hooks.
Probably necessary, but what about?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-on: [push, pull_request]
+on: pull_request
 
 jobs:
   build_and_test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build_and_test:
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -1,4 +1,5 @@
-on: push
+on: [push, pull_request]
+
 name: Clippy check
 jobs:
   clippy_check:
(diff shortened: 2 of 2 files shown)


k-nasa:
GitHub actions worked. Would have been necessary

スクリーンショット 2019-10-01 10 50 28

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
```
yoshuawuyts: This should probably only be on `pull_request`. Else you get a double event triggering for each commit, which can be really noisy.

.github/workflows/clippy.yml
```diff
@@ -1,4 +1,5 @@
-on: push
+on: [push, pull_request]
```
yoshuawuyts: This should probably only be on `pull_request`. Else you get a double event triggering for each commit, which can be really noisy.

k-nasa:
Ok. I fixed. 
However, is ci not working when pushed to master branch?

yoshuawuyts:
> However, is ci not working when pushed to master branch?

Not sure; in general all of our commits are through PRs anyway, and if Bors works then we don't need to run the *extra* CI run on master merges.

It looks like Clippy is failing though; not sure what's going on as it's a lot of JSON output.

k-nasa:
@yoshuawuyts  My imagination is that this access token is mine (the person who pushed it) and I don't have write access, so it may have failed.
```yml
token: ${{ secrets.GITHUB_TOKEN }}
```

yoshuawuyts:
Alright then; let's merge this and see if it works :crossed_fingers:

PR #268, opened 2019-10-01
Adds partial_cmp.rs file and partial_cmp signature to mod.rs

assemblaj:
Ports partial_cmp functionality from std::iter to async-std.


src/stream/stream/partial_cmp.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl<L: Stream, R: Stream> Future for PartialCmpFuture<L, R> 
+where 
+    L: Stream + Unpin + Sized,
+    R: Stream + Unpin + Sized,
+    L::Item: PartialOrd<R::Item>
+{
+    type Output = Option<Ordering>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
montekki: uhhh, this doesn't look right. Values might arrive from streams at different points at time, but if this happens this doesn't mean that the streams are not equal. The more correct behaviour here would probably be as follows:
1) We cache both elements we are currently comparing in the state of the `PartialCmpFuture`, in `Option`s.
2) As soon as both elements are received (Both options are `.is_some()`) we can compare elements.
3) We have to consider all edge cases: i.e.
 a) we have `l_next`, `l` has closed, but we don't have `r_next` and `r` is not yet closed
 b) the same for `r_next`
 c) both streams are closed but we have only one element left in cache
 d) and so on

assemblaj: I see. I wanted to start by replicating the std::iter [code](https://doc.rust-lang.org/1.3.0/src/core/iter.rs.html#3229-3243) for partial_cmp as much as possible, but you're correct that it doesn't account for values from each stream arriving at different times. Thanks for the feedback.

montekki: I also think that putting `fuse`ing both streams would help.

src/stream/stream/partial_cmp.rs
```diff
@@ -0,0 +1,56 @@
...
+            (None, None) => Poll::Ready(Some(Ordering::Equal)),
+            (None, _   ) => Poll::Ready(Some(Ordering::Less)),
+            (_,    None) => Poll::Ready(Some(Ordering::Greater)),
+            (Some(x), Some(y)) =>  match x.partial_cmp(&y) {
+                    Some(Ordering::Equal) => { 
+                        // wakes task to pull the next item from stream
+                        cx.waker().wake_by_ref();
+                        Poll::Pending
+                    },
+                    non_eq      => Poll::Ready(non_eq),
```
yoshuawuyts: This is probably going to fail `cargo fmt` checks.

src/stream/stream/partial_cmp.rs
```diff
@@ -0,0 +1,93 @@
...
+                self.as_mut().l_cache = None;  
+                self.as_mut().r_cache = None;
+            } else {
+                // Return non equal value 
+                return Poll::Ready(result);
+            }            
+        }
+
+        // wakes task to pull the next item from stream
+        cx.waker().wake_by_ref();
```
montekki: instead of this probably we should do a `loop` like in `fold` or `skip_while`

src/stream/stream/partial_cmp.rs
```diff
@@ -0,0 +1,93 @@
...
+            l_cache: None, 
+            r_cache: None, 
+        }
+    }
+}
+
+impl<L: Stream, R: Stream> Future for PartialCmpFuture<L, R> 
+where 
+    L: Stream + Unpin + Sized,
+    R: Stream + Unpin + Sized,
```
montekki: I don't believe `Unpin` bounds are necessary here, are they?

src/stream/stream/mod.rs
```diff
@@ -1030,6 +1032,50 @@ extension_trait! {
...
+
+            let result_equal = vec![1.].into_iter().collect::<VecDeque<f64>>()
+                .partial_cmp(vec![1.].into_iter().collect::<VecDeque<f64>>()).await;
+            let result_less = vec![1.].into_iter().collect::<VecDeque<f64>>()
+                .partial_cmp(vec![1., 2.].into_iter().collect::<VecDeque<f64>>()).await;
+            let result_greater = vec![1., 2.].into_iter().collect::<VecDeque<f64>>()
+                .partial_cmp(vec![1.].into_iter().collect::<VecDeque<f64>>()).await;
+            let result_none = vec![std::f64::NAN].into_iter().collect::<VecDeque<f64>>()
+                .partial_cmp(vec![1.].into_iter().collect::<VecDeque<f64>>()).await;
+
```
montekki: Probably also tests for `Greater` and `Less` for streams of the same number of elements but with different values?

src/stream/stream/partial_cmp.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use super::fuse::Fuse;
+use crate::prelude::*;
+use crate::future::Future;
+use crate::stream::Stream;
+use crate::task::{Context, Poll};
+
+// Lexicographically compares the elements of this `Stream` with those
+// of another.
+#[derive(Debug)]
```
montekki: I think that we hide docs for all `Future` types

```suggestion
#[doc(hidden)]
#[allow(missing_debug_implementations)]
```

src/stream/stream/partial_cmp.rs
assemblaj: Generally, when should  looping be used instead of `cx.waker().wake_by_ref()` ?  I've seen either approach used throughout the codebase and I wasn't sure which was the most applicable for this situation.

montekki:
@yoshuawuyts Shall we r+?

yoshuawuyts:
@assemblaj it seems like the rebase didn't go as expected; this is changing 1800 lines and 79 files!

assemblaj:
@yoshuawuyts  I honestly have no idea what's happening here.  Currently trying to fix this.

stjepang:
There are still some conflicts that need to be resolved.

assemblaj:
@yoshuawuyts @stjepang  Rebased again just now. Everything should be good.

stjepang:
It looks like `map`, `for_each`, and `try_for_each` are getting removed by this PR. I suppose that was not intentional?

assemblaj:
@yoshuawuyts  @stjepang  The files for `map`, `for_each`,  and `try_for_each` are there, they just keep getting removed from mod.rs for reason reason.  There's been a lot of issues with this branch related to rebasing, so I created a new branch and opened a new PR here: #327

stjepang:
Thanks for fixing this and opening a new PR!

PR #269, opened 2019-10-01
Adds try_for_each combinator

montekki:
This should probably be hidden behind the unstable feature.

---
Ref: #129 
Stdlib: https://doc.rust-lang.org/beta/std/iter/trait.Iterator.html#method.try_for_each

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,7 +47,6 @@
 #![doc(test(attr(allow(unused_extern_crates, unused_variables))))]
 #![doc(html_logo_url = "https://async.rs/images/logo--hero.svg")]
 #![recursion_limit = "1024"]
-#![feature(try_trait)]
 
 use cfg_if::cfg_if;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,6 +47,7 @@
 #![doc(test(attr(allow(unused_extern_crates, unused_variables))))]
 #![doc(html_logo_url = "https://async.rs/images/logo--hero.svg")]
 #![recursion_limit = "1024"]
+#![feature(try_trait)]
 
 use cfg_if::cfg_if;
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -47,6 +47,7 @@
 #![doc(test(attr(allow(unused_extern_crates, unused_variables))))]
 #![doc(html_logo_url = "https://async.rs/images/logo--hero.svg")]
 #![recursion_limit = "1024"]
+#![feature(try_trait)]
```
yoshuawuyts: Depending on a nightly feature may be tricky for us because it means we wont' be able to run on stable for the 1.39 release. I'm thinking it may perhaps be better to scope `try_for_each` to `Result` for now, and mention this as a limitation to the lang team.

src/stream/stream/mod.rs
```diff
@@ -921,6 +924,52 @@ extension_trait! {
...
+            # fn main() { async_std::task::block_on(async {
+            #
+            use std::collections::VecDeque;
+            use std::sync::mpsc::channel;
+            use async_std::prelude::*;
+
+            let (tx, rx) = channel();
+
+            let s: VecDeque<usize> = vec![1, 2, 3].into_iter().collect();
+            let s = s.try_for_each(|v| {
```
yoshuawuyts: Do you think it's possible to make this operate on `|v| async {`?

montekki: I will update #262 as soon as this is merged, yes.

src/lib.rs
montekki: Ok, but how does `std` work then? (i haven't had a moment to look closely atm)

yoshuawuyts:
It looks like nightly is currently broken; https://github.com/async-rs/async-std/pull/269/checks?check_run_id=243557011 passed, so going to go ahead and merge this manually. Thanks heaps!

PR #270, opened 2019-10-01
Fix clippy warning

k-nasa:
Looking at GitHub actions, it seems that clippy warnnings are accumulating.
```
Clippy results: 0 ICE, 0 errors, 31 warnings, 0 notes, 0 help
```

### DIFF
--- a/src/task/block_on.rs
+++ b/src/task/block_on.rs
@@ -69,12 +69,11 @@ where
         let future = task_local::add_finalizer(future);
 
         let future = async move {
-            let res = future.await;
+            future.await;
             trace!("block_on completed", {
                 parent_id: parent_id,
                 child_id: child_id,
             });
-            res
         };
 
         // Pin the future onto the stack.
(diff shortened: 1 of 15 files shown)


skade:
Especially the `Default` implementations are not implemented in stdlib and we should think about following that lint or turning it off instead.

src/task/block_on.rs
```diff
@@ -69,12 +69,11 @@ where
         let future = task_local::add_finalizer(future);
 
         let future = async move {
```
skade: For @stjepang: This is not an equivalent transform. Reading the surrounding code, I also don't think that matters, but can you please revalidate?

k-nasa:
@skade I think the default trait should be as much as possible.
The library user should not be saddened by the lack of Default trati.
So I think lint should not be turned off.
How about ❓

src/task/block_on.rs
stjepang: This change looks fine by me. It doesn't really matter because `res` is `()`.

k-nasa:
I don't have the right to see it, but it looks like bors is waiting

bors ping

yoshuawuyts:
Yep, merging manually -- this has been an issue for all PRs for the last day or so.

PR #272, opened 2019-10-02
Fix uds listener hanging on accept

### LINKED ISSUE #248: unix sockets don't work since the 0.99.6 release
Expected behaviour:
```sh
loop
connected
loop
```
Actual behaviour:
```sh
loop
```
`server.rs`
```rust
use async_std::prelude::*;
use async_std::os::unix::net::UnixListener;
use async_std::task;

async fn run() {
    let socket = UnixListener::bind("/tmp/test.sock").await.expect("bind");
    let mut incoming = socket.incoming();
    loop {
        println!("loop");
        if let Some(stream) = incoming.next().await {
            println!("connected");
        }
    }
}

fn main() {
    task::block_on(run());
}
```
`client.rs`
```rust
use async_std::task;
use async_std::prelude::*;
use async_std::os::unix::net::UnixStream;

async fn run() {
    let mut socket = UnixStream::connect("/tmp/test.sock").await.expect("connect");
    socket.write_all(b"hello world").await.expect("write");
    socket.flush().await.expect("flush");
}

fn main() {
    task::block_on(run());
}
```

mvucenovic:
UDS listener was hanging because the accept method would return `Poll::Pending` without registering the task to be awoken in the case when underlying unix listener returns a `WouldBlock` that gets converted to `None`. This is a hacky fix for this case.

Should fix #248

### DIFF
--- a/src/os/unix/net/listener.rs
+++ b/src/os/unix/net/listener.rs
@@ -95,9 +95,9 @@ impl UnixListener {
         future::poll_fn(|cx| {
             let res = futures_core::ready!(self.watcher.poll_read_with(cx, |inner| {
                 match inner.accept_std() {
-                    // Converting to `would block` so that the watcher will
+                    // Converting to `WouldBlock` so that the watcher will
                     // add the waker of this task to a list of readers.
-                    Ok(None) => Err(std::io::ErrorKind::WouldBlock.into()),
+                    Ok(None) => Err(io::ErrorKind::WouldBlock.into()),
                     res => res,
                 }
             }));
@@ -110,7 +110,7 @@ impl UnixListener {
                     };
                     Poll::Ready(Ok((stream, addr)))
                 }
-                // This should never happen since None is converted to WouldBlock
+                // This should never happen since `None` is converted to `WouldBlock`
                 None => unreachable!(),
             }
         })
--- a/src/os/unix/net/listener.rs
+++ b/src/os/unix/net/listener.rs
@@ -93,11 +93,16 @@ impl UnixListener {
     /// ```
     pub async fn accept(&self) -> io::Result<(UnixStream, SocketAddr)> {
         future::poll_fn(|cx| {
-            let res =
-                futures_core::ready!(self.watcher.poll_read_with(cx, |inner| inner.accept_std()));
+            let res = futures_core::ready!(self.watcher.poll_read_with(cx, |inner| {
+                match inner.accept_std() {
+                    // Converting to `would block` so that the watcher will
+                    // add the waker of this task to a list of readers.
+                    Ok(None) => Err(std::io::ErrorKind::WouldBlock.into()),
+                    res => res,
+                }
+            }));
 
             match res? {
-                None => Poll::Pending,
                 Some((io, addr)) => {
                     let mio_stream = mio_uds::UnixStream::from_stream(io)?;
                     let stream = UnixStream {
@@ -105,6 +110,8 @@ impl UnixListener {
                     };
                     Poll::Ready(Ok((stream, addr)))
                 }
+                // This should never happen since None is converted to WouldBlock
+                None => unreachable!(),
             }
         })
         .await
--- a/tests/uds.rs
+++ b/tests/uds.rs
@@ -1,8 +1,13 @@
 #![cfg(unix)]
 
 use async_std::io;
-use async_std::os::unix::net::UnixDatagram;
+use async_std::os::unix::net::{UnixDatagram, UnixListener, UnixStream};
+use async_std::prelude::*;
 use async_std::task;
+
+use tempdir::TempDir;
+
+use std::time::Duration;
 
 const JULIUS_CAESAR: &[u8] = b"
     Friends, Romans, countrymen - lend me your ears!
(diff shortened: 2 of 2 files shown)


skade:
This would be great chance to cover this with a test case to avoid regressions. @mvucenovic, would you be able to find a case that exhibits the (previously broken) behaviour?

mvucenovic:
@skade Sure thing, I would like to write a bigger integration test that reproduces this issue, as I've mentioned in the #248. I won't be able to do this today, but will try to finish it before the weekend. I've submitted this PR early, because I consider this change a hacky one, so any suggestions how this should be done better are welcome.

skade:
@mvucenovic Perfect :+1: thanks for letting me know!

mvucenovic:
Hey @skade, I've added a test that reproduces the listener hanging forever. For me it fails on the master, but passes on this branch.

Interesting thing, my first version of test was not running in two different threads, but used join macro, to run server and client concurrently, and that version was green on the current master. It looked something like this:
```
     // rest of the code is mostly the same
     task::block_on(async {
         let listener = UnixListener::bind(&socket).await.expect("Socket bind");

         let server = ping_pong_server(listener, iter_cnt);
         let client = ping_pong_client(&socket, iter_cnt);

         async_std::io::timeout(
             Duration::from_secs(5),
             async_std::future::try_join!(server, client),
         ).await
     }).expect("not finished successfully");
```
I'll try to investigate why was that the case, just wanted to mention if somebody has any idea.

jD91mZM2:
Hi, sorry for the notification, I just wanted to say that I needed to use unix sockets and switching to your fork solved it for me, just like that. You're awesome! :)

stjepang:
@mvucenovic If you want to force tasks to be executed on separate threads, you can spawn threads manually with `std::thread::spawn` and then use `async_std::task::block_on` to run futures on those threads.

src/os/unix/net/listener.rs
```diff
@@ -93,18 +93,25 @@ impl UnixListener {
     /// ```
     pub async fn accept(&self) -> io::Result<(UnixStream, SocketAddr)> {
         future::poll_fn(|cx| {
-            let res =
-                futures_core::ready!(self.watcher.poll_read_with(cx, |inner| inner.accept_std()));
+            let res = futures_core::ready!(self.watcher.poll_read_with(cx, |inner| {
+                match inner.accept_std() {
+                    // Converting to `would block` so that the watcher will
+                    // add the waker of this task to a list of readers.
+                    Ok(None) => Err(std::io::ErrorKind::WouldBlock.into()),
```
stjepang:
```suggestion
                    Ok(None) => Err(io::ErrorKind::WouldBlock.into()),
```

src/os/unix/net/listener.rs
```diff
@@ -93,18 +93,25 @@ impl UnixListener {
     /// ```
     pub async fn accept(&self) -> io::Result<(UnixStream, SocketAddr)> {
         future::poll_fn(|cx| {
-            let res =
-                futures_core::ready!(self.watcher.poll_read_with(cx, |inner| inner.accept_std()));
+            let res = futures_core::ready!(self.watcher.poll_read_with(cx, |inner| {
+                match inner.accept_std() {
+                    // Converting to `would block` so that the watcher will
```
stjepang:
```suggestion
                    // Converting to `WouldBlock` so that the watcher will
```

src/os/unix/net/listener.rs
```diff
@@ -93,18 +93,25 @@ impl UnixListener {
...
             match res? {
-                None => Poll::Pending,
                 Some((io, addr)) => {
                     let mio_stream = mio_uds::UnixStream::from_stream(io)?;
                     let stream = UnixStream {
                         watcher: Watcher::new(mio_stream),
                     };
                     Poll::Ready(Ok((stream, addr)))
                 }
+                // This should never happen since None is converted to WouldBlock
```
stjepang:
```suggestion
                // This should never happen since `None` is converted to `WouldBlock`.
```

tests/uds.rs
```diff
@@ -39,3 +44,53 @@ fn into_raw_fd() -> io::Result<()> {
         Ok(())
     })
 }
+
+const PING: &[u8] = b"ping";
+const PONG: &[u8] = b"pong";
+const TEST_TIMEOUT: std::time::Duration = Duration::from_secs(3);
```
stjepang:
```suggestion
const TEST_TIMEOUT: Duration = Duration::from_secs(3);
```

stjepang:
@skade @yoshuawuyts It seems Clippy checks are failing in this pull request. Should we make Clippy tests not required to pass in CI?

yoshuawuyts:
@stjepang the failure here was an upstream failure iirc that's now been fixed in actions-rs. Re-running the check should solve it. Clippy currently doesn't fail CI.

Additionally #270 should be landing soon now which means clippy errors should be resolved. This would allow us to then enable https://github.com/actions-rs/clippy-check/issues/2#issuecomment-538670997 once it's implemented (if we so desire).

Anyway, going to restart the CI here which *should* fix the error!

mvucenovic:
Thanks for the review @stjepang, I've updated the PR with your suggestions.

PR #277, opened 2019-10-05
Adds a from_fn stream implementation

montekki:
Hope the example is not too ugly.

--- 
Ref: #129 
Stdlib: https://doc.rust-lang.org/std/iter/fn.from_fn.html

### DIFF
--- a/src/stream/from_fn.rs
+++ b/src/stream/from_fn.rs
@@ -5,7 +5,7 @@ use crate::future::Future;
 use crate::stream::Stream;
 use crate::task::{Context, Poll};
 
-/// A stream that yeilds elements by calling a closure.
+/// A stream that yields elements by calling a closure.
 ///
 /// This stream is constructed by [`from_fn`] function.
 ///
--- a/src/stream/from_fn.rs
+++ b/src/stream/from_fn.rs
@@ -1,0 +1,99 @@
+use std::marker::PhantomData;
+use std::pin::Pin;
+
+use crate::future::Future;
+use crate::stream::Stream;
+use crate::task::{Context, Poll};
+
+/// A stream that yeilds elements by calling a closure.
+///
+/// This stream is constructed by [`from_fn`] function.
+///
+/// [`from_fn`]: fn.from_fn.html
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
Related #130 -- we closed that PR because async closures aren't really useful right now. But I guess it's "later" now, and with #262 and #279 we're already introducing several functions relying on async closures, so this might feel less out of place.

src/stream/from_fn.rs
```diff
@@ -0,0 +1,99 @@
...
+///     async move {
+///         *count.lock().unwrap() += 1;
+///
+///         if *count.lock().unwrap() > 3 {
+///             None
+///         } else {
+///             Some(*count.lock().unwrap())
+///         }
+///     }
+/// });
```
yoshuawuyts: I like how this example somewhat creatively works around the limitations with async closures right now. This is the kind of thing people may want to do for sure, and I think there's value in showing how to do it _now_ also.

src/stream/from_fn.rs
```diff
@@ -0,0 +1,99 @@
...
+impl<F, Fut, T> Stream for FromFn<F, Fut, T>
+where
+    F: FnMut() -> Fut,
+    Fut: Future<Output = Option<T>>,
+{
+    type Item = T;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        loop {
+            match self.future.is_some() {
```
yoshuawuyts: I think matching `self.future` with `Some,None` would be better here.

src/stream/from_fn.rs
```diff
@@ -0,0 +1,99 @@
+use std::marker::PhantomData;
+use std::pin::Pin;
+
+use crate::future::Future;
+use crate::stream::Stream;
+use crate::task::{Context, Poll};
+
+/// A stream that yeilds elements by calling a closure.
```
yoshuawuyts:
```suggestion
/// A stream that yields elements by calling a closure.
```

PR #278, opened 2019-10-05
Fixes feature flag used in generating the docs in the README

Michael-J-Ward:
I thought I'd take a stab at fixing at this https://github.com/async-rs/async-std/issues/82, but I ran into a few errors following the instruction in the README.md

The first error and one that is addressed in this pull request is that the feature flag used to generate the docs is incorrect- it should be `docs` and not `docs.rs`
```
$ cargo doc --features docs.rs --open

error: Package `async-std v0.99.8 (/home/mike/workspace/async-std)` does not have these features: `docs.rs`
```

Perhaps additionally, it should be `--features docs,unstable` as it is in `.travis.yml` . Should I add that to the README? https://github.com/async-rs/async-std/blob/master/.travis.yml#L53

Lastly, I had to use the `+nightly` flag to get it to the docs and example to run. Should I add this or is it assumed that the user would have this configured somewhere?

Thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,4 +54,4 @@ jobs:
       run: cargo fmt --all -- --check
 
     - name: Docs
-      run: cargo doc --features docs,unstable
+      run: cargo doc --features docs
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ matrix:
       rust: nightly
       os: linux
       script:
-        - cargo doc --features docs,unstable
+        - cargo doc --features docs
 
     - name: book
       rust: nightly
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ git clone git@github.com:async-rs/async-std.git && cd async-std
 Generate docs:
 
 ```
-cargo doc --features docs.rs --open
+cargo doc --features docs --open
 ```
 
 Check out the [examples](examples). To run an example:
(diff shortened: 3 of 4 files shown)


stjepang:
> Lastly, I had to use the `+nightly` flag to get it to the docs and example to run. Should I add this or is it assumed that the user would have this configured somewhere?

I think it would be good to use `+nightly` in `README.md`.

> Perhaps additionally, it should be `--features docs,unstable` as it is in `.travis.yml` . Should I add that to the README? https://github.com/async-rs/async-std/blob/master/.travis.yml#L53

Maybe the other way around instead. :)

We use `#[cfg(any(feature = "unstable", feature = "docs"))]` in our code, which means `--features docs` is enough to render all APIs, even unstable ones. So we can safely remove `,unstable` from `.travis.yml`.

Michael-J-Ward:
I've added `+nightly` to the `cargo` commands in the README and removed the `unstable` feature flag from `travis.yml` and `github/workflows/ci.yml`

It looks like `#[cfg(any(feature = "unstable", feature = "docs"))]` is used everywhere except for maybe here: https://github.com/async-rs/async-std/blob/ddee47231ae32c2dbaf7a7a30190d29771b19179/src/task/mod.rs#L28-L31

I'm not yet familiar enough to know if that should be changed because it looks it might be handled in another way there.

stjepang:
@Michael-J-Ward That's a good catch! This line:

```rust
#[cfg(feature = "unstable")] 
```

should be replaced with:

```rust
#[cfg(any(feature = "unstable", feature = "docs"))]
```

yoshuawuyts:
bors is hanging, merging this manually!

PR #280, opened 2019-10-05
install rustfmt with fallback for gh-actions

yoshuawuyts:
This uses https://github.com/actions-rs/components-nightly to fall back to an older version of the rust toolchain if installation of a component fails. This fixes a bug that's currently happening on nightly where `rustfmt` isn't available. Thanks!


yoshuawuyts:
Okay, all tests passing means this should fix our builds! -- going to go ahead and merge!

PR #283, opened 2019-10-06
Add GitHub Actions to the merge requirement of bors

taiki-e:
cc @yoshuawuyts @stjepang


.github/workflows/ci.yml
```diff
@@ -1,4 +1,11 @@
-on: pull_request
+name: CI
```
taiki-e: If you do not set the name, the file path will be displayed and it will be difficult to read.

.github/workflows/ci.yml
```diff
@@ -65,3 +72,21 @@ jobs:
 
     - name: Docs
       run: cargo doc --features docs,unstable
+
+  clippy_check:
```
taiki-e: It is cumbersome to set triggers in multiple places, so merge clippy.yml into ci.yml.

bors.toml
```diff
@@ -1 +1,10 @@
-status = ["continuous-integration/travis-ci/push"]
+status = [
+    # Travis CI
+    "continuous-integration/travis-ci/push",
+    # GitHub Actions
+    "Build and test on ubuntu-latest",
+    "Build and test on windows-latest",
+    "Build and test on macOS-latest",
+    "Checking fmt and docs",
+    "Clippy check",
```
taiki-e: Unfortunately, as far as I know, all job names need to be written. I have checked that the pattern mentioned in [documentation](https://bors.tech/documentation/) cannot handle this properly (https://github.com/taiki-e/derive_utils/pull/22).
> Patterns are written using SQL LIKE syntax. For example, this status directive will match any CI job that starts with ci/gitlab/:

.github/workflows/ci.yml
yoshuawuyts: Agree. I like the name of "test" you used in your repo slightly better here though!

bors.toml
yoshuawuyts: ah dang yeah that's unfortunate but okay I think. Thanks for checking this!

yoshuawuyts: Hm actually we may want to fill in the full matrix here. I think keeping the version as part of the name may be better and set us up for beta testing also (which we'll probably want to start doing soon, in addition to still testing nightly!)

yoshuawuyts: Ah nevermind, we can cross that bridge once we get to it. Thanks!

.github/workflows/ci.yml
taiki-e: I don't have a strong opinion, but as I mentioned in https://github.com/async-rs/async-std/pull/283#discussion_r331788937, I added all the checks to this workflow, so I think "CI" is fine.

taiki-e:
Hmm..., bors does not seem to be able to get past the “waiting in the queue” stage...

bors ping

taiki-e:
bors r-

In any case, I think bors needs to be re-executed.

yoshuawuyts:
Okay, let's try again!

bors r+

yoshuawuyts:
@taiki-e could you perhaps try removing `Travis` from the bors.toml requirements? I'm wondering what's making bors hang right now..

yoshuawuyts:
hmm, seems like bors is failing with "resource not accessible by integration". Maybe we need to re-add bors?

taiki-e:
> Maybe we need to re-authenticate bors?

Probably, yes. I have confirmed changes included in this PR worked in the fork (https://github.com/taiki-e/async-std/pull/1), so I think the changes themselves are not wrong.

PR #285, opened 2019-10-07
Adds Stream:ge   

assemblaj:
Draft PR: Blocked by #268 
For issue #129
Based on : https://github.com/rust-lang/rust/blob/master/src/libcore/iter/traits/iterator.rs#L2847


src/stream/stream/mod.rs
```diff
@@ -1030,6 +1034,94 @@ extension_trait! {
...
+            let result_less_count = vec![1.].into_iter().collect::<VecDeque<f64>>()
+                .partial_cmp(vec![1., 2.].into_iter().collect::<VecDeque<f64>>()).await;
+            let result_greater_count = vec![1., 2.].into_iter().collect::<VecDeque<f64>>()
+                .partial_cmp(vec![1.].into_iter().collect::<VecDeque<f64>>()).await;           
+            let result_less_vals = vec![1., 2., 3.].into_iter().collect::<VecDeque<f64>>()
+                .partial_cmp(vec![1., 2., 4.].into_iter().collect::<VecDeque<f64>>()).await;
+            let result_greater_vals = vec![1., 2., 4.].into_iter().collect::<VecDeque<f64>>()
+                .partial_cmp(vec![1., 2., 3.].into_iter().collect::<VecDeque<f64>>()).await;
+            let result_none = vec![std::f64::NAN].into_iter().collect::<VecDeque<f64>>()
+                .partial_cmp(vec![1.].into_iter().collect::<VecDeque<f64>>()).await;
```
yoshuawuyts: Wow, this is quite the comparison haha. I'm wondering if we could perhaps move most of this to a dedicated test under `src/stream/stream/partial_cmp.rs`, and then add a more reduced example as part of the doc comment. That way we keep the thoroughness of the tests to catch regressions while gaining legibility.

I'm afraid that in its current form folks might be somewhat overwhelmed by the documentation, which is something we generally want to avoid.

assemblaj: This PR was branched from assemblaj-partial_cmp, that's the example code for `partial_cmp`. I can go back and update the `partial_cmp` example it to be cleaner and then update this branch. 

The issue I was having is that the std::iter [example code](https://github.com/rust-lang/rust/blob/master/src/libcore/iter/traits/iterator.rs#L2616) for `partial_cmp` just creates and compares the iterators inline; doing this with VecDeque's doesn't look as clean but I'm not sure about making a few at the top of the function and copying them everywhere either (this is what I did for the Stream::ge examples, see below).  How do you think I should approach this? 

```rust 
let single:     VecDeque<isize> = vec![1].into_iter().collect();
let single_gt:  VecDeque<isize> = vec![10].into_iter().collect();
let multi:      VecDeque<isize> = vec![1,2].into_iter().collect();
let multi_gt:   VecDeque<isize> = vec![1,5].into_iter().collect();

assert_eq!(single.clone().ge(single.clone()).await, true);
assert_eq!(single_gt.clone().ge(single.clone()).await, true);

assert_eq!(multi.clone().ge(single_gt.clone()).await, false);
assert_eq!(multi_gt.clone().ge(multi.clone()).await, true);
```

Issue #48, opened 2019-08-16
Test non x86-64 targets before 1.0.0 release


jamesmunns:
Currently, all CI tests is x86_64. We should have a policy and testing for other platforms, such as ARM, or non-64 bit x86 targets.

yoshuawuyts:
It's been a while since this was opened; @jamesmunns is this something you can / would like to pick up?

stjepang:
I believe github actions don't support non-x86-64 targets. What do we do here? Can we close the issue and reconsider later?

taiki-e:
I'd like to use [cross](https://github.com/rust-embedded/cross); see [this change](https://github.com/async-rs/async-std/pull/286/files#diff-e9f950f17198d3d5e3122a44230a09b9R49-R81) in #286

PR #286, opened 2019-10-07
Use AtomicCell<u64> on targets with target_has_atomic less than 64

taiki-e:
Closes #141

cc #48

Related: https://github.com/tokio-rs/tokio/pull/1538

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,39 @@ jobs:
         command: test
         args: --doc --features "unstable attributes"
 
+  cross:
+    name: Cross compile
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target:
+          - i686-unknown-linux-gnu
+          - powerpc-unknown-linux-gnu
+          - powerpc64-unknown-linux-gnu
+          - mips-unknown-linux-gnu
+          - arm-linux-androideabi
+
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Install nightly
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: nightly
+        override: true
+
+    - name: Install cross
+      run: cargo install cross
+
+    - name: check
+      run: cross check --all --target ${{ matrix.target }}
+
+    - name: check unstable
+      run: cross check --all --features unstable --target ${{ matrix.target }}
+
+    - name: test
+      run: cross test --all --features unstable --target ${{ matrix.target }}
+
   check_fmt_and_docs:
     name: Checking fmt and docs
     runs-on: ubuntu-latest
--- a/src/sync/atomic.rs
+++ b/src/sync/atomic.rs
@@ -1,0 +1,37 @@
+pub(crate) use self::imp::AtomicU64;
+
+// `AtomicU64` can only be used on targets with `target_has_atomic` is 64 or greater.
+// Once `cfg_target_has_atomic` feature is stable, we can replace it with
+// `#[cfg(target_has_atomic = "64")]`.
+// Refs: https://github.com/rust-lang/rust/tree/master/src/librustc_target
+#[cfg(not(any(target_arch = "arm", target_arch = "mips", target_arch = "powerpc")))]
+mod imp {
+    pub(crate) use std::sync::atomic::AtomicU64;
+}
+
+#[cfg(any(target_arch = "arm", target_arch = "mips", target_arch = "powerpc"))]
+mod imp {
+    use std::sync::atomic::Ordering;
+
+    use crossbeam_utils::atomic::AtomicCell;
+
+    pub(crate) struct AtomicU64(AtomicCell<u64>);
+
+    impl AtomicU64 {
+        pub(crate) const fn new(val: u64) -> Self {
+            Self(AtomicCell::new(val))
+        }
+
+        pub(crate) fn load(&self, _: Ordering) -> u64 {
+            self.0.load()
+        }
+
+        pub(crate) fn fetch_add(&self, val: u64, _: Ordering) -> u64 {
+            self.0.fetch_add(val)
+        }
+
+        pub(crate) fn fetch_sub(&self, val: u64, _: Ordering) -> u64 {
+            self.0.fetch_sub(val)
+        }
+    }
+}
--- a/tests/io_timeout.rs
+++ b/tests/io_timeout.rs
@@ -3,7 +3,6 @@ use std::time::Duration;
 use async_std::io;
 use async_std::task;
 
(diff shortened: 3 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -18,22 +18,44 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@master
 
-    - name: tests
-      uses: actions-rs/cargo@v1
+    - name: Install nightly
+      uses: actions-rs/toolchain@v1
       with:
-        command: test
-        args: --all --doc --features unstable
```
taiki-e: Note that if `cargo test` contains `--doc` arguments, only doc test is executed (fixed in this pr).

src/sync/atomic.rs
```diff
@@ -0,0 +1,39 @@
+// `AtomicU64` can only be used on targets with `target_has_atomic` is 64 or greater.
+// Once `cfg_target_has_atomic` feature is stable, we can replace it with
+// `#[cfg(target_has_atomic = "64")]`.
+// Refs: https://github.com/rust-lang/rust/tree/master/src/librustc_target
+cfg_if::cfg_if! {
+    if #[cfg(not(any(target_arch = "arm", target_arch = "mips", target_arch = "powerpc")))] {
```
taiki-e: Many ` target_arch = "arm"` targets are `target_has_atomic = "64"`, but only some targets seems different. 
* [arm-linux-androideabi](https://github.com/rust-lang/rust/blob/master/src/librustc_target/spec/arm_linux_androideabi.rs#L7)
* [armv5te-unknown-linux-musleabi](https://github.com/rust-lang/rust/blob/master/src/librustc_target/spec/armv5te_unknown_linux_musleabi.rs#L23)

taiki-e:
CI failure is due to the fact that [I enabled the test.](https://github.com/async-rs/async-std/pull/286#discussion_r331923531)

src/sync/atomic.rs
taiki-e: cc #48 Should we support these targets?

skade: I would definitely support `arm-linux-androideabi`, I'm hard pressed to find a deployment target with `armv5` that is relevant for async-std.

stjepang:
This PR needs another rebase.

taiki-e:
This PR should not have changed the behavior on macOS, but [this test](https://github.com/async-rs/async-std/blob/a14b11c0325cd362295d9558e2028f51ca3ff5f2/tests/io_timeout.rs#L6-L19) failed on macOS on CI... But I couldn't reproduce this locally.

nbdd0121:
Would it be better to name it as something else rather than AtomicU64? Rust atomics are guaranteed to be layout-wise identical to their corresponding integer type, so naming it as AtomicU64 may incorrectly represent that.

Also: std::sync::Mutex is too heavy to protect just an integer. It'd be better to have a spinlock instead.

taiki-e:
> Would it be better to name it as something else rather than AtomicU64? Rust atomics are guaranteed to be layout-wise identical to their corresponding integer type, so naming it as AtomicU64 may incorrectly represent that.

I don't intend to make these implementations public, so I don't think it's a problem. However, if #355 is merged, we may need to consider this.

> Also: std::sync::Mutex is too heavy to protect just an integer. It'd be better to have a spinlock instead.

I agree. And another advantage of this is `AtomicU64::new` can be const function.

nbdd0121:
Another thought, would it make sense to do similar thing as crossbeam's [AtomicCell](https://docs.rs/crossbeam/0.7.2/crossbeam/atomic/struct.AtomicCell.html) which uses a global Mutex instead? This can guarantee layout compatibility with u64, as well as a const new fn.

tests/io_timeout.rs
```diff
@@ -3,6 +3,7 @@ use std::time::Duration;
 use async_std::io;
 use async_std::task;
 
+#[ignore]
```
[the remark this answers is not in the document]
taiki-e: https://github.com/async-rs/async-std/pull/286#issuecomment-543616225
> This PR should not have changed the behavior on macOS, but [this test](https://github.com/async-rs/async-std/blob/a14b11c0325cd362295d9558e2028f51ca3ff5f2/tests/io_timeout.rs#L6-L19) failed on macOS on CI... But I couldn't reproduce this locally.

...but, since it is a comment more than 2 weeks ago, I will delete that `#[ignore]` and see if CI succeeds.

taiki-e:
> I like the idea of using `AtomicCell<u64>` from `crossbeam-channel` as a cross-platform version of `AtomicU64` :)
> 
> @taiki-e let's do that?

I think it needs a new release of crossbeam-utils including https://github.com/crossbeam-rs/crossbeam/pull/437. Filed https://github.com/crossbeam-rs/crossbeam/pull/444.

stjepang:
@taiki-e New versions of crossbeam have been published now.

taiki-e:
looks like ci failure is https://github.com/async-rs/async-std/pull/286#discussion_r342188128.

stjepang:
@taiki-e Can you try rebasing/merging with the master branch to see if that helps?

taiki-e:
Updated. It seems to succeed on the existing targets (linux,macos,windows), but fails on the newly added targets.

faulesocke:
Rebasing this is almost trivial (just remove the Cargo.toml changes). It then applies cleanly and makes async-std work on MIPS and other platforms.

taiki-e:
Hmm, still a test mentioned in https://github.com/async-rs/async-std/pull/286#discussion_r342188128 is broken on the newly added targets.

faulesocke:
Why was this closed? It still does not compile on MIPS.

flxo:
I found this PR today while trying to cross compile `async-std` for `arm-linux-androideabi` (32bit). There's no u64 atomic support on 32bit arm v7 and thus also the here proposed usage of `crossbeam::atomic` doesn't solve the problem. The proposal to use `AtomicUsize` from #451 sounds reasonable to me but it got closed with a reference to here. Maybe someone can elaborate on the state? I'm happy to support on this issue (I would have to brew my own solution/hack anyway...)

kolapapa:
> I found this PR today while trying to cross compile `async-std` for `arm-linux-androideabi` (32bit). There's no u64 atomic support on 32bit arm v7 and thus also the here proposed usage of `crossbeam::atomic` doesn't solve the problem. The proposal to use `AtomicUsize` from #451 sounds reasonable to me but it got closed with a reference to here. Maybe someone can elaborate on the state? I'm happy to support on this issue (I would have to brew my own solution/hack anyway...)

I thought this PR was going to pass, so I closed it first……

taiki-e:
> Why was this closed?

Closed because inactive

Issue #128, opened 2019-08-29
Implement Stream::size_hint


yoshuawuyts:
Equal to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint. Don't think this needs to be marked `async`. This would've been useful to calculate the right pre-alloc size for calling `collect` into a vec in #125. Thanks!

taiki-e:
Could you explain the value of `size_hint` that cannot be implemented manually?

yoshuawuyts:
@taiki-e I feel its mostly API consistency? With iter size hint it's useful when pre-allocating structures. For example when using `collect` for vec it would've been useful to get an estimate of how much to allocate. Right now we start with an empty vector and allocate as we go, which is not nearly as efficient as it could be.

taiki-e:
The problem is that `size_hint` always returns the default value because `async_std::stream::Stream` cannot be implemented manually. So, currently, calling `size_hint` for optimization is a needless operation.

taiki-e:
IMO, we should wait for [the upstream to add size_hint](https://github.com/rust-lang-nursery/futures-rs/issues/170), instead of adding such a “useless API”.

yoshuawuyts:
@taiki-e ah yeah, fair point.

Though I don't disagree, I do see value in implementing all APIs sooner rather than later though. This way we can have the external API that folks can call, which can be incorporated in app logic — and later when we fix the impl nobody needs to rewrite their code.

Or well, that's the theory.

yoshuawuyts:
Yay, https://github.com/rust-lang-nursery/futures-rs/pull/1853 landed!

yoshuawuyts:
We're now blocked on `futures-preview@0.3.0-alpha.19` landing to implement this.

yoshuawuyts:
Sweet, this has landed! This means we can now rely on `size_hint` for optimizing allocations in `FromStream` impls!

cc/ @sunjay

PR #287, opened 2019-10-08
Optimizing allocations using Stream::size_hint

k-nasa:
close: https://github.com/async-rs/async-std/issues/128
Is this the intention of this issue?

### DIFF
--- a/src/collections/linked_list/extend.rs
+++ b/src/collections/linked_list/extend.rs
@@ -10,9 +10,6 @@ impl<T> Extend<T> for LinkedList<T> {
         stream: S,
     ) -> Pin<Box<dyn Future<Output = ()> + 'a>> {
         let stream = stream.into_stream();
-        //TODO: Add this back in when size_hint is added to Stream/StreamExt
-        //let (lower_bound, _) = stream.size_hint();
-        //self.reserve(lower_bound);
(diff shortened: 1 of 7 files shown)


src/collections/linked_list/extend.rs
```diff
@@ -10,9 +10,6 @@ impl<T> Extend<T> for LinkedList<T> {
         stream: S,
     ) -> Pin<Box<dyn Future<Output = ()> + 'a>> {
         let stream = stream.into_stream();
-        //TODO: Add this back in when size_hint is added to Stream/StreamExt
-        //let (lower_bound, _) = stream.size_hint();
-        //self.reserve(lower_bound);
```
yoshuawuyts: Did you intend to add the size hint?

yoshuawuyts: Did you intend to add the size hint?

k-nasa: sorry! I forgot to ask a question.
A LinkedList should consist of a chain of pointers. So, isn't it difficult to make a size or change the size?

yoshuawuyts: @k-nasa oh yeah, that's accurate! I see what you mean; I agree this doesn't make sense.

PR #296, opened 2019-10-09
Cleanup BufWriter

stjepang:
Just some refactoring, no real changes.

### DIFF
--- a/src/io/buf_writer.rs
+++ b/src/io/buf_writer.rs
@@ -1,9 +1,10 @@
-use crate::task::{Context, Poll};
-use futures_core::ready;
-use futures_io::{AsyncSeek, AsyncWrite, SeekFrom};
 use std::fmt;
-use std::io;
 use std::pin::Pin;
+
+use futures_core::ready;
+
+use crate::io::{self, Seek, SeekFrom, Write};
+use crate::task::{Context, Poll};
 
 const DEFAULT_CAPACITY: usize = 8 * 1024;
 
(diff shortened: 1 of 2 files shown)


src/io/buf_writer.rs
```diff
@@ -301,80 +293,3 @@ impl<W: AsyncWrite + AsyncSeek> AsyncSeek for BufWriter<W> {
...
-        task::block_on(async {
-            let mut w = BufWriter::with_capacity(3, io::Cursor::new(Vec::new()));
-            w.write_all(&[0, 1, 2, 3, 4, 5]).await.unwrap();
-            w.write_all(&[6, 7]).await.unwrap();
-            assert_eq!(w.seek(SeekFrom::Current(0)).await.ok(), Some(8));
-            assert_eq!(&w.get_ref().get_ref()[..], &[0, 1, 2, 3, 4, 5, 6, 7][..]);
-            assert_eq!(w.seek(SeekFrom::Start(2)).await.ok(), Some(2));
-        })
-    }
-}
```
yoshuawuyts: I kind of like these inline tests though; we have them for some of the `sync` modules as well.

yoshuawuyts: (don't feel strongly about this at all; but figured I'd still mention it :blush:)

taiki-e: IIRC, this removes the need to recompile crate every time you change unit tests.

stjepang: Only for `Barrier`, right? Tests for `Mutex` and `RwLock` are in the `tests` directory. I don't have a strong preference either, but would like to be consistent about it :)

My general rule of thumb is that if tests are only poking the public API, they should go into the `tests` directory. If they are checking internal (private) state of the thing that is being tested, it should go inside the module where it's implemented.

yoshuawuyts: Seems fair enough!

Issue #290, opened 2019-10-08
[discussion] Add task::yield_now()?


yoshuawuyts:
## Motivation

Sometimes it can be useful to break up longer execution into smaller parts. As a general rule `task::blocking` fills that role for us (#251). But std has another method for smoothing out load within threads: [`thread::yield_now`](https://doc.rust-lang.org/std/thread/fn.yield_now.html). This function hints to the OS-scheduler that it's ready to yield for a while so other work can resume, improving tail-latency.

When dealing with longer-running tasks, especially with `task::blocking`, it can make sense to occasionally pro-actively yield control back to schedulers. This can help with throughput, especially if there we're hitting upper limits on system resources (e.g. CPU, threads).

## Implementation

I'm imagining inside `task::blocking` we'd wrap [`thread::yield_now`](https://doc.rust-lang.org/std/thread/fn.yield_now.html) inside a future to let other threads run on-cpu instead.

Inside `task::spawn` / `task::block_on` this would re-schedule the current task on the futures executor. In general this seems less important than the `task::blocking` use, but it might still come in useful, and overall it seems good to still let it do more or less what you'd expect it to.

The signature would be:

```rust
async fn yield_now();
```

## Example
```rust
async fn main() {
    let a = task::spawn(async { println!("hello world"); });
    task::yield_now();
    a.await;
}
```

Issue #289, opened 2019-10-08
[discussion] task::block_on -> thread::block_on?


yoshuawuyts:
Okay so this might be a terrible idea, and I'm not convinced myself -- but I think it's still worth bringing up because there might be value to it.

In https://github.com/async-rs/async-std/pull/251 we're going to add `task::blocking`, but the name is kind of similar to `task::block_on`. The difference between the two is that one takes a closure that will block a thread, while the other one blocks the current thread:

```rust
fn block_on(fut: impl Future<Output = T>) -> T;
fn blocking(fut: impl Future<Output = T>) -> JoinHandle<T>;
```

What I'm slightly worried about is that people will not catch the difference between the two methods, and use `block_on` instead of `blocking`. And to be fair; this will probably work in most situations -- except performance will be terrible because this is not at all intended for this.

So instead I'm wondering if it would perhaps make sense to move `block_on` to be part of the `thread` submodule. It already has [several other functions related to blocking the current thread](https://doc.rust-lang.org/std/thread/fn.yield_now.html?search=), and this would be another one in that line.

```rust
// current
fn thread::park();
fn thread::park_timeout(dur: Duration);
fn thread::sleep(dur: Duration);
fn thread::yield_now();

// added
fn block_on(fut: impl Future<Output = T>) -> T;
```

A reason why I'm kind of wondering about using the `thread` submodule is because we probably want to give people access to the threadpool to configure it, and I feel `thread` may be a good option to put it (though I have no idea _which_ knobs we'd want to allow people to turn). Which if we have that submodule anyway we may want to consider this direction also. I'm not sure though!

Perhaps another way to go about this would be to make `block_on` fail if called inside of a `task` context:

```rust
use crate::task::worker::get_task;

let task = get_task(|task| task.clone());
debug_assert!(task.is_none(), "`task::block_on()` called inside the context of a task. Did you mean to call `task::blocking?`");
```

Oh also; none of this is pressing at all, and just wanted to bring this up as something that may be interesting to consider. Thanks!

yoshuawuyts:
Okay, riffing on this a bit more. Now that #299 is a thing, things have changed even more as `spawn_blocking` takes a regular closure rather than an async block now:

```rust
fn task::spawn(impl Future) -> JoinHandle;
fn task::spawn_thread(impl FnOnce) -> JoinHandle;
```

So there seems to be somewhat of a parallel here now with `thread`. I think we could do something similar there:
```rust
[...]

PR #299, opened 2019-10-09
Blocking updates

yoshuawuyts:
Renames `task::blocking` to `task::spawn_blocking`, and changes the input type from `impl Future` to `impl FnOnce` so that it becomes clear that this function is for non-async code only.

We received [public feedback on the signature](https://www.reddit.com/r/rust/comments/dffxqv/asyncstd_v0999_released/f33pv8u/), and @stjepang and I spoke privately about the method name and agreed that it was indeed confusing with `block_on` also being a thing.

Because this API has changed, it means that #290 will no longer be able to be called inside `task::blocking`, but instead tasks should rely on `thread::yield_now` instead. This overall feels like a positive to me.

Thanks!

### DIFF
--- a/src/fs/canonicalize.rs
+++ b/src/fs/canonicalize.rs
@@ -32,5 +32,5 @@ use crate::task::blocking;
 /// ```
 pub async fn canonicalize<P: AsRef<Path>>(path: P) -> io::Result<PathBuf> {
     let path = path.as_ref().to_owned();
-    blocking::spawn(async move { std::fs::canonicalize(&path).map(Into::into) }).await
+    blocking::spawn(move || std::fs::canonicalize(&path).map(Into::into)).await
 }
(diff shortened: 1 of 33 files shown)


yoshuawuyts:
As per #289, we may want to consider renaming this to `task::spawn_thread` instead of `task::spawn_blocking` as to have a counterpart to `thread::spawn_task`.

stjepang:
Need to think about this.

I'll just throw in another idea for the signature of this function:

```rust
pub async fn asyncify<F, R>(f: F)
where
    F: FnOnce() -> R + Send + 'static,
    R: Send + 'static,
{
    blocking::spawn(f).await
}
```

Might also consider putting this function into the `async_std::future` module.

Unsure about these ideas though. Need to mull over them :)

PR #300, opened 2019-10-09
Add task::yield_now as "unstable"

### LINKED ISSUE #290: [discussion] Add task::yield_now()?

yoshuawuyts:
Closes #290. Thanks!

## Screenshots

Screenshot_2019-10-10 async_std task yield_now - Rust(1)

### DIFF
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -49,6 +49,13 @@ mod worker;
 
 pub(crate) mod blocking;
 
+cfg_if::cfg_if! {
+    if #[cfg(any(feature = "unstable", feature = "docs"))] {
+        mod yield_now;
+        pub use yield_now::yield_now;
+    }
+}
+
 /// Spawns a blocking task.
 ///
 /// The task will be spawned onto a thread pool specifically dedicated to blocking tasks. This
--- a/src/task/yield_now.rs
+++ b/src/task/yield_now.rs
@@ -1,0 +1,53 @@
+use crate::future::Future;
+use crate::task::{Context, Poll};
+
+use std::pin::Pin;
+
+/// Cooperatively gives up a timeslice to the task scheduler.
+///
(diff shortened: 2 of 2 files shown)


yoshuawuyts:
As was pointed out to me on Twitter, there's prior art for this function in Node's [`setimmediate`](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) call.

chpio:
could this clash with a ["message passing optimization"](https://tokio.rs/blog/2019-10-scheduler/#optimizing-for-message-passing-patterns)?

stjepang:
@chpio not really, we can still do all kinds of optimizations :)

PR #301, opened 2019-10-10
task::block_on -> thread::spawn_task

yoshuawuyts:
Replaces `task::block_on` with `thread::spawn_task`. Ref #289 #299. Not sure if this is a great idea or terrible, but figured I'd open it anyway. We should probably think about this long and hard before we decide to add this.

This also means the addition of a new `thread` submodule.

Oh another gotcha is that because we want to block the current thread, we can't get back a `JoinHandle`. We could jump through hoops to spawn a new thread and get a handle back that way -- but that seems like a somewhat odd choice to make, and I think we should not spawn new threads just for this purpose.

Thanks!

cc/ @stjepang

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,7 +135,7 @@ task::blocking(async {
 
 - Refactored the scheduling algorithm of our executor to use work stealing
 - Refactored the network driver, removing 400 lines of code
-- Removed the `Send` bound from `task::block_on`
+- Removed the `Send` bound from `thread::spawn_task`
 - Removed `Unpin` bound from `impl<T: futures::stream::Stream> Stream for T`
 
 # [0.99.5] - 2019-09-12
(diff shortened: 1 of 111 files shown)


bill-myers:
"spawn" is associated with starting something without waiting for it, but this waits for it.

I think `thread::run_task`, `thread::execute_task` or even just `thread::execute` are better names.

skade:
How about `task::main`, `task::block_thread`? Other then that, I like `thread::execute_task`.

I think we should keep `task` as the main module to use, people will orient towards it for all task spawning needs.

We could have an alias here.

yoshuawuyts:
@skade I think those are both reasonable suggestions. Either way, we shouldn't merge this PR as-is right now so I'm going to go ahead and close it. Let's ponder on this for a bit.

PR #305, opened 2019-10-10
[Draft PR] Adds Stream::le

assemblaj:
Blocked by #268
For issue #129
Based on : https://github.com/rust-lang/rust/blob/master/src/libcore/iter/traits/iterator.rs#L2807


stjepang:
This was added in another PR.

PR #306, opened 2019-10-11
[Draft PR] Adds Stream::lt

assemblaj:
Blocked by #268
For issue #129
Based on : https://github.com/rust-lang/rust/blob/58b54911fabcd1b328fab78014b6833a4e8cfaa3/src/libcore/iter/traits/iterator.rs#L2788


stjepang:
This was added in another PR.

PR #308, opened 2019-10-11
init FutureExt

yoshuawuyts:
Initializes `FutureExt` which allows us to provide extension traits for the `Future` trait. I had to change some requirements inside our code to not use `crate::future::Future` in order to get this to compile during the doc stage.

I'm currently hitting a bug though where adding doc comments to `type Output` and `fn poll` breaks compilation with:

```txt
error: macro expansion ends with an incomplete expression: expected one of `async`, `const`, `extern`, `fn`, `type`, or `unsafe`
   --> src/future/future.rs:101:12
    |
101 |         "#]
    |            ^ expected one of `async`, `const`, `extern`, `fn`, `type`, or `unsafe` here
error: aborting due to previous error
error: could not compile `async-std`.
To learn more, run the command again with --verbose.
```

If anyone has an idea what might be going on there I'd be interested in knowing!

The goal of this PR is to allow us to add more methods to `Future`. Thanks!

### DIFF
--- a/src/fs/dir_builder.rs
+++ b/src/fs/dir_builder.rs
@@ -1,5 +1,7 @@
 use std::future::Future;
 
+use cfg_if::cfg_if;
+
 use crate::io;
 use crate::path::Path;
 use crate::task::blocking;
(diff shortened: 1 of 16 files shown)


stjepang:
I have simplified the `extension_trait` macro and now everything works fine :)

PR #309, opened 2019-10-11
Stream::delay

yoshuawuyts:
Depends on #298. Adds `Stream::delay` which delays the initial yield of a stream by `dur`. This is useful when you want to pass a stream that needs to sleep for a while, rather than instruct people to call `task::sleep`. Also makes for some nice demos of how streamlined Rust can be, which I think is valuable.

Marked as unstable. Thanks!


k-nasa:
@yoshuawuyts  You are a busy person. If I still need this feature, will I continue?
Apparently it is almost complete

yoshuawuyts:
@k-nasa please do!

k-nasa:
@yoshuawuyts I implemented this!
Please review this PR.

src/stream/stream/mod.rs
```diff
@@ -573,6 +575,37 @@ extension_trait! {
...
+            # fn main() { async_std::task::block_on(async {
+            #
+            use async_std::prelude::*;
+            use async_std::stream;
+            use std::time::Duration;
+
+            let mut s = stream::from_iter(vec![0u8, 1, 2]).delay(Duration::from_millis(200));
+
+            assert_eq!(s.next().await, Some(0));
+            assert_eq!(s.next().await, Some(1));
```
yoshuawuyts: Should we assert at least 200 millis have passed between the call to the first `next` and when it actually resolves?

yoshuawuyts:
@k-nasa this patch looks 💯. I can't sign off on it through the GitHub UI because I started it. But think it should be good to merge — just had one question.

src/stream/stream/mod.rs
k-nasa: Yah! You're right. 
I added delay time test with this commit.
fb1fb6c903fca2b68871706b873f3f548148b4b4

PR #312, opened 2019-10-12
Add BufRead::split

yoshuawuyts:
Add `BufRead::split`. Ref #131. Thanks!

## Screenshots

2019-10-13-015556_1920x1080

### DIFF
--- a/src/io/buf_read/mod.rs
+++ b/src/io/buf_read/mod.rs
@@ -1,8 +1,11 @@
 mod lines;
 mod read_line;
 mod read_until;
+mod split;
 
 pub use lines::Lines;
+pub use split::Split;
+
 use read_line::ReadLineFuture;
 use read_until::ReadUntilFuture;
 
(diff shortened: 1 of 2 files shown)


src/io/buf_read/mod.rs
```diff
@@ -226,6 +229,57 @@ extension_trait! {
                 read: 0,
             }
         }
+
+        #[doc = r#"
+            Returns a stream over the contents of this reader split on the byte byte.
```
mzji: Looks like here is a duplicated word of 'byte'.

yoshuawuyts: Changed it to byte `byte` to match std's docs: https://doc.rust-lang.org/std/io/trait.BufRead.html#method.split

src/io/buf_read/mod.rs
```diff
@@ -226,6 +229,57 @@ extension_trait! {
...
+            assert_eq!(split_iter.next().await, Some(b"ipsum".to_vec()));
+            assert_eq!(split_iter.next().await, Some(b"dolor".to_vec()));
+            assert_eq!(split_iter.next().await, None);
+            #
+            # Ok(()) }) }
+            ```
+        "#]
+        fn split(self, byte: u8) -> Split<Self>
+        where
+            Self: Unpin + Sized,
```
stjepang: The `Self: Unpin` bound is meaningless here because we're taking `self` by value, i.e. moving it into the function and then into the `Split` structure.

```suggestion
            Self: Sized,
```

PR #314, opened 2019-10-13
Add `Write::by_ref`

yoshuawuyts:
Ref #131, implements `Write::by_ref`. Thanks!

## References
- https://doc.rust-lang.org/std/io/trait.Write.html#method.by_ref

### DIFF
--- a/src/io/write/mod.rs
+++ b/src/io/write/mod.rs
@@ -197,6 +197,36 @@ extension_trait! {
         {
             WriteAllFuture { writer: self, buf }
         }
+
+        #[doc = r#"
+            Creates a "by reference" adaptor for this instance of `Write`.
+
+            The returned adaptor also implements `Write` and will simply borrow this
+            current writer.
+
+            # Examples
+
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
Oh dang yeah; I just remembered the conversation with @dignifiedquire and while this method can be implemented, it cannot be called.

Which is to say: there's no way to disambiguate between `Read::by_ref` and `Write::by_ref` because you'd have to write:

```rust
WriteExt::by_ref(f)
```

Which isn't valid. This needs to be an inherent method on `AsyncRead` and `AsyncWrite` to have it work for us.

stjepang:
We should also add the same method for `Read` and `Stream` :)

yoshuawuyts:
@stjepang oh yeah `Read` already has this, but `Stream` doesn't -- you're right we should add it! We probably can't merge this PR as-is though; this is something that should be part of `futures-core` because of the reasons I mentioned above :disappointed:

PR #319, opened 2019-10-13
Add print, println, eprint, eprintln

yoshuawuyts:
Adds async `println!, eprintln!, print!, eprint!`, and fixes a bug in #311. Depends on #311 landing first. Thanks!

## Screenshots

Screenshot_2019-10-13 async_std - Rust


stjepang:
Are these insta-stable? Can we put them behind the `unstable` feature flag?

yoshuawuyts:
@stjepang will do!

PR #320, opened 2019-10-13
Implement async Path & PathBuf

### LINKED ISSUE #183: [tracking] path
The `std::path::Path` struct has some blocking methods like [`read_dir`](https://doc.rust-lang.org/nightly/std/path/struct.Path.html#method.read_dir) which should be asynchronous in our library.

That means we'll need to have our own `async_std::path::Path` that is convertible to/from `std::path::Path`. Moreover, since `PathBuf` derefs into `Path`, we'll also need an async version of `PathBuf`.

Right now, we're using `std::path::Path` throughout `async-std`, which might cause users to accidentally call blocking functions without realizing. For example:

```rust
let mut dir = fs::read_dir(".").await?;

while let Some(entry) = dir.next().await {
    let path = entry?.path();

    // Ooops, the `path.canonicalize()` call here is blocking!
    //
    // We should use an async version of `PathBuf` here and
    // do `path.canonicalize().await` instead.
    println!("{:?}", path.canonicalize()?);
}
```

### PathBuf

- [ ] `path::PathBuf::as_path`
- [ ] `path::PathBuf::capacity`
- [ ] `path::PathBuf::clear`
- [ ] `path::PathBuf::into_boxed_path`
- [ ] `path::PathBuf::into_os_string`
- [ ] `path::PathBuf::new`
- [ ] `path::PathBuf::pop`
- [ ] `path::PathBuf::push`
- [ ] `path::PathBuf::reserve`
- [ ] `path::PathBuf::reserve_exact`
- [ ] `path::PathBuf::set_extension`
- [ ] `path::PathBuf::set_file_name`
- [ ] `path::PathBuf::shrink_to`
- [ ] `path::PathBuf::shrink_to_fit`
- [ ] `path::PathBuf::with_capacity`
- [ ] `impl<P: AsRef<Path>> FromStream<P> for PathBuf`

### Path

- [ ] `path::Path::ancestors`
- [ ] `path::Path::as_os_str`
- [ ] `path::Path::canonicalize`
- [ ] `path::Path::components`
- [ ] `path::Path::display`
- [ ] `path::Path::ends_with`
- [ ] `path::Path::exists`
- [ ] `path::Path::extension`
- [ ] `path::Path::file_name`
- [ ] `path::Path::file_stem`
- [ ] `path::Path::has_root`
- [ ] `path::Path::into_path_buf`
- [ ] `path::Path::is_absolute`
- [ ] `path::Path::is_dir`
- [ ] `path::Path::is_file`
- [ ] `path::Path::is_relative`
- [ ] `path::Path::iter`
- [ ] `path::Path::join`
- [ ] `path::Path::metadata`
- [ ] `path::Path::new`
- [ ] `path::Path::parent`
- [ ] `path::Path::read_dir`
- [ ] `path::Path::read_link`
- [ ] `path::Path::starts_with`
- [ ] `path::Path::strip_prefix`
- [ ] `path::Path::symlink_metadata`
- [ ] `path::Path::to_path_buf`
- [ ] `path::Path::to_str`
- [ ] `path::Path::to_string_lossy`
- [ ] `path::Path::with_extension`
- [ ] `path::Path::with_file_name`

Wassasin:
Partially closes #183. Tracking PR as quite a lot of methods need to be implemented.

# PathBuf
- [x]  path::PathBuf::as_path
- [x]  path::PathBuf::into_boxed_path
- [x]  path::PathBuf::into_os_string
- [x]  path::PathBuf::new
- [x]  path::PathBuf::pop
- [x]  path::PathBuf::push
- [x]  path::PathBuf::set_extension
- [x]  path::PathBuf::set_file_name

# Path
- [x]  path::Path::ancestors
- [x]  path::Path::as_os_str
- [x]  path::Path::canonicalize
- [x]  path::Path::components
- [x]  path::Path::display
- [x]  path::Path::ends_with
- [x]  path::Path::exists
- [x]  path::Path::extension
- [x]  path::Path::file_name
- [x]  path::Path::file_stem
- [x]  path::Path::has_root
- [x]  path::Path::into_path_buf
- [x]  path::Path::is_absolute
- [x]  path::Path::is_dir
- [x]  path::Path::is_file
- [x]  path::Path::is_relative
- [x]  path::Path::iter
- [x]  path::Path::join
- [x]  path::Path::metadata
- [x]  path::Path::new
- [x]  path::Path::parent
- [x]  path::Path::read_dir
- [x]  path::Path::read_link
- [x]  path::Path::starts_with
- [x]  path::Path::strip_prefix
- [x]  path::Path::symlink_metadata
- [x]  path::Path::to_path_buf
- [x]  path::Path::to_str
- [x]  path::Path::to_string_lossy
- [x]  path::Path::with_extension
- [x]  path::Path::with_file_name

### DIFF
--- a/src/path/path.rs
+++ b/src/path/path.rs
@@ -462,29 +462,6 @@ impl Path {
         fs::metadata(self).await
     }
 
-    /// Queries the metadata about a file without following symlinks.
-    ///
-    /// This is an alias to [`fs::symlink_metadata`].
-    ///
-    /// [`fs::symlink_metadata`]: ../fs/fn.symlink_metadata.html
-    ///
-    /// # Examples
-    ///
-    /// ```no_run
-    /// # fn main() -> std::io::Result<()> { async_std::task::block_on(async {
-    /// #
-    /// use async_std::path::Path;
-    ///
-    /// let path = Path::new("/Minas/tirith");
-    /// let metadata = path.symlink_metadata().await.expect("symlink_metadata call failed");
-    /// println!("{:?}", metadata.file_type());
-    /// #
-    /// # Ok(()) }) }
-    /// ```
-    pub async fn symlink_metadata(&self) -> io::Result<fs::Metadata> {
-        fs::symlink_metadata(self).await
-    }
-
     /// Directly wraps a string slice as a `Path` slice.
     ///
     /// This is a cost-free conversion.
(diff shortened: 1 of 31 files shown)


src/path/path.rs
```diff
@@ -0,0 +1,105 @@
+use std::ffi::OsStr;
+
+use crate::path::PathBuf;
+use crate::{fs, io};
+
+/// This struct is an async version of [`std::path::Path`].
+///
+/// [`std::path::Path`]: https://doc.rust-lang.org/std/path/struct.Path.html
+pub struct Path {
+    inner: OsStr,
```
yoshuawuyts: I'm wondering whether it might perhaps be easier to store a `std::path::PathBuf` internally, and delegate all non-io methods to it?

That way we can prevent having to re-implement some of the trickier logic such as going between `OsStr` and `Path`.

Wassasin: How about using std::path::Path in Path, and std::path::PathBuf in PathBuf?

yoshuawuyts: @wassasin ah oops yeah; that's what I meant :sweat_smile:

Wassasin:
@yoshuawuyts I'm unsure as how to feature gate the unstable `path_buf_capacity` feature within async-std. The remaining methods depend on this feature.

Also the `ancestors` method yields and iterator over `std::path::Path`, instead of our own `Path`. It may be worth considering mapping this iterator and yielding `impl iterator<Item=Path>`.

stjepang:
> I'm unsure as how to feature gate the unstable `path_buf_capacity` feature within async-std. The remaining methods depend on this feature.

Let's just omit unstable functions from the standard library.

> Also the `ancestors` method yields and iterator over `std::path::Path`, instead of our own `Path`. It may be worth considering mapping this iterator and yielding `impl iterator<Item=Path>`.

We should define our own `Ancestors` iterator instead of re-exporting the one from `std`.

yoshuawuyts:
@wassasin thanks so much for all your work on this! This has been great!!

PR #323, opened 2019-10-13
stabilize stream::{FromStream,IntoStream,Extend}

yoshuawuyts:
I'd like to propose we remove the "unstable" marker from `stream::{FromStream,IntoStream,Extend}` and `Stream::collect`. We've proven that we can use these traits to achieve parity with std, and they seem to work really well! Also we've had them for close to a month now without any problems.

I guess this is one of the first times we're moving something large from "unstable" to stable, so I guess this is somewhat defining. My reasoning here is not that we're sure everything is bug-free; but that we're sure that at least the external API is correct and we don't think we'll want to change it. To that extent this seems to pass that test.

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,18 +66,18 @@ cfg_if! {
         pub mod path;
         #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
         pub mod pin;
-
-        mod unit;
-        mod vec;
-        mod result;
-        mod option;
-        mod string;
-        mod collections;
     }
 }
 
-mod macros;
 pub(crate) mod utils;
+
+mod collections;
+mod macros;
+mod option;
+mod result;
+mod string;
+mod unit;
+mod vec;
 
 #[doc(inline)]
 pub use std::{write, writeln};
--- a/src/stream/extend.rs
+++ b/src/stream/extend.rs
@@ -27,7 +27,6 @@ use crate::stream::IntoStream;
 /// #
 /// # }) }
 /// ```
-#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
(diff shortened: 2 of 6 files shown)


stjepang:
Should any of these traits live in the `futures` crate? I recall there was one trait that *had to* live there but forgot which one was it?

yoshuawuyts:
> Should we wait for futures to create the IntoStream trait since we can't do as much as we'd like until then?

I think there's value in defining our own now. Besides it's required for `collect` to work. If `futures` defines one we can move over to using that instead; it wouldn't be a breaking change.

stjepang:
I will take another look at this but a quick question:

Since there is a method `stream_extend()`, shouldn't the trait be named `StreamExtend`? Generally, single-method traits are named after the method. I'm also fine with the name `Extend` but felt a mild inconsistency. On the other hand, `Extend` follows std library more closely, but since we've had to rename `extend()` to `stream_extend()`... I wonder what you think?

yoshuawuyts:
@stjepang there was a reason for it I believe because the trait names conflict and we couldn't distinguish between the two otherwise.

stjepang:
Right, perhaps my question wasn't clear enough. :)

Typically, if a trait has just a single method `stream_extend()`, it would be named `StreamExtend`. Just like we have `Drop::drop()` or `From::from()`, or `ToString::to_string()` and so on.

In that sense `Extend::stream_extend()` feels a bit odd to me whereas `StreamExtend::stream_extend()` feels a bit better, so I was curious if you think similarly?

I'll the decision to you, but just wanted to point that out.

yoshuawuyts:
We have come to a resolution of the naming of `Extend` in #464.

PR #326, opened 2019-10-14
add stream::FusedStream as "unstable"

yoshuawuyts:
Implements `Stream::FusedStream`. This is a [different API from futures-rs](https://docs.rs/futures-core-preview/0.3.0-alpha.19/futures_core/stream/trait.FusedStream.html) that instead is closer to std, and resembles the API futures-rs had prior to `0.3.0-alpha.10` (added in https://github.com/rust-lang-nursery/futures-rs/pull/1259, October 2018).

Because we don't require `select! {}` for stream multiplexing in async-std, the `is_terminated` method doesn't really serve a purpose for us, and really only requires extra bookkeeping.

Because `FusedStream` is used for optimization purposes only, I think it's not a big deal to diverge from futures-rs on this account.

Thanks!

### DIFF
--- a/src/stream/fused_stream.rs
+++ b/src/stream/fused_stream.rs
@@ -1,0 +1,21 @@
+use crate::stream::Stream;
+
+/// A stream that always continues to yield `None` when exhausted.
+///
+/// Calling next on a fused stream that has returned `None` once is guaranteed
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
Ref #129

PR #328, opened 2019-10-14
Re export macros from submodules

yoshuawuyts:
Depends on https://github.com/async-rs/async-macros/pull/7. Re-exports various macros from their respective submodules. We were talking about this, and mostly wanted to test out how this feels to use.

Feedback would be great!

## Screenshots

Screenshot_2019-10-14 async_std task - Rust
Screenshot_2019-10-14 async_std io - Rust
Screenshot_2019-10-14 async_std - Rust


stjepang:
Let's keep this PR on hold. I'm having mixed feelings about it :)

yoshuawuyts:
I believe nobody's really pushing for this anymore, so I'm going to go ahead and close this. Thanks!

PR #332, opened 2019-10-15
impl Stream::take_while adapter

starsheriff:
Hei,
my first take on adding a stream adapter.

For issue #129

### DIFF
--- a/src/stream/stream/mod.rs
+++ b/src/stream/stream/mod.rs
@@ -265,9 +265,9 @@ extension_trait! {
             # }) }
         "#]
         fn take_while<P>(self, predicate: P) -> TakeWhile<Self, P, Self::Item>
-            where
-                Self: Sized,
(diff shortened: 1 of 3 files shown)


src/stream/stream/mod.rs
```diff
@@ -241,6 +243,35 @@ extension_trait! {
...
+            assert_eq!(s.next().await, Some(2));
+            assert_eq!(s.next().await, None);
+
+            #
+            # }) }
+        "#]
+        fn take_while<P>(self, predicate: P) -> TakeWhile<Self, P, Self::Item>
+            where
+                Self: Sized,
+                P: FnMut(&Self::Item) -> bool,
```
stjepang: nit: can we unindent these three lines to keep the same code formatting as in other methods? :)

starsheriff: sure

Issue #229, opened 2019-09-21
Locking for stdin


jonathandturner:
In sync std, you can lock stdin to ensure you maintain control of it.  Something like:

```rust
let stdin = std::io::stdin();
let locked = stdin.lock();
```

Currently, there isn't a way to do the equivalent in async-std.

cc @yoshuawuyts

k-nasa:
I want to try this implementation!
But I can't imagine what kind of implementation should be done (due to my lack of ability)
@yoshuawuyts  Could you show me a little way?

yoshuawuyts:
@k-nasa yay, thanks for volunteering!

So there are two important things to be aware of here:
- [the way `std::io::Stdin::lock` works](https://doc.rust-lang.org/std/io/struct.Stdin.html#method.lock)
- [the way `async_std::io::Stdin` is implemented](https://github.com/async-rs/async-std/blob/master/src/io/stdin.rs#L47)

```rust
// definition in async-std, `State` holds a reference to std's Stdin
pub struct Stdin(Mutex<State>);
```

We want to have an `async fn lock` on `Stdin` that returns a new struct `StdinLock` which implements `Read` and `BufRead`.

The way I'd approach this by making our method acquire outer async lock, then call `lock` on the reference to `std::io::Stdin` and borrow std's `StdinLock`. Then from there we can forward calls, to it, and make sure we properly handle `io::error::WouldBlock`. Example code for this should be part of our `Stdin` code already.

Hope this provides a good starting point; let me know if you need any more pointers!

stjepang:
To add a little bit to that, since `std::io::Stdin::lock()` blocks, we would have to invoke that function inside `task::blocking()`.

Moreover, the returned `StdinLock<'_>` will hold a reference to the `Stdin` instance, so we should probably make sure that it's `'static` by putting it into `lazy_static!`:

```rust
lazy_static! {
    static ref STDIN: Stdin = std::io::stdin();
}
```

A locking function might then look like this:

```rust
use crate::task::blocking;

async fn lock() -> StdinLock<'static> {
    blocking::spawn(async move { STDIN.lock() }).await
}
```

k-nasa:
@yoshuawuyts @stjepang Thx!! I try it.

yoshuawuyts:
ref #131

k-nasa:
Apparently, StdinLock does not implement the Send trait. So it doesn't seem to work with blocking :: spawn.

```
--> src/io/stdin.rs:167:9
[...]

PR #334, opened 2019-10-15
Locking for stdin

k-nasa:
Implement 
- Stdin::lock
- Stdout::lock
- Stderr::lock

ref: https://github.com/async-rs/async-std/issues/229


stjepang:
Thanks for the PR! I think we should make async versions of `StdinLock`, `StdoutLock`, and `StderrLock`. If we lock an async I/O handle, we should also get back an async I/O handle.

Also, methods like `std::io::StdinLock::lock()` are blocking operations, so we should perhaps spawn a blocking task that acquires the lock.

src/io/stdin.rs
```diff
@@ -46,6 +47,11 @@ pub fn stdin() -> Stdin {
 #[derive(Debug)]
 pub struct Stdin(Mutex<State>);
 
+#[derive(Debug)]
+pub struct StdinLock<'a>(std::io::StdinLock<'a>);
+
+unsafe impl Send for StdinLock<'_> {}
```
k-nasa: @stjepang I changed only StdinLock.
How about such an implementation ❓

src/io/stdin.rs
```diff
@@ -218,3 +255,13 @@ cfg_if! {
...
     }
 }
+
+impl Read for StdinLock<'_> {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        _cx: &mut Context<'_>,
+        _buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        unimplemented!()
```
k-nasa: Is this better to mimic the Stdin implementation, or?

yoshuawuyts: This should forward the inner read impl so we can actually read the bytes from Stdin. So, yes! :D

src/io/stdin.rs
yoshuawuyts: Looks good! The only thing missing here is a doc comment now I believe!

src/io/stdin.rs
k-nasa: thanks !

src/io/stderr.rs
```diff
@@ -179,3 +214,21 @@ cfg_windows! {
...
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        unimplemented!()
+    }
+
+    fn poll_close(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        unimplemented!()
+    }
+}
```
yoshuawuyts: These impls should forward to the inner `write` impl.

src/io/stdout.rs
```diff
@@ -179,3 +214,21 @@ cfg_windows! {
...
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        unimplemented!()
+    }
+
+    fn poll_close(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        unimplemented!()
+    }
+}
```
yoshuawuyts: These impls should forward to the inner `Write` impl

src/io/stdin.rs
yoshuawuyts: This needs to forward to the inner `Read` impl.

k-nasa:
Yh!
However, I am a little worried about what to implement.
please wait a moment.

k-nasa:
Oh. I accidentally erased the branch...

k-nasa:
It is probably CI upset that CI will fail.
I believe this PR will be solved.
https://github.com/async-rs/async-std/pull/414

yoshuawuyts:
@k-nasa merged the branch you linked!

src/io/stderr.rs
```diff
@@ -87,6 +99,33 @@ enum Operation {
...
+    /// use async_std::prelude::*;
+    ///
+    /// let stderr = io::stderr();
+    /// let mut handle = stderr.lock().await;
+    ///
+    /// handle.write_all(b"hello world").await?;
+    /// #
+    /// # Ok(()) }) }
+    /// ```
+    pub async fn lock(&self) -> StderrLock<'static> {
```
yoshuawuyts: Ah dang, I just realized something I didn't spot before: we should probably mark the APIs introduced in this patch as "unstable". `async/await` is coming up real soon, and our 1.0 release is coming with it. By marking it as "unstable" we give ourselves some time to fix any bugs we might encounter.

Sorry for spotting this so late! -- Otherwise this PR is looking :100: :100: :100:

PR #344, opened 2019-10-15
Implemented StreamExt::try_fold

Wassasin:
See #129. The std::iter variants of `try_fold` and `try_for_each` get the iterator `&mut`, with the note that on failure the iterator can be resumed. Might it be an idea to also get this behaviour into async-std?


yoshuawuyts:
@wassasin if we could get that to work that'd be fantastic!

PR #347, opened 2019-10-16
add Stream::last

starsheriff:
I have troubles getting the assignment to `self.last` working. I am not so familiar with _pinning_, so any hint would be appreciated :grimacing: 

Ref #129


src/stream/stream/last.rs
```diff
@@ -0,0 +1,42 @@
...
+{
+    type Output = Option<S::Item>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let next = futures_core::ready!(self.as_mut().stream().poll_next(cx));
+
+        match next {
+            Some(new) => {
+                cx.waker().wake_by_ref();
+                *self.as_mut().last() = Some(new);
```
starsheriff: This assignment does not compile, how can I get that working. The `MinByFuture` does exactly the same, but obviously I am missing _something_ but can't see it.

starsheriff: @yoshuawuyts @stjepang

starsheriff: Compile Error:
```
error[E0594]: cannot assign to data in a dereference of `std::pin::Pin<&mut std::option::Option<<S as futures_core::stream::Stream>::Item>>`
  --> src/stream/stream/last.rs:36:17
   |
36 |                 *self.as_mut().last() = Some(new);
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot assign
   |
   = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `std::pin::Pin<&mut std::option::Option<<S as futures_core::stream::Stream>::Item>>`

error: aborting due to previous error

error: could not compile `async-std`.
```

src/stream/stream/last.rs
```diff
@@ -0,0 +1,42 @@
...
+#[doc(hidden)]
+#[allow(missing_debug_implementations)]
+pub struct LastFuture<S, T> {
+    stream: S,
+    last: Option<T>,
+}
+
+impl<S, T> LastFuture<S, T> {
+    pin_utils::unsafe_pinned!(stream: S);
+    pin_utils::unsafe_pinned!(last: Option<T>);
```
stjepang: If you make `last` unpinned, the code will compile:

```suggestion
    pin_utils::unsafe_unpinned!(last: Option<T>);
```

This is okay to do because `last` is never really pinned.

starsheriff: Nice, thanks!

stjepang:
Looks like this just needs a `cargo fmt` now :)

src/stream/stream/mod.rs
```diff
@@ -456,6 +458,54 @@ extension_trait! {
...
+            let last  = s.last().await;
+            assert_eq!(last, None);
+            #
+            # }) }
+            ```
+            
+        "#]
+        fn last(
+            self,
+        ) -> impl Future<Output = Option<Self::Item>> + '_ [LastFuture<Self, Self::Item>]
```
stjepang:
```suggestion
        ) -> impl Future<Output = Option<Self::Item>> [LastFuture<Self, Self::Item>]
```

src/stream/stream/mod.rs
```diff
@@ -456,6 +458,54 @@ extension_trait! {
...
+            let last  = s.last().await;
+            assert_eq!(last, None);
+            #
+            # }) }
+            ```
+            
+        "#]
+        fn last(
+            self,
+        ) -> impl Future<Output = Option<Self::Item>> + 'static [LastFuture<Self, Self::Item>]
```
stjepang: The `+ 'static` is not necessary (doesn't have an effect in this context) and we don't use it other similar methods :)

```suggestion
        ) -> impl Future<Output = Option<Self::Item>> [LastFuture<Self, Self::Item>]
```

src/stream/stream/mod.rs
starsheriff: right, the lifetime is not needed. I got a bit lost in the type signature I guess :grimacing:

stjepang: No worries, thanks! :)

src/stream/stream/last.rs
taiki-e: FYI: If you want to assign to the pinned value, you can use [`Pin::set`](https://doc.rust-lang.org/nightly/core/pin/struct.Pin.html#method.set).

```rust
self.as_mut().last().set(Some(new));
```

starsheriff: right, I have to take a proper look at pinning. I haven't used it before.

Issue #340, opened 2019-10-15
Add Stream::timeout


yoshuawuyts:
Similar to the `future::timeout` free function we should add a `Stream::timeout` method. This method should be marked as "unstable".

## Example

```rust
// This example will never actually time out, but you get the idea.
let s = stream::repeat(1u8).take(100);
let s = s.timeout(Duration::from_secs(1));
while let Some(_) = s.next().await {
    dbg!();
}

PR #348, opened 2019-10-16
Add stream timeout

k-nasa:
close: https://github.com/async-rs/async-std/issues/340

Currently, the process that times out cannot be tested.
Is there any stream processing that takes time easily?


src/stream/stream/timeout.rs
```diff
@@ -0,0 +1,57 @@
...
+            },
+        }
+    }
+}
+
+/// An error returned when a stream times out.
+#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
+#[cfg(any(feature = "unstable", feature = "docs"))]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct TimeoutError;
```
yoshuawuyts: Should we export this too?

stjepang: Also, this type should not be constructible by end-users. To do prevent users from being able to create instances of `TimeoutError`, do:

```rust
pub struct TimeoutError {
    _private: (),
}
```

k-nasa: I fixed it. https://github.com/async-rs/async-std/pull/348/commits/b6edd123cb26bd93e06eda1d9f999bdb17fd074b

src/stream/stream/mod.rs
```diff
@@ -104,13 +104,16 @@ cfg_if! {
...
+        mod timeout;
 
         use std::pin::Pin;
+        use std::time::Duration;
 
         use crate::future::Future;
         use crate::stream::FromStream;
 
         pub use merge::Merge;
+        pub use timeout::TimeoutStream;
```
```suggestion
        pub use timeout::{TimeoutStream, Timeouterror};
```

src/stream/stream/timeout.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::pin::Pin;
+use std::time::Duration;
+
+use futures_timer::Delay;
+
+use crate::future::Future;
+use crate::stream::Stream;
+use crate::task::{Context, Poll};
+
+#[doc(hidden)]
```
yoshuawuyts: Remove `#[doc(hidden)]`, and add a doc comment linking back to `Stream::timeout`.

src/stream/stream/timeout.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Debug)]
+pub struct TimeoutStream<S> {
```
yoshuawuyts: We should rename this to `Timeout` so it can be exported + enumerated from the list of streams, similar to all the other streams.

src/stream/stream/mod.rs
```diff
@@ -1044,6 +1047,40 @@ extension_trait! {
...
+            while let Some(v) = s.next().await {
+                assert_eq!(v, Ok(1));
+            }
+            #
+            # Ok(()) }) }
+            ```
+        "#]
+        #[cfg(any(feature = "unstable", feature = "docs"))]
+        #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
+        fn timeout(self, dur: Duration) -> TimeoutStream<Self>
```
stjepang: We don't add the "Stream" suffix to stream types so this should be just `Timeout<Self>`. See other similar stream types for reference (e.g. `Empty`, `Repeat`, `Chain`).

k-nasa:
@yoshuawuyts @stjepang 
Fixed it and rebased the master branch.
Can you review it again?

Issue #341, opened 2019-10-15
Add Future::delay


yoshuawuyts:
Similar to https://github.com/async-rs/async-std/pull/309 we should add a method to `Future` that delays execution for a `Duration`. This method should be marked as "unstable".

This is useful in some cases where you want to define a `Future` that has a delay, rather than combining `task::sleep` with an async block and another future.

## Examples
```rust
use async_std::future;

let a = future::ready(1).delay(Duration::from_millis(200));
let b = future::ready(2).delay(Duration::from_millis(100));
let c = future::ready(3).delay(Duration::from_millis(300));

dbg!(future::join!(a, b, c).await);
```

This is simlilar to this, but less nice to write:

```rust
let a = async {
    task::sleep(Duration::from_millis(200).await;
    future::ready(1).await
};
let b = async {
    task::sleep(Duration::from_millis(100).await;
    future::ready(2).await
};
let c = async {
    task::sleep(Duration::from_millis(300).await;
    future::ready(3).await
};

dbg!(future::join!(a, b, c).await);

matt961:
I'll give it a try, not sure how long it will take me so anyone else feel free to make a PR  in the mean time :P

PR #349, opened 2019-10-16
Add future::delay

k-nasa:
close: https://github.com/async-rs/async-std/issues/341


src/future/delay.rs
```diff
@@ -0,0 +1,61 @@
...
+/// Creates a future that is delayed before it starts yielding items.
+///
+/// # Examples
+///
+/// ```
+/// # async_std::task::block_on(async {
+/// use async_std::future;
+/// use std::time::Duration;
+
+/// let a = future::delay(future::ready(1) ,Duration::from_millis(2000));
```
yoshuawuyts: I was thinking we make this method part of FutureExt (similar to how we define new stream methods) rather than expose it as a free function

k-nasa: Like this?   b251fc9

yoshuawuyts: yeah, that's closer!

src/future/future.rs
```diff
@@ -105,6 +114,28 @@ extension_trait! {
...
+        /// use std::time::Duration;
+        /// use async_std::future::FutureExt;
+        ///
+        /// let a = future::ready(1).delay(Duration::from_millis(2000));
+        /// dbg!(a.await);
+        /// # })
+        /// ```
+        #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
+        #[cfg(any(feature = "unstable", feature = "docs"))]
+        fn delay(self, dur: Duration) -> DelayFuture<Self>
```
yoshuawuyts: We should obfuscate the type here, as we're awaiting `async fn` in traits to be released. Look at the `ret!` macro in the stream module

src/future/mod.rs
```diff
@@ -51,6 +51,7 @@ pub use async_macros::{select, try_select};
 use cfg_if::cfg_if;
 
 pub use future::Future;
+pub use future::FutureExt;
```
yoshuawuyts: We shouldn't export the `FutureExt` trait directly here; instead it should be part of the `prelude`, similar to `StreamExt`.

k-nasa:
I fixed it and rebase master branch.
Can you review again?

src/future/future.rs
skade: If a concrete type is returned from a trait, people might start relying on it, making this a compatibility hazard towards a future `impl Future...` version.

src/future/future.rs
```diff
@@ -99,6 +107,28 @@ extension_trait! {
     }
 
     pub trait FutureExt: std::future::Future {
+        /// Creates a future that is delayed before it starts yielding items.
```
skade: Looking at the implementation, I'm not sure if this documentation is correct.

To my reading, it delays _execution_ of the future for the time time given. It will never be polled before the delay resolves. (which might or might not be what you want)

k-nasa: fixed. I'm not fluent in English, so it might still be wrong ...

src/future/future/delay.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use futures_timer::Delay;
+use pin_project_lite::pin_project;
+
+use crate::future::Future;
+use crate::task::{Context, Poll};
+
+pin_project! {
+#[doc(hidden)]
+#[derive(Debug)]
```
yoshuawuyts:
```suggestion
    #[derive(Debug)]
```

src/future/future/delay.rs
```diff
@@ -0,0 +1,43 @@
...
+use std::time::Duration;
+
+use futures_timer::Delay;
+use pin_project_lite::pin_project;
+
+use crate::future::Future;
+use crate::task::{Context, Poll};
+
+pin_project! {
+#[doc(hidden)]
```
yoshuawuyts:
```suggestion
    #[doc(hidden)]
```

k-nasa:
Thanks!

PR #355, opened 2019-10-16
expose std::sync::atomic

yoshuawuyts:
This exposes `async_std::sync::atomic`. If we want to document the `sync` submodule similar to std, we will need to expose this. I feel it's also nice to continue to move towards parity with std.

Thanks!

### DIFF
--- a/src/sync/atomic/mod.rs
+++ b/src/sync/atomic/mod.rs
@@ -1,0 +1,122 @@
+//! Atomic types
+//!
+//! Atomic types provide primitive shared-memory communication between
+//! threads, and are the building blocks of other concurrent
+//! types.
+//!
+//! This module defines atomic versions of a select number of primitive
+//! types, including [`AtomicBool`], [`AtomicIsize`], [`AtomicUsize`],
+//! [`AtomicI8`], [`AtomicU16`], etc.
+//! Atomic types present operations that, when used correctly, synchronize
+//! updates between threads.
+//!
+//! [`AtomicBool`]: struct.AtomicBool.html
+//! [`AtomicIsize`]: struct.AtomicIsize.html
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
@stjepang I did that, and it also exposes all the deprecated items. Figured this may look a bit cleaner (:

nbdd0121:
What if an architecture has some atomics (e.g. AtomicU64 on 32-bit platforms) missing?

taiki-e:
@nbdd0121 The current async-std cannot be compiled with those targets (see https://github.com/async-rs/async-std/pull/141).

stjepang:
I'm a bit on the fence here. My impression is that atomics are not very common in async code and are kind of low-level (memory orderings, how do they work?) so I don't know if they belong into `async-std`.

nbdd0121:
> @nbdd0121 The current async-std cannot be compiled on those targets (see #141).

But I believe that this project intends to support those platforms, so we'd better fix that issue and not introducing new platform-compatibility issues.

yoshuawuyts:
@stjepang I tend to err on the side of "eh, let's add it" if it directly re-exports a part of std in an existing submodule. I wanted to port std's docs, which talk about Atomics a bunch so figured we might as well export them 😅

taiki-e:
> But I believe that this project intends to support those platforms, so we'd better fix that issue and not introducing new platform-compatibility issues.

FYI: #286 fixes that issue.

taiki-e:
I personally prefer to only include things that are related to async code.

Also, `alloc::sync` does not reexport `core::sync::atomic` for similar reasons. (see https://github.com/rust-lang/rust/pull/58175)

yoshuawuyts:
I don't feel too strongly about this. I'm fine to close this.

PR #356, opened 2019-10-16
Implemented StreamExt::throttle

### LINKED ISSUE #342: Add Stream::throttle
We should add a `Stream::throttle` method as "unstable", similar to [RxJs's `throttle` function](https://www.learnrxjs.io/operators/filtering/throttle.html).

The goal of this function is to limit the amount of items yielded per timeslice in a stream. This is usually considered a somewhat brute tool, but it does have its use and we should give people access to this.

## Examples
```rust
// emit value every 1 second
let s = stream::interval(Duration::from_secs(1)).enumerate();

// throttle for 2 seconds, emit latest value
// in this case it means we skip two iterations, and let through the third
let s = s.throttle(Duration::from_secs(2));

s.for_each(|(_, n)| dbg!(n)).await;
// => 0 .. 3 .. 6 .. 9
```

## References
- https://www.learnrxjs.io/operators/filtering/throttle.html
- https://rxjs.dev/api/operators/throttle

Wassasin:
Closes #342. Still need to write documentation and tests.


yoshuawuyts:
@Wassasin as per @binarybana's suggestion in #342 we should probably not drop any messages in `throttle` itself, but instead leave it up to the stream source to implement the queueing strategy. This was an oversight on my part. Would you be willing to update the implementation to match that?

Wassasin:
Certainly. I'll work on it during the impl days. ;)

examples/throttle.rs
```diff
@@ -0,0 +1,27 @@
...
+
+            s.for_each(|(n, _)| {
+                dbg!(n);
+            })
+            .await;
+            // => 0 .. 1 .. 2 .. 3
+            // with a pause of 2 seconds between each print
+        })
+    }
+}
```
yoshuawuyts: If we could convert this to an actual test that would be ideal. I don't think it might be a bit too specific to have as a standalone example like this.

src/stream/stream/mod.rs
```diff
@@ -313,6 +315,41 @@ extension_trait! {
...
+
+            # Examples
+            ```ignore
+            # fn main() { async_std::task::block_on(async {
+            #
+            use async_std::stream;
+            use std::time::Duration;
+
+            // emit value every 1 second
+            let s = stream::interval(Duration::from_nanos(1000000)).enumerate();
```
yoshuawuyts: Perhaps if we use `stream::interval(Duration::from_millis(5)).take(5))` or something similar we could convert this into an automated test. It'd be great if we could remove the `no_run` tag at the top so it tests the functionality.

src/stream/stream/mod.rs
```diff
@@ -313,6 +315,41 @@ extension_trait! {
...
+
+            // throttle for 2 seconds
+            let s = s.throttle(Duration::from_secs(2));
+
+            s.for_each(|(n, _)| {
+                dbg!(n);
+            })
```
yoshuawuyts: We could use `.fold` instead of `for_each` here to gather the total time waited into a single duration. Then after we're done running we can verify that the total time elapsed was within a certain bound.

src/stream/stream/throttle.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::time::Duration;
+
+use futures_timer::Delay;
+use pin_project_lite::pin_project;
+
+use crate::stream::Stream;
+use crate::task::{Context, Poll};
+
+pin_project! {
+    /// A stream that only yields one element once every `duration`, and applies backpressure. Does not drop any elements.
```
yoshuawuyts: We should probably wrap around this comment

yoshuawuyts:
```suggestion
    /// A stream that only yields one element once every `duration`.
    ///
    /// This `struct` is created by the [`throttle`] method on [`Stream`]. See its
    /// documentation for more.
    ///
    /// [`throttle`]: trait.Stream.html#method.throttle
    /// [`Stream`]: trait.Stream.html
```

src/stream/stream/mod.rs
```diff
@@ -313,6 +315,41 @@ extension_trait! {
             TakeWhile::new(self, predicate)
         }
 
+        #[doc = r#"
+            Limit the amount of items yielded per timeslice in a stream.
+
```
yoshuawuyts: It'd be nice to add some more information that this stream does not drop items, and is intended for backpressure only.

```suggestion

This stream does not drop any items, but will only limit the rate at which items pass through.
```

src/stream/stream/throttle.rs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        match this.stream.poll_next(cx) {
+            Poll::Pending => {
+                cx.waker().wake_by_ref(); // Continue driving even though emitting Pending
+                Poll::Pending
+            }
+            Poll::Ready(None) => Poll::Ready(None),
+            Poll::Ready(Some(v)) => {
+                this.delay.set(Some(Delay::new(*this.duration)));
```
yoshuawuyts: You can use [`Delay::reset`](https://docs.rs/futures-timer/2.0.0/futures_timer/struct.Delay.html#method.reset) here instead to be slightly more efficient.

yoshuawuyts: We could take this even further, and always keep a delay around, but add a boolean flag to keep track of whether the delay is currently active. That way the delay never gets deallocated + unregistered through `this.delay.set(None);`, and we'll see some overall perf gain.

Wassasin:
Even with 50ms/150ms (450ms max) the CI-test fails. Apparantly the CI-server is really really slow. Any ideas on how to fix it? I am reluctant to increase the max exec time to > 1 second for the test.

yoshuawuyts:
@Wassasin perhaps all we should do is test the lower bounds. Ensure that *at least* n millis have passed, which is the only hard guarantee we provide about the throttle duration anyway.

Wassasin:
Testing on the CI is very insightful. Everything should be OK now.

examples/throttle.rs
yoshuawuyts: I still don't think we quite need this example here, but I'll file a follow-up PR to do so. Thanks heaps!

PR #357, opened 2019-10-16
Enable tests on CI

taiki-e:
As mentioned in https://github.com/async-rs/async-std/pull/286#discussion_r331923531, if cargo test contains --doc arguments, only doc test is executed.

### DIFF
--- a/tests/buf_writer.rs
+++ b/tests/buf_writer.rs
@@ -48,13 +48,13 @@ fn test_buffered_writer() {
 }
 
 #[test]
-fn test_buffered_writer_inner_into_inner_does_not_flush() {
+fn test_buffered_writer_inner_into_inner_flushes() {
     task::block_on(async {
         let mut w = BufWriter::with_capacity(3, Vec::new());
         w.write(&[0, 1]).await.unwrap();
         assert_eq!(*w.get_ref(), []);
         let w = w.into_inner().await.unwrap();
-        assert_eq!(w, []);
+        assert_eq!(w, [0, 1]);
     })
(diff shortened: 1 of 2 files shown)


taiki-e:
Some tests seem broken (because they were merged without being tested on CI).

tests/buf_writer.rs
```diff
@@ -47,6 +47,7 @@ fn test_buffered_writer() {
     })
 }
 
+#[ignore]
```
stjepang: I'm curious, what's the error here?

taiki-e: This assertion is failed:
https://github.com/async-rs/async-std/blob/ca80ca981ea55374a5b1d29819fe3d2e337dd916/tests/buf_writer.rs#L57

Error:
```text
---- test_buffered_writer_inner_into_inner_does_not_flush stdout ----
thread 'test_buffered_writer_inner_into_inner_does_not_flush' panicked at 'assertion failed: `(left == right)`
  left: `[0, 1]`,
 right: `[]`', tests/buf_writer.rs:58:9
```

I think this is because #315 changed the behavior of lnto_inner: https://github.com/async-rs/async-std/pull/315/files#diff-a2b217372e599101d2b0f1dbab63ee2eR205

taiki-e:
> could you rename `test_buffered_writer_inner_into_inner_does_not_flush` to `test_buffered_writer_inner_into_inner_flushes`? :)

Done.

> Feel free to merge the PR after the change.

I don't have merge right...

PR #358, opened 2019-10-16
Make sure each invocation of block_on uses its own Parker

stjepang:
This fixes a bug where we have recursive (nested) invocations of `block_on`:

```rust
task::block_on(async {
    task::block_on(async {
        // ...
    })
})
```

Each `block_on` behaves as an independent task and has an associated waker.

The problem was that both of these `block_on` invocations reused the same `Thread` descriptor for wakeups. If someone called `.wake()`, we had no way of distinguishing which of the two `block_on` tasks was actually woken.

Sometimes the outer task should get woken but the inner task would pick up the wakeup because it took over the `Thread` descriptor.

The problem is solved by associating a distinct `Parker` with each `block_on` invocation so that they don't get shared among recursive calls. Note that a `Parker` is just an implementation of `std::thread::park()` and `std::thread::unpark()`, except it was extracted out and moved into `crossbeam-utils`.

This should fix deadlocks @svartalf and @jebrosen have reported this week. Can you perhaps confirm the problem goes away with this branch?


jebrosen:
Yes, this branch does appear to fix the deadlock I observed.

svartalf:
Yes, I can confirm that it resolves the issue I had. Great work, as usual!

PR #359, opened 2019-10-17
Add stream successors

k-nasa:
Ref: https://github.com/async-rs/async-std/issues/129


stjepang:
We should make `successors` take an `async fn` rather than a normal closure like in #363. :) I would accept this PR since it was submitted first, but feel free to work with @felipesere together on it if you want to!

felipesere:
@k-nasa would you mind if we merge mine? 😺

k-nasa:
That looks better!
OK!

PR #361, opened 2019-10-17
Cleanup: replace cfg-if with our macros

stjepang:
This is a negative diff that only reshuffles code and cleans it up.

I added a few convenience macros that replace all uses of `cfg_if!` with new macros `unstable!`, `unix!`, `windows!`, `docs!`, and `not_docs!`.


yoshuawuyts:
@stjepang maybe prefixing the macro names with `cfg_` would make it clearer what's going on for people not familiar with them? E.g. `cfg_unix!`

stjepang:
@yoshuawuyts Sounds good! I added the `cfg_` prefix to the new macros.

src/utils.rs
```diff
@@ -20,6 +20,69 @@ pub fn abort_on_panic<T>(f: impl FnOnce() -> T) -> T {
     t
 }
 
+/// Declares unstable items.
+#[doc(hidden)]
+#[macro_export]
```
taiki-e: Why are these macros exported?

I think it is enough to use `#[macro_use]` for the `utils` module if these macros are only used in this crate.
```rust
#[macro_use]
mod utils;
```

stjepang: This doesn't work - we have to keep `#[macro_export]` or else the macro won't be visible from other modules. At least that's my understanding.

Do you know how to make this work without `#[macro_export]`?

taiki-e: I was able to make it work by doing the following. (see https://github.com/stjepang/async-std/pull/1)
* Apply `#[macro_use]` to the module where the macro is defined.
* Remove `crate::` from all internal macro calls.
* Remove `#[macro_export]` from the internal macro.

stjepang:
@taiki-e Thanks for removing `#[macro_export]`! If this looks good now, feel free to merge.

PR #363, opened 2019-10-17
Async successors

felipesere:
This doesn't compile at the moment, but I felt it was easier to get pointers on code than hypotheticals in Discord.

The current compile error is:
```
error[E0599]: no method named `poll` found for type `Fut` in the current scope
  --> src/stream/stream/successor.rs:51:52
   |
51 |         match self.as_mut().successor()(self.next).poll(cx) {
   |                                                    ^^^^ method not found in `Fut`
   |
   = help: items from traits can only be used if the type parameter is bounded by the trait
help: the following trait defines an item `poll`, perhaps you need to restrict type parameter `Fut` with it:
   |
41 | impl <F, Fut: core::future::future::Future, T> Stream for Successor<F, Fut, T>
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

which confuses me because I thought I had that trait bound on the `where`?


stjepang:
I think the error is saying you're calling `.poll(cx)` on a `Fut` whereas you should be calling it on a `Pin<&mut Fut>`, i.e. the future is not pinned.

We should store the future returned by `successor(next)` into the `Successor` struct and then poll it. Note that if polling returns `Poll::Pending`, later on we'll want to poll that same future again instead of asking `successor(next)` to give a fresh future.

felipesere:
Anything I’d we would need to do to get this merged? 🤩

src/stream/mod.rs
```diff
@@ -29,6 +29,7 @@ pub use repeat_with::{repeat_with, RepeatWith};
 pub use stream::{
     Chain, Filter, Fuse, Inspect, Scan, Skip, SkipWhile, StepBy, Stream, Take, TakeWhile, Zip,
 };
+pub use successor::{successor, Successor};
```
stjepang: Let's rename "successor" to "successors" :)

```suggestion
pub use successors::{successors, Successors};
```

src/stream/successor.rs
```diff
@@ -0,0 +1,98 @@
...
+where
+    Fut: Future<Output = T>,
+{
+    successor: F,
+    future: Option<Fut>,
+    next: T,
+    _marker: PhantomData<Fut>,
+}
+
+/// Creates a new stream where to produce each new element a clousre is called with the previous
```
stjepang:
```suggestion
/// Creates a new stream where to produce each new element a closure is called with the previous
```

src/stream/successor.rs
```diff
@@ -0,0 +1,98 @@
...
+/// assert_eq!(s.next().await, Some(25));
+/// #
+/// # }) }
+///
+/// ```
+pub fn successor<F, Fut, T>(start: T, func: F) -> Successor<F, Fut, T>
+where
+    F: FnMut(T) -> Fut,
+    Fut: Future<Output = T>,
+    T: Copy,
```
stjepang: In order to be an async version of `successors` from the standard library, we'd need to:

1. Remove the `T: Copy` bound.
2. Change the bound on `F` to `FnMut(&T) -> Fut`.
3. Change the output type of the future to `Option<T>`.
4. Change the type of the `start` argument to `Option<T>`.

See here for the type signature in the standard library:
https://doc.rust-lang.org/std/iter/fn.successors.html

felipesere: I am not sure how to drop the `T: Copy` bounce, as I need to pass a reference to it into the passed in function _and_ return it in `Poll::Ready(value)`.
Unless we go all the way to Rc?

felipesere: I've done 3 & 4, but I am not sure how to tackle 1 & 2.

yoshuawuyts: @felipesere a way to do this is by always staying one behind, and only returning the result of the previous computation once the result of the next computation is known. Until `None` is hit, at which point we flush. If you look closely, this is how the iter version works too: https://doc.rust-lang.org/src/core/iter/sources.rs.html#611. `item` is only returned *after* next has been computed.

src/stream/successors.rs
```diff
@@ -58,31 +63,19 @@ where
 /// # }) }
 ///
 /// ```
-pub fn successors<F, Fut, T>(start: Option<T>, func: F) -> Successors<F, Fut, T>
+pub fn successors<F, Fut, T>(first: Option<T>, succ: F) -> Successors<F, Fut, T>
```
yoshuawuyts: Can we perhaps mark this function as "unstable"?

src/stream/successors.rs
```diff
@@ -40,6 +44,7 @@ where
 /// });
 ///
 /// pin_utils::pin_mut!(s);
```
yoshuawuyts: Is this still needed?

felipesere: Nope :)

src/stream/successors.rs
felipesere: I think I did this now? I am not sure if I marked all the spots?

src/stream/successors.rs
```diff
@@ -0,0 +1,113 @@
...
+/// # }) }
+///
+/// ```
+#[cfg(feature = "unstable")]
+#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
+pub fn successors<F, Fut, T>(first: Option<T>, succ: F) -> Successors<F, Fut, T>
+where
+    F: FnMut(T) -> Fut,
+    Fut: Future<Output = Option<T>>,
+    T: Copy,
```
k-nasa: Do need a copy trait? It seems too restrictive to me.

felipesere: I managed to remove the trait bound. The FnMut was also wrong, as the one in std takes a &T and my version to a T. 
This should be up to date now.

src/stream/successors.rs
yoshuawuyts: can we perhaps remove this then? (:

yoshuawuyts:
Created a backup of the async impl here for future reference (e.g. when we get async closures): https://gist.github.com/yoshuawuyts/32fb44959c0bd66a683008ee1d540500. Thanks!

PR #364, opened 2019-10-17
Don't flush files if they weren't written to

stjepang:
Before this PR, files that were opened and then never written to would get flushed on drop, which is a useless operation and especially concerning when small files are read.

Newly opened files now start in flushed state. When constructing `async_std::fs::File` from a `std::fs::File`, we still have to start in "dirty" (not flushed) state because we don't know if the handle was previously written to.

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -66,6 +66,23 @@ pub struct File {
 }
 
 impl File {
+    /// Creates an async file handle.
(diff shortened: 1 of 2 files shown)


skade:
I agree with @yoshuawuyts there, I think an enum would be preferable. Still, as this is internal, I'd merge and patch.

PR #368, opened 2019-10-18
add stream::count

starsheriff:
ref #129


src/stream/stream/count.rs
```diff
@@ -0,0 +1,41 @@
...
+    pin_utils::unsafe_unpinned!(count: usize);
+
+    pub(crate) fn new(stream: S) -> Self {
+        CountFuture { stream, count: 0 }
+    }
+}
+
+impl<S> Future for CountFuture<S>
+where
+    S: Sized + Stream,
```
stjepang: I believe `S: Sized` is redundant here, i.e. `S` is assumed to be `Sized` by default unless we opt-out with `S: ?Sized`.

```suggestion
    S: Stream,
```

starsheriff: Actually I don't remember why I used it in the first place. I think the compiler was complaining about it at some point. I actually don't see anymore why it should be `Sized` in the first place.

Will remove it.

src/stream/stream/mod.rs
```diff
@@ -1392,6 +1394,33 @@ extension_trait! {
...
+
+            assert_eq!(s1.count().await, 1);
+            assert_eq!(s2.count().await, 3);
+            #
+            # }) }
+            ```
+        "#]
+        fn count(self) -> impl Future<Output = Ordering> [CountFuture<Self>]
+        where
+            Self: Sized + Stream,
```
k-nasa: Do you need this 'Stream'?

starsheriff: Actually not, your're right.

yoshuawuyts:
Filed https://github.com/async-rs/async-std/pull/401 to fix CI. Let's re-run it after we merge that. Thanks!

src/stream/stream/count.rs
```diff
@@ -0,0 +1,41 @@
...
+#[doc(hidden)]
+#[allow(missing_debug_implementations)]
+pub struct CountFuture<S> {
+    stream: S,
+    count: usize,
+}
+
+impl<S> CountFuture<S> {
+    pin_utils::unsafe_pinned!(stream: S);
+    pin_utils::unsafe_unpinned!(count: usize);
```
stjepang: Can we rebase this branch and use `pin_project!` here?

starsheriff: Yes, can do that. Sorry for the late reply, I was quite busy the last weeks.

starsheriff:
so, changed to using `pin_project`. Guess it is ready now.

starsheriff:
@stjepang, can you have a look?

src/stream/stream/mod.rs
```diff
@@ -1802,6 +1804,33 @@ extension_trait! {
...
+            let s1 = stream::from_iter(vec![0]);
+            let s2 = stream::from_iter(vec![1, 2, 3]);
+
+            assert_eq!(s1.count().await, 1);
+            assert_eq!(s2.count().await, 3);
+            #
+            # }) }
+            ```
+        "#]
+        fn count(self) -> impl Future<Output = Ordering> [CountFuture<Self>]
```
k-nasa: https://github.com/async-rs/async-std/pull/368/files#diff-14e289ac8d84db1191a59c7117c5a476R29
Here is usize.

Why did you choose Ordering here?

```suggestion
        fn count(self) -> impl Future<Output = usize> [CountFuture<Self>]
```

starsheriff: I tried to keep it as generic as possible first, but I guess there is no point in it.

Issue #217, opened 2019-09-19
[tracking] sync


yoshuawuyts:
This is a tracking issue for feature parity with [`std::sync`](https://doc.rust-lang.org/std/sync/index.html). Not everything needs to be ported, but this is a full overview of what's in `std::sync` and `async_std::sync`.

@stjepang could you take a look and cross off what's not needed? I think we have most of it except maybe `Barrier`, `Condvar`, and `Arc`. Thanks!

## Structs
- [ ] `sync::Arc`
- [ ] `sync::Condvar`
- [x] `sync::Mutex`
- [ ] `sync::BarrierWaitResult`
- [x] `sync::MutexGuard`
- [ ] `sync::Once`
- [x] `sync::PoisonError`  (not required)
- [x] `sync::RwLock`
- [x] `sync::RwLockReadGuard`
- [x] `sync::RwLockWriteGuard`
- [ ] `sync::WaitTimeoutResult`
- [ ] `sync::Weak`

## Enums
- [x] `sync::TryLockError` (not required)

tmccombs:
I could take a look at `Condvar` and maybe `Once`

nbdd0121:
I looked into this briefly when trying to improve `Mutex`. I think it should be fairly easy: for `Condvar`, we simply store a list of `Waker`s, and `notify_one` or `notify_all` will wake one or more of these wakers. Dropping the future removes its own wakers from the list. The important thing about sync `Condvar` is that it unlocks and sleep atomically, which requires OS assistance usually, but in async case we don't.

For the wait with timeout APIs, it's basically `async_std::future::timeout`.

nbdd0121:
I think it should be a regular notify. Condvar timeouts are never meant to be precise. If we prioritise timeout to a regular notify, then we may lose a notify_one call if we are notifying an user that is very close to timeout, which isn't really desirable.

I looked at your implementation and have a few worries:
* The implementation seems not robust against repeat polls (poll again before waker called, which is possible in cases like using `either!`)
* In the current implementation we can still lose a notify_one call if we receive notify_one after we decided to return timeout, but before the destructor remove the waker from the blocker list.

stjepang:
I wonder how useful `wait_timeout` is since futures have baked-in support for timeouts. For example, `TcpStream` in the standard library has method `read_timeout` but we don't have it in `async-std` because we have futures combinators for timeouts.

Maybe we can just omit `wait_timeout` and only implement `wait`?

[...]

PR #369, opened 2019-10-18
Implement async_std::sync::Condvar

tmccombs:
Part of #217

### DIFF
--- a/src/sync/condvar.rs
+++ b/src/sync/condvar.rs
@@ -361,14 +361,23 @@ impl Condvar {
 
 impl fmt::Debug for Condvar {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        //f.debug_struct("Condvar").finish()
         f.pad("Condvar { .. }")
     }
 }
 
+/// A future that waits for another task to notify the condition variable.
+///
+/// This is an internal future that `wait` and `wait_until` await on.
 struct AwaitNotify<'a, 'b, T> {
+    /// The condition variable that we are waiting on
     cond: &'a Condvar,
+    /// The lock used with `cond`.
+    /// This will be released the first time the future is polled,
+    /// after registering the context to be notified.
     guard: Option<MutexGuard<'b, T>>,
+    /// A key into the conditions variable's `WakerSet`.
+    /// This is set to the index of the `Waker` for the context each time
+    /// the future is polled and not completed.
     key: Option<usize>,
 }
 
--- a/src/sync/condvar.rs
+++ b/src/sync/condvar.rs
@@ -1,11 +1,10 @@
 use std::pin::Pin;
 use std::time::Duration;
 
-use futures_timer::Delay;
 use slab::Slab;
 
 use super::mutex::{guard_lock, MutexGuard};
-use crate::future::Future;
+use crate::future::{timeout, Future};
 use crate::task::{Context, Poll, Waker};
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
@@ -29,7 +28,7 @@ impl WaitTimeoutResult {
 /// # Examples
 ///
 /// ```
-/// # fn main() { async_std::task::block_on(async {
+/// # async_std::task::block_on(async {
 /// #
 /// use std::sync::Arc;
 ///
@@ -55,7 +54,7 @@ impl WaitTimeoutResult {
 ///     started = cvar.wait(started).await;
 /// }
 ///
-/// # }) }
+/// # })
 /// ```
 #[derive(Debug)]
 pub struct Condvar {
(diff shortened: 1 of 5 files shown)


nbdd0121:
> @nbdd0121 do you have any suggestion on how to resolve your concerns?

I think notifying another waker when a WaitFuture is dropped with its opt_waker set to None should be enough. This should handle the notify_one well, but it may cause extra tasks being notified if we
have the following case:
* Task A calls wait
* Task B calls notify_all
* Task C calls wait
* Task A cancels wait

Task C will ends up being waken up. However I don't think it's a huge problem, because std's Condvar mentions that:
> Note that this function is susceptible to spurious wakeups. Condition variables normally have a boolean predicate associated with them, and the predicate must always be checked each time this function returns to protect against spurious wakeups.

nbdd0121:
IMO we don't need `has_blocked`. Unlike Mutex which assumes most calls don't block, for Condvar I think we can assume there are tasks blocking when `notify_one` or `notify_all` is called.

roblabla:
I feel like the wait/wait_timeout are kinda big footguns. If I understand them properly, the returned future will always return Ready the second time they're polled, without necessarily being *actually* notified. This can happen quite easily by using e.g. the `join` adapter from futures-rs.

This is *technically* okay according to Condvars' spurious wakeup conditions, but that is not mentioned at all in the docs. Besides, it seems very trivial to cause a spurious wakeup, and will likely produce bad, hard to debug code. I kinda question the usefulness of those functions compared to `wait_until`.

tmccombs:
>  If I understand them properly, the returned future will always return Ready the second time they're polled

Kind of. They will return `Ready` if they are polled and the attached Mutex is available. I could potentially use an `Arc<AtomicBool>` to keep track of if it was _actually_ notified. But I think (though I don't have benchmarks) that in most cases that would hurt performance.

nbdd0121:
> Kind of. They will return `Ready` if they are polled and the attached Mutex is available. I could potentially use an `Arc<AtomicBool>` to keep track of if it was _actually_ notified. But I think (though I don't have benchmarks) that in most cases that would hurt performance.

Probably only need to check if `Option<Waker>` is turned to `None`. Otherwise it'll be spurious wake up.

tmccombs:
>  Probably only need to check if Option<Waker> is turned to None. Otherwise it'll be spurious wake up.

Yes, but that requires locking the mutex for the wakers when polled, which is doable, but could hurt performance if the condition variable has very much contention. And if `wait_until`/`wait_timeout_until` are used, then it isn't an issue.

nbdd0121:
You need to lock the mutex to remove the waker when cancelling/completing the notify call, or to insert the waker when beginning a new call. I don't see why it'll hurt performance compared to wait_until.

tmccombs:
Ok, I'll take a stab at implementing that. It looks like there is also now a `WakerSet` type that I should use instead of a Slab of Wakers.

tmccombs:
@nbdd0121 I think my latest push should address your concerns.

src/sync/waker_set.rs
```diff
@@ -80,6 +80,28 @@ impl WakerSet {
         }
     }
 
+    /// If the waker for this key is still waiting for a notification, then update
+    /// the waker for the entry, and return false. If the waker has been notified,
+    /// treat the entry as completed and return true.
+    #[cfg(feature = "unstable")]
+    pub fn remove_if_notified(&self, key: usize, cx: &Context<'_>) -> bool {
```
tmccombs: I can't think of a really good name for this.

nbdd0121:
WakerSet's notify_one does not wake up anything if some task is woken already. This is slightly different from Condvar's notify_one which should always wake up one task (in the case of two concurrent notify_one call for example).

tmccombs:
If i understand the code on master that is how `notify_any` behaves, but `notify_one` behaves as we want for condvar.

nbdd0121:
Sorry, you're right. That change is introduced in e9edadf which isn't part of a PR so I am not aware of.

src/sync/condvar.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// # }) }
+    /// ```
+    pub fn notify_all(&self) {
+        self.wakers.notify_all();
+    }
+}
+
+impl fmt::Debug for Condvar {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        //f.debug_struct("Condvar").finish()
```
dignifiedquire, line 364: what about this comment?

src/sync/condvar.rs
```diff
@@ -0,0 +1,408 @@
...
+        f.pad("Condvar { .. }")
+    }
+}
+
+struct AwaitNotify<'a, 'b, T> {
```
dignifiedquire, line 369: Could you add a description of the purpose of the struct and its fields, that will make it easiert to understand how this is structured.

src/sync/waker_set.rs
dignifiedquire: seems descriptive at least :)

tmccombs:
I rebased and addressed your comments. Should I squash as well?

dignifiedquire:
Thank you

Issue #362, opened 2019-10-17
async_std::sync::Mutex is too heavy


nbdd0121:
I've noticed that async-std uses `std::sync::Mutex` in quite a few places. It's well known that parking_lot offers much better performance, and more noticeably space saving, compared to the std mutex, thus enabling very fine-grained locks.

The current `async_std::sync::Mutex` and other few places of async_std seems to be using `std::sync::Mutex`, which has quite a few allocations and is very heavy-weight. Is there any existing discussion about performance and code bloat impact of switching over to parking_lot?

There is also a probability that we can use ideas similar to parking_lot to create a very cheap and light-weight (i.e. only a few bytes, no extra allocation) async Mutex that optimises for non-contended cases.

stjepang:
Perhaps we could reuse `Spinlock` from `crossbeam-channel`?

nbdd0121:
I did a few experiments, and I think I got an idea that can reduce `async_std::sync::Mutex` to the size of usize and makes it allocation-free when there isn't contention.

Here's the brief list of my plan. If you think it looks alright, I'll proceed to implementation and benchmark the performance differences.

Step 1: Replace Slab with a double linked list
* insertion/removal into/from the slab is replaced with linked list insertion/removal.
* replace the item in the slab is replaced with a direct write to the list node.
* benefits:
** slab does not shrink itself, while linked list is.
** as we don't need to know the length of waker list, only if it's empty, we reduce size requirement of `blocked` to 16 bytes (from 40 bytes for slab) plus Mutex overhead.

Step 2: Replace doubly linked list with single linked list
* disallow non-list-head to acquire the lock. Note that the current unlock code only wakes the first item in the iterator, so it won't deadlock. This will actually give the Mutex a bit of extra fairness.
* now we no longer need to remove arbitrary elements from the queue, so we can change doubly linked list to single linked list, reduce `blocked` (which will be the linked list head pointer) to 8 bytes plus Mutex overhead.

Step 3: Pack everything into an AtomicUsize
* we can use the lowest 2-3 bits of the linked list head pointer to replace the LOCKED, BLOCKED and the Mutex, so packing everything into the same AtomicUsize.

After all steps the size of `Mutex<()>` will reduce from 64 bytes to 8 bytes, free to construct and destruct.

stjepang:
Do we allocate nodes for the linked list on the heap? If so, I am worried that these allocations will be slower than using the slab.

[...]

PR #370, opened 2019-10-18
Improve performance and space of async_std::sync::Mutex

nbdd0121:
Related #362

### DIFF
--- a/src/sync/mutex.rs
+++ b/src/sync/mutex.rs
@@ -42,7 +42,6 @@ impl RawMutex {
         RawLockFuture {
             mutex: self,
             opt_key: None,
-            acquired: false,
         }
     }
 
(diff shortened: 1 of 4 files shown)


nbdd0121:
Some numbers

Before this PR:
test mutex_contention        ... bench:   1,866,362 ns/iter (+/- 162,639)
test mutex_mimick_contention ... bench:       1,448 ns/iter (+/- 131)
test mutex_no_contention     ... bench:     341,444 ns/iter (+/- 17,727)
test mutex_unused            ... bench:          27 ns/iter (+/- 1)
Size of `Mutex<()>` is 64 bytes

After this PR:
test mutex_contention        ... bench:   1,669,542 ns/iter (+/- 167,686)
test mutex_mimick_contention ... bench:       1,034 ns/iter (+/- 148)
test mutex_no_contention     ... bench:     258,535 ns/iter (+/- 38,157)
test mutex_unused            ... bench:           0 ns/iter (+/- 0)
Size of `Mutex<()>` is 16 bytes

The PR also makes sure `Mutex<T>` only uses 1 copy of its cold path across different `T`s, reducing bloat.

nbdd0121:
@stjepang This PR is ready. I scrapped the plan to squeeze everything into 1 usize (this PR makes it 2 usize), because squeezing further would hurt code clarity and the code would not be reusable for RwLock.

src/sync/mutex.rs
```diff
@@ -219,8 +304,9 @@ impl<T> Mutex<T> {
     /// #
     /// # })
     /// ```
+    #[inline]
```
skade: In general, forced inlining of trivial operations is considered an antipattern. Do we run into issues that these methods are not inlined?

nbdd0121: #[inline] is not forcing inline (different from #[inline(always)]; it is just hinting the compiler that this should probably be inlined.

Usually it's a good idea to have the #[inline] notation if the method is going to be inlined across the crate boundary. In the case Mutex is polymorphic over T, so we may get away with it, but a monomorphic function will never be inlined across crate boundary because the compiler cannot see the underlying implementation.

skade: Correct, I misspoke there a little.

That's correct as a hint, this is only in the absence of lto, though. Considering that libstd uses the same annotations, we should follow that practice.

dignifiedquire:
@nbdd0121 we now have a Mutex implementation based on a WakerList, is this still needed?

nbdd0121:
@dignifiedquire The currently implementation is still using Slab. It'll still be better to use a linked-list based implementation for predictable performance upper bound and avoid memory leak.

However I'm quite busy recently. Maybe I'll take a look again and rebase next month.

nbdd0121:
This would be no longer necessary after #822 lands, closing.

PR #371, opened 2019-10-19
Condense FromStream/Extend impls

stjepang:
Just a negative diff that collects `FromStream`/`Extend` impls into singular files.

I found it a bit difficult to navigate the codebase since there are [so many](https://github.com/async-rs/async-std/tree/ec23632f3e7d2d8f9ed49995b85741eeb0a5be69/src) directories. After this change, there are [fewer](https://github.com/stjepang/async-std/tree/squash-stream-impls/src) directories in the source tree.

I suppose originally we intended to re-export modules `collections`, `vec`, `string`, `unit`, and so on, but we won't do that since those modules don't host async types. Before this PR we also had re-exports of types like `Vec`, which were hidden and unused.

Also wrote a macro that simplifies a lot of identical `FromStream` impls.


yoshuawuyts:
> but we won't do that since those modules don't host async types.

I'm still planning to tho. Off the top of my head I can think of at least 4 more types per submodule we may want to export. Which if we follow through on that would mean restoring these submodules.

I'm open to the idea of using a macro to squash similar impls. But I don't think removing the submodules is something we should be doing at this time.

stjepang:
> I'm still planning to tho. Off the top of my head I can think of at least 4 more types per submodule we may want to export. Which if we follow through on that would mean restoring these submodules.

But why if we're not adding any functionality to the modules? Could you mention those 4 types because I can't guess what they would be?

I think we'd then also need to have the `array` module because right now we have this impl in `vec/from_stream.rs`, which doesn't use `Vec`:

```rust
impl<'b, T: Clone> FromStream<T> for Cow<'b, [T]>
```

> But I don't think removing the submodules is something we should be doing at this time.

So my motivation for removing the submodules was twofold. First, I was sure we'll never ever export those modules and make them public, so why keep them? :) Second, the codebase feels big and intimidating with so many modules (at least that's my feeling while skimming through it).

It's also not clear to me what modules belong or don't belong into the library by that logic. I'd personally like to avoid a situation where `async-std` becomes the whole `std` and contains almost all modules, even those that are not "very async". At least in the near future.

Looking at modules in the `src/` directory it's not obvious which modules are public and which are private. Granted, we have `macros.rs` and `utils.rs` but those are special cases.

I guess my perspective is that if `async-std` has module `path`, then users will expect in there to be `Path` different from the one in `std`. But that would not apply to modules like `vec` or `collections`. So the line between `std` and `async-std` is blurred and I worry we could end up in a situation where users are e.g. confused whether they should use `async_std::collections` or `std::collections`. That would be confusing to me at least.

Idk. I admit I'm feeling a bit lost in the codebase right now and not understanding what goes where. :( What would help is making the codebase smaller and constraining the module structure to only those modules from `std` that have been "asyncified".

I also feel in this situation the "burden of proof" is on expanding the library, not on keeping it smaller. So (in my opinion) this PR needs good motivation for *not* being merged rather than getting merged.

yoshuawuyts:
This PR is now quite out of sync, and would require some heavy rebasing to move forward. Also it seems this topic has somewhat settled down. I'm going to go ahead and close this so it's not kept open indefinitely, but that doesn't mean we can't revisit this in the future. Thanks!

Issue #22, opened 2019-08-14
Port std::process


skade:
Design document with active discussion:

https://paper.dropbox.com/doc/async-process--Ae7VXYrJ4sSucoYlMC7XYBQvAg-Fbg2Jq7UbhqihtnWpc1EY

External process execution is clearly asynchronous. This API should make it possible to both asynchronously handle IO streams and also just wait for them to finish their execution.

OddCoincidence:
I'd like to work on this if nobody else is already

skade:
@OddCoincidence I know of no one currently working. Please keep in mind many of us are on RustConf this week, so mentoring might be a little delayed :).

Feel free to hop on our discord or open a draft PR here!

yaahc:
just pinging the issue to let ppl know that I'm interested in working on this.

stjepang:
@OddCoincidence Hi, sorry about the slow response! Did you perhaps make any progress? 

@yaahc Let's see what's the current status first. If the issue is not taken yet, I can assign it to you.

As for the implementation plan, my suggestion is we kick off by building a minimum working module with:

1. Add `Command` struct with methods `new` and `spawn`.
2. Add `Child` struct with no fields and no methods.
3. Implement `Future` for `Child`.

Then, we can do the remaining work in follow-up PRs (in no particular order):

* Add the remaining methods on `Command`.
* Add the remaining methods on `Child`.
* Add `ChildStdin`, `ChildStdout`, and `ChildStderr`.

yaahc:
On it boss _salutes @stjepang_

stjepang:
@yaahc Hey, just wanted to check in and ask how it's going :) If you need help with this task or have any questions, let me know!

yaahc:
Hey @stjepang, I haven't been able to start yet because I've been swamped with procrastinating / slowly working on my talk for cogoldrust, ADHD problems >_>, I'm still interested in working on this but its an intimidatingly large task so I'm having trouble starting it. My hope is to start working on it after Thursday, which is when I'm giving the trial version of my talk @ the bay area rust meetup. After that I should be essentially done with my talk and back in "ready for other things" mode.

I'm sorry for claiming this task and taking so long to do it :S, if you're feeling pressure to start this soon lmk and that might help me get going easier or feel free to assign to someone else or w/e.

vhakulinen:
Has there been any progress with this?

[...]

PR #373, opened 2019-10-19
[WIP] adds skeleton for process Child and Command

Michael-J-Ward:


yaahc:
part of https://github.com/async-rs/async-std/issues/22

PR #377, opened 2019-10-22
Add `Stream::sum()` and `Stream::product()` implementations

ktomsic:
These are the stream equivalents to `std::iter::Iterator::sum()` and
`std::iter::Iterator::product()`.

Note that this changeset tweaks the `Stream::Sum` and `Stream::Product`
traits a little: rather than returning a generic future `F`, they return
a pinned, boxed, `Future` trait object now. This is in line with other
traits that return a future, e.g. `FromStream`.


src/option/product.rs
```diff
@@ -0,0 +1,66 @@
...
+    T: Product<U>,
+{
+    #[doc = r#"
+        Takes each element in the `Stream`: if it is a `None`, no further
+        elements are taken, and the `None` is returned. Should no `None` occur,
+        the product of all elements is returned.
+
+        # Examples
+
+        This sums up the position of the character 'a' in a vector of strings,
```
felipesere: I don't think the written example lines up with the code example anymore?

ktomsic: Whoops. Good catch!

PR #379, opened 2019-10-22
Add max_by

felipesere:
Theoretically its was possibly to create an equivalent of max_by by using min_by.
That would have been awkward to read for users though.

I first tried to Box the comparison function and simply passing it down from max_by to min_by with `reverse()` applied. That got too tedious and meant users were exposed to the Box.

This variation adds a new field to the strut - now renamed MinMaxBy - that tells us whether we are maximising or minimising.

### DIFF
--- a/src/stream/stream/min_by.rs
+++ b/src/stream/stream/min_by.rs
@@ -20,19 +20,28 @@ pin_project! {
 }
 
 #[derive(PartialEq, Eq)]
-pub(crate) enum Direction {
+enum Direction {
     Maximizing,
     Minimizing,
 }
 
 
 impl<S, F, T> MinMaxByFuture<S, F, T> {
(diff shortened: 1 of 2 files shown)


felipesere:
#129

yoshuawuyts:
Hey @felipesere! -- thanks so much for this PR!

I noticed #384 partially overlaps with this PR. I would've left some notes on this PR, but basically #384 was exactly what I was hoping for these methods, so I've gone ahead and merged that instead.

I still want to take a moment to thank you for the work you've done here! -- I've seen you've put in a few patches, and they're all really good! I'm going to go ahead and close this PR in favor of #384, but please don't take it as a discouragement from continuing to contribute. I'm really stoked you've been helping out, and would love to continue to have you here!

Thanks so much!

PR #380, opened 2019-10-22
Add channel behind unstable feature flag

### LINKED ISSUE #212: Design of async channels

stjepang:
This is a simple bounded MPMC channel based on `crossbeam-channel`.

We're only adding the following to `async_std::sync`:
* Function `bounded(cap)` that constructs a channel.
* Types `Sender` and `Receiver`.

Closes #212

### DIFF
--- a/src/sync/channel.rs
+++ b/src/sync/channel.rs
@@ -1,0 +1,1134 @@
+#![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
+
+use std::cell::UnsafeCell;
+use std::fmt;
+use std::future::Future;
+use std::isize;
+use std::marker::PhantomData;
+use std::mem;
+use std::ops::{Deref, DerefMut};
+use std::pin::Pin;
+use std::process;
+use std::ptr;
+use std::sync::atomic::{self, AtomicBool, AtomicUsize, Ordering};
+use std::sync::Arc;
+use std::task::{Context, Poll, Waker};
+
+use crossbeam_utils::{Backoff, CachePadded};
+use futures_core::stream::Stream;
+use slab::Slab;
+
+/// Creates a bounded multi-producer multi-consumer channel.
+///
+/// This channel has a buffer that can hold at most `cap` messages at a time.
+///
+/// Senders and receivers can be cloned. When all senders associated with a channel get dropped, it
+/// becomes closed. Receive operations on a closed and empty channel return `None` instead of
+/// blocking.
+///
+/// # Panics
+///
+/// If `cap` is zero, this function will panic.
+///
+/// # Examples
+///
+/// ```
+/// # async_std::task::block_on(async {
+/// #
+/// use std::time::Duration;
+///
+/// use async_std::sync::channel;
+/// use async_std::task;
+///
+/// let (s, r) = channel(1);
+///
+/// // This call returns immediately because there is enough space in the channel.
+/// s.send(1).await;
+///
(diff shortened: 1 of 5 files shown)


src/sync/channel.rs
```diff
@@ -0,0 +1,1134 @@
+#![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
```
taiki-e:
```suggestion
```
Seems we already apply these lints at lib.rs.
https://github.com/async-rs/async-std/blob/944e43d4bf1e228b9e9243c79a3a600d97855308/src/lib.rs#L45

stjepang:
@yoshuawuyts has approved this PR verbally so I'm going to merge now :)

PR #383, opened 2019-10-23
Add Stream::debounce()

dylanhart:


yoshuawuyts:
@dylanhart thanks so much for starting work on this! It's been a bit since the last activity on this PR. I'd love to see it land; is there anything we can help with to move this forward?

dylanhart:
Thanks for the interest. I just need spend a couple hours trying to come up  with a test harness for this. I haven't yet investigated if there are other tests that have a mock clock that I can use. Beyond that I need to audit the use of unsafe as this involves a couple of pins and mutable state. And then docs of course.

yoshuawuyts:
@dylanhart we've recently started adopting `pin-utils-lite` which removes a lot of the unsafe code in our stream adapters. I think it would be a good fit here too.

Currently we don't have any mock clocks available, so instead we tend to use real timeouts in the order of milliseconds to run the tests. We've found this to work reasonably well.

Also in terms of testing: more is always welcome, but having a single doctest that tests the method would be enough for us to land it — if you're short on time, don't worry about going the extra mile right away (:

PR #385, opened 2019-10-23
add stream::min_by_key method

yjhmelody:
#129


src/stream/stream/min_by_key.rs
```diff
@@ -0,0 +1,58 @@
...
+        match next {
+            Some(new) => {
+                let new = self.as_mut().key_by()(&new);
+                cx.waker().wake_by_ref();
+                match self.as_mut().min().take() {
+                    None => *self.as_mut().min() = Some(new),
+
+                    Some(old) => match new.cmp(&old) {
+                        Ordering::Less => *self.as_mut().min() = Some(new),
+                        _ => *self.as_mut().min() = Some(old),
```
k-nasa: Since `self.as_mut (). Min ()` is described many times, how about putting it in a variable once?

like this
```rust
let min = self.as_mut().min();

match min.take() {
    None => *min = Some(new),

    Some(old) => match new.cmp(&old) {
        Ordering::Less => *min = Some(new),
        _ => *min = Some(old),
    },
}
```

src/stream/stream/min_by_key.rs
```diff
@@ -0,0 +1,60 @@
...
+            stream,
+            min: None,
+            key_by,
+        }
+    }
+}
+
+impl<S, K> Future for MinByKeyFuture<S, S::Item, K>
+where
+    S: Stream + Unpin + Sized,
```
taiki-e:
```suggestion
    S: Stream,
```

src/stream/stream/min_by_key.rs
```diff
@@ -0,0 +1,60 @@
...
+    K: FnMut(&S::Item) -> S::Item,
+    S::Item: Ord + Copy,
```
taiki-e:
```suggestion
    S::Item: Ord,
```

src/stream/stream/min_by_key.rs
```diff
@@ -0,0 +1,60 @@
...
+                    None => *this.min = Some(new),
+
+                    Some(old) => match new.cmp(&old) {
+                        Ordering::Less => *this.min = Some(new),
+                        _ => *this.min = Some(old),
+                    },
+                }
+                Poll::Pending
+            }
+            None => Poll::Ready(*this.min),
```
taiki-e:
```suggestion
            None => Poll::Ready(this.min.take()),
```

src/stream/stream/mod.rs
```diff
@@ -600,6 +602,42 @@ extension_trait! {
...
+            # }) }
+            ```
+        "#]
+        fn min_by_key<K>(
+            self,
+            key_by: K,
+        ) -> impl Future<Output = Option<Self::Item>> [MinByKeyFuture<Self, Self::Item, K>]
+        where
+            Self: Sized,
+            K: FnMut(&Self::Item) -> Self::Item,
```
taiki-e: please add the following trait bounds:
```diff
+             Self::Item: Ord,
```

PR #388, opened 2019-10-24
Implement stream::cycle(..)

felipesere:
This is a rough initial step that is able to take a vec and cycle it indefinitly.

I saw there were comments on #34 about the API and using `IntoStream`.
Would I have to impl IntoStream for a `Vec<T>` myself to be able to have the example be liek below?

```
use stream::IntoStream;

let vals = vec![1,2,3];
let s = vals.into_stream().cycle();

assert_eq!(s.next().await, Some(1));
assert_eq!(s.next().await, Some(2));
assert_eq!(s.next().await, Some(3));
assert_eq!(s.next().await, Some(1));
assert_eq!(s.next().await, Some(2));
```

I'm happy to get some pointers and see this through :) 

Ref: #129


src/stream/cycle.rs
```diff
@@ -0,0 +1,60 @@
...
+///
+/// assert_eq!(s.next().await, Some(1));
+/// assert_eq!(s.next().await, Some(2));
+/// assert_eq!(s.next().await, Some(3));
+/// assert_eq!(s.next().await, Some(1));
+/// assert_eq!(s.next().await, Some(2));
+/// #
+/// # })
+/// ```
+pub fn cycle<T: Copy>(source: Vec<T>) -> impl Stream<Item = T> {
```
k-nasa: I thought Clone was better here, but what about it?

k-nasa: Shouldn't it be a stream instead of a vec?

felipesere: Oh I thought I had taken Copy from somewhere else.
Good shout though.
I have changed the bound to `Clone` now and added calls to it where the compiler complained 😸

felipesere:
@yoshuawuyts The bounds are adjusted and we take as Stream as argument.

What API changes do you have in mind? Should I attach it to stream itself so once can write

```
let s = stream::once(7).cycle();
```

src/stream/cycle.rs
```diff
@@ -40,21 +71,21 @@ impl<T: Copy> Stream for Cycle<T> {
...
+/// assert_eq!(s.next().await, Some(7));
+/// assert_eq!(s.next().await, Some(7));
+/// assert_eq!(s.next().await, Some(7));
+/// assert_eq!(s.next().await, Some(7));
 /// #
 /// # })
 /// ```
-pub fn cycle<T: Copy>(source: Vec<T>) -> impl Stream<Item = T> {
-    let len = source.len();
+pub fn cycle<S: Stream<Item = T>, T: Clone>(source: S) -> impl Stream<Item = S::Item> {
```
yoshuawuyts: This should probably be a method on `Stream`, and not a free function. Similar to [`Iterator::cycle`](https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.cycle)

yoshuawuyts:
@felipesere yes, that's exactly what I was hoping you'd say! -- this should indeed be a method on `Stream`. Very excited for progress on this; we're getting real close to having all methods on `Stream` implemented!

felipesere:
@yoshuawuyts I've moved it, but I think clippy is unhappy on CI and not in a mood to run? 😃

src/stream/cycle.rs
felipesere: Done 🎉

src/stream/stream/cycle.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+#[derive(Eq, PartialEq)]
+enum CycleState {
+    FromStream,
+    FromBuffer,
+}
+
+impl<T: Clone, S: Stream<Item = T>> Cycle<S, T> {
```
k-nasa: I think this code is better.  I don't feel the need to derive a generic type T.
Do you think?

```rust
impl<S> Cycle<S, S::Item>
where
    S: Stream,
    S::Item: Clone,
{
```

felipesere: Is it also possible to use the same trick for the struct definition?

src/stream/stream/mod.rs
```diff
@@ -369,6 +371,38 @@ extension_trait! {
             Chain::new(self, other)
         }
 
+        #[doc = r#"
+            Creats a stream that yields the provided values infinitely and in order.
```
```suggestion
            Creates a stream that yields the provided values infinitely and in order.
```

src/stream/stream/cycle.rs
```diff
@@ -0,0 +1,73 @@
...
+            state: CycleState::FromStream,
+        }
+    }
+}
+
+impl<S, T> Stream for Cycle<S, T>
+where
+    S: Stream<Item = T>,
+    T: Clone,
+{
```
k-nasa: I don't feel the need to derive a generic type T.
It ’s the same as before.

felipesere:
@k-nasa @yoshuawuyts anything else to add? 😅

felipesere:
@yoshuawuyts anything I should still do on this? :)

PR #384, opened 2019-10-23
add stream::max_by method

yjhmelody:
I impl it by reference `min_by`. 
Though altering `x.cmp(y)` to `y.cmp(x)` can also do the same thing.

#129

PR #392, opened 2019-10-26
Remove usage of actions-rs/clippy-check

taiki-e:
#286, #384, and #385's CI is failing because of this error. I think this action is broken.

https://github.com/async-rs/async-std/pull/286/checks?check_run_id=276113673:

```text
##[error]Unable to create clippy annotations! Reason: HttpError: Resource not accessible by integration
##[warning]It seems that this Action is executed from the forked repository.
##[warning]GitHub Actions are not allowed to create Check annotations, when executed for a forked repos. See https://github.com/actions-rs/clippy-check/issues/2 for details.
Posting clippy checks here instead.
```

Related: https://github.com/actions-rs/clippy-check/issues/2#issuecomment-538655240

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,12 +7,13 @@ on:
       - staging
       - trying
 
+env:
+  RUSTFLAGS: -Dwarnings
+
 jobs:
   build_and_test:
     name: Build and test
     runs-on: ${{ matrix.os }}
-    env:
-      RUSTFLAGS: -Dwarnings
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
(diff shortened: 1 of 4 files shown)


taiki-e:
Nope, CI fails are due to clippy bug...
```text
error: do not call a function with mutable arguments inside of `debug_assert!`
  --> src/io/buf_read/read_line.rs:40:27
   |
40 |             debug_assert!(buf.is_empty());
   |                           ^^^^^^^^^^^^^^
   |
   = note: `#[deny(clippy::debug_assert_with_mut_call)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#debug_assert_with_mut_call
```

stjepang:
Let's just delete the whole Clippy check? I think Clippy is useful during development but is not really suitable for continuous integration :)

taiki-e:
> Let's just delete the whole Clippy check? I think Clippy is useful during development but is not really suitable for continuous integration :)

I think we should use the beta's clippy. nightly's clippy often adds broken lint, so I think it's not really suitable for CI.

skade:
Let's try out Beta for a while and see if it improves things.

PR #395, opened 2019-10-27
add stream mod docs

yoshuawuyts:
Add mod-level docs for `stream`. Thanks!

## Screenshot

Screenshot_2019-10-28 async_std stream - Rust

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -156,7 +156,7 @@
 //!
 //! # while let Loops and IntoStream
 //!
-//! Rust's `while let` loop syntax is actually sugar for streams. Here's a basic
+//! Rust's `while let` loop syntax is an idiomatic way to iterate over streams. Here's a basic
 //! example of `while let`:
 //!
 //! ```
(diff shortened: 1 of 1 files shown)


src/stream/mod.rs
```diff
@@ -2,24 +2,303 @@
...
-//! let mut s = stream::repeat(9).take(3);
+//! Calling `next().await` this way gets repetitive. Rust has a construct which
+//! can call `next()` on your stream, until it reaches `None`. Let's go over
+//! that next.
 //!
-//! while let Some(v) = s.next().await {
-//!     assert_eq!(v, 9);
+//! # while let Loops and IntoStream
+//!
+//! Rust's `while let` loop syntax is actually sugar for streams. Here's a basic
```
skade:
```suggestion
//! Rust's `while let` loop syntax is an idiomatic way to iterate over streams. Here's a basic
```

The while let syntax is not really sugar for streams, it was always there.

src/stream/mod.rs
```diff
@@ -2,24 +2,303 @@
...
+//! support yet. This means that automatic conversions like with `for` loops
+//! doesn't occur yet, and `into_stream` will always have to be called manually.
+//!
+//! [`IntoStream`]: trait.IntoStream.html
+//! [`into_stream`]: trait.IntoStream.html#tymethod.into_stream
+//!
+//! # Adapters
+//!
+//! Functions which take an [`Stream`] and return another [`Stream`] are
+//! often called 'stream adapters', as they're a form of the 'adapter
```
skade:
```suggestion
//! often called 'stream adapters', as they are a form of the 'adapter
```

PR #396, opened 2019-10-27
standardize net docs

yoshuawuyts:
Port over the [`std::net`](https://doc.rust-lang.org/std/net/index.html) docs to `async_std::net`. Thanks!

## Screenshots

Screenshot_2019-10-28 async_std net - Rust

### DIFF
--- a/src/net/mod.rs
+++ b/src/net/mod.rs
@@ -1,14 +1,42 @@
 //! Networking primitives for TCP/UDP communication.
 //!
-//! For OS-specific networking primitives like Unix domain sockets, refer to the [`async_std::os`]
-//! module.
+//! This module provides networking functionality for the Transmission Control and User
+//! Datagram Protocols, as well as types for IP and socket addresses.
 //!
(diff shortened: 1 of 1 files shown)


src/net/mod.rs
```diff
@@ -1,14 +1,42 @@
 //! Networking primitives for TCP/UDP communication.
 //!
-//! For OS-specific networking primitives like Unix domain sockets, refer to the [`async_std::os`]
-//! module.
+//! This module provides networking functionality for the Transmission Control and User
+//! Datagram Protocols, as well as types for IP and socket addresses.
```
skade:
```suggestion
//! Datagram Protocols (TCP and UDP), as well as types for IP and socket addresses.
```

yoshuawuyts: Afaik TCP is not commonly described as a "datagram protocol" (it does use packets, but not _datagrams_). I think it's okay to lean on the wording the stdlib provides.

PR #397, opened 2019-10-27
add mod level docs for sync

yoshuawuyts:
Port the mod-level docs from std for `sync`. There's a bunch about memory fences etc in there that I wasn't sure if we wanted to adapt, but figured I'd just leave it in. At least it now resembles std, and we can always circle back on the wording later. Thanks!

## Screenshots

Screenshot_2019-10-28 async_std sync - Rust

### DIFF
--- a/src/sync/mod.rs
+++ b/src/sync/mod.rs
@@ -6,6 +6,9 @@
 //!
 //! ## The need for synchronization
 //!
+//! async-std's sync primitives are scheduler-aware, making it possible to
+//! `.await` their operations - for example the locking of a [`Mutex`].
+//!
 //! Conceptually, a Rust program is a series of operations which will
 //! be executed on a computer. The timeline of events happening in the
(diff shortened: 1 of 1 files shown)


src/sync/mod.rs
```diff
@@ -4,6 +4,149 @@
 //!
 //! [`std::sync`]: https://doc.rust-lang.org/std/sync/index.html
 //!
+//! ## The need for synchronization
```
skade: I think it's worth adding a sentence right on top on _why_ `async-std` has their own structures (scheduler-awareness). It's a bit hidden below.

skade: "async-std's sync primitives are scheduler-aware, making it possible to `.await` their operations - for example the locking of a Mutex"

yoshuawuyts: @skade that's an excellent suggestion. Adding it!

PR #401, opened 2019-10-28
fix doc recursion limit

yoshuawuyts:
Fix doc recursion limit on master. Building docs currently fails, and this fixes that. Thanks!


yoshuawuyts:
@k-nasa *nods* -- I left some review comments on that earlier though, and it seems there's now merge conflicts. I didn't want to assume you'd be able to resolve this straight away, so I figured I'd fix CI ahead of time. Thanks for that though!

PR #405, opened 2019-10-28
Replace select!/try_select! with Future::{race,try_race}

### LINKED ISSUE #387: Can not use async block or async fn in `async_std::future::select!`.
I would expect this to work:

```rust
async fn foo() {
}

async_std::future::select!(foo(), async { });
```

yoshuawuyts:
This changes the `select` and `try_select` macros into methods on `Future`. Resolves #387. Thanks!

## Motivation

The output of `join` and `try_join` is polymorphic (`n`-sized tuple) which requires a macro. `select` and `try_select` don't require polymorphism which makes more sense for it to be a method.

## Docs issues

~~Currently the only problem with this patch is that, uhh, the docs no longer build. There's something in the macro that's not working, and I'm not sure what. cc/ @stjepang maybe you have a clue?~~

## Future work

I think we should consider adding `join` and `try_join` as methods too, in addition to the existing macros. That way there's polymorphic + non-polymorphic versions of both available, and the APIs become symmetrical again for the non-polymorphic case. Besides, it seems likely that joining 2 futures is the most common case, which can be a nice shorthand.

## Screenshot

Screenshot_2019-10-28 async_std future - Rust


yoshuawuyts:
okay yay, got the docs to compile! -- needed to set the params to `std::future::Future`. While not great, at least it works! :grin:

src/future/future/mod.rs
```diff
@@ -129,6 +133,90 @@ extension_trait! {
...
         }
+
+        #[doc = r#"
+            Waits for one of two similarly-typed futures to complete.
+
+            Awaits multiple futures simultaneously, returning all results once complete.
+
+            This function will return a new future which awaits for either one of both
+            futures to complete. If multiple futures are completed at the same time,
+            resolution will occur in the order that they have been passed.
```
stjepang: Do we want to have this guarantee or make it unspecified which result is returned if both features are ready at the same time? My hunch is that "random" could be a better default.

src/future/future/mod.rs
```diff
@@ -129,6 +133,90 @@ extension_trait! {
         {
             DelayFuture::new(self, dur)
         }
+
+        #[doc = r#"
+            Waits for one of two similarly-typed futures to complete.
+
+            Awaits multiple futures simultaneously, returning all results once complete.
```
stjepang: Is it "all results once complete" or "the result that completes first"?

src/future/future/select.rs
```diff
@@ -0,0 +1,53 @@
...
+use pin_project_lite::pin_project;
+use async_macros::MaybeDone;
+
+use std::future::Future;
+use crate::task::{Context, Poll};
+
+pin_project! {
+    #[allow(missing_docs)]
+    #[allow(missing_debug_implementations)]
+    pub struct Select<L, R> where L: Future, R: Future<Output = L::Output> {
```
stjepang: Are the bounds necessary here? I believe having them only on `impl` blocks should be enough.

src/utils.rs
```diff
@@ -144,7 +144,7 @@ macro_rules! extension_trait {
     };
 
     // Parse the return type in an extension method.
-    (@doc ($($head:tt)*) -> impl Future<Output = $out:ty> [$f:ty] $($tail:tt)*) => {
+    (@doc ($($head:tt)*) -> impl Future<Output = $out:ty> $(+ $lt:lifetime)? [$f:ty] $($tail:tt)*) => {
```
stjepang: Ooops, I don't know how this slipped through. Thanks for fixing!

yoshuawuyts:
> By that logic, select should indeed be a trait method, although probably it wouldn't hurt to also have it as a standalone function. But I'm not insisting, just thinking out loud. Not sure what to make of this :)

I think that's quite reasonable! Especially for `join` / `try_join`. But since they're macros, I wonder if `select / try_select` should also be macros? This is a bit tricky because there's not really any precedent in the stdlib for this either. Good questions tho; not sure either!

src/future/future/mod.rs
yoshuawuyts: Agreed; let's leave it unspecified for now, and actually randomize it later.

src/future/future/mod.rs
yoshuawuyts: Good catch; it's the latter

src/future/future/select.rs
yoshuawuyts: Just verified this, and yes they're needed. I think that's okay though; don't think we need to spend any time chasing this down (:

yoshuawuyts:
Implemented all feedback!

yoshuawuyts:
Tests are currently broken, filed https://github.com/async-rs/async-std/pull/431 to fix it.

PR #416, opened 2019-10-30
Use once_cell instead of lazy_static

### LINKED ISSUE #406: use once_cell instead of lazy_static
`once_cell` by @matklad provides a neat way of initializing lazy singletons. There's a related RFC out now to add this to the stdlib.

In anticipation of this, and in order to help test it, it might be nice if we switched our code over to use `once_cell` from `lazy-static`.

## references

- RFC: https://github.com/matklad/rfcs/blob/std-lazy/text/0000-standard-lazy-types.md
- crate: https://crates.io/crates/once_cell

wusyong:
`once_cell` provides a neat way of initializing lazy singletons without
macro. This PR use `sync::Lazy` to streamline same pattern proposed in
related rust RFC.

Resolve #406


src/task/pool.rs
```diff
@@ -111,28 +111,26 @@ impl Pool {
 
 #[inline]
 pub(crate) fn get() -> &'static Pool {
-    lazy_static! {
-        static ref POOL: Pool = {
-            let num_threads = num_cpus::get().max(1);
-            let mut stealers = Vec::new();
+    static POOL: Lazy<Pool> = Lazy::new(|| {
```
matklad: As this doesn't really make heavy use of `Deref` for Lazy, a more minimal and direct approach would be to use OnceCell.

```rust
static POOL: OnceCell<Pool> = OnceCell::new();
POOL.get_or_init(|| ...)
```

But, as we are using `Lazy` elsewhere, I guess it's better to use it here as well, just for consistency

wusyong: Yeah, I make use of `Lazy` mostly to streamline with on-going rust RFC. I think this will be easier for later maintenance.

matklad: The RFC proposes both `OnceCell` and `Lazy`, as `OnceCell` is strictly more powerful.

PR #418, opened 2019-10-30
Small example for a TCP server that both handles IP v4 and v6

skade:
This works quite nicely through Stream#merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: check
-        args: --all --bins --examples
+        args: --all --bins --tests
 
     - name: check unstable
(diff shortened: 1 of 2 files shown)


stjepang:
I like this!

We should change CI to use `--features unstable` because this PR is currently failing due to the use of unstable features in examples.

skade:
We currently have a `check` and a `check unstable` step, but that doesn't properly cover examples. :/

stjepang:
@skade Can you edit `.github/workflows/ci.yml` and change this:

```yml
    - name: check
      uses: actions-rs/cargo@v1
      with:
        command: check
        args: --all --bins --examples
```

to this:

```yml
    - name: check
      uses: actions-rs/cargo@v1
      with:
        command: check
        args: --all
```

PR #419, opened 2019-10-30
Modernize a-chat

skade:
This has to be considered a draft.

The 2 major changes are:
1) Move the client to use 2 concurrent tasks and selecting on them for any errors. The tasks are sufficiently diverse to merit 2 tasks.
2) Move the 2 select statements to unified streams and having the broker and the client handle a set of "Events" themselves. This moves their logic down to a nice match.

The stream construction is still a bit ugly, though, especially in the client handling code of the server.

### DIFF
--- a/examples/a-chat/server.rs
+++ b/examples/a-chat/server.rs
@@ -3,18 +3,16 @@ use std::{
     sync::Arc,
 };
 
-use futures::{channel::mpsc, select, FutureExt, SinkExt};
-
 use async_std::{
     io::BufReader,
     net::{TcpListener, TcpStream, ToSocketAddrs},
     prelude::*,
     task,
+    sync::{channel, Sender, Receiver},
+    stream,
 };
 
 type Result<T> = std::result::Result<T, Box<dyn std::error::Error + Send + Sync>>;
-type Sender<T> = mpsc::UnboundedSender<T>;
-type Receiver<T> = mpsc::UnboundedReceiver<T>;
 
 #[derive(Debug)]
 enum Void {}
@@ -26,7 +24,7 @@ pub(crate) fn main() -> Result<()> {
 async fn accept_loop(addr: impl ToSocketAddrs) -> Result<()> {
     let listener = TcpListener::bind(addr).await?;
 
-    let (broker_sender, broker_receiver) = mpsc::unbounded();
+    let (broker_sender, broker_receiver) = channel(10);
     let broker = task::spawn(broker_loop(broker_receiver));
     let mut incoming = listener.incoming();
     while let Some(stream) = incoming.next().await {
@@ -39,7 +37,7 @@ async fn accept_loop(addr: impl ToSocketAddrs) -> Result<()> {
     Ok(())
 }
 
-async fn connection_loop(mut broker: Sender<Event>, stream: TcpStream) -> Result<()> {
+async fn connection_loop(broker: Sender<Event>, stream: TcpStream) -> Result<()> {
     let stream = Arc::new(stream);
     let reader = BufReader::new(&*stream);
     let mut lines = reader.lines();
@@ -48,15 +46,14 @@ async fn connection_loop(mut broker: Sender<Event>, stream: TcpStream) -> Result
         None => return Err("peer disconnected immediately".into()),
         Some(line) => line?,
     };
-    let (_shutdown_sender, shutdown_receiver) = mpsc::unbounded::<Void>();
+    let (_shutdown_sender, shutdown_receiver) = channel::<Void>(0);
     broker
         .send(Event::NewPeer {
             name: name.clone(),
             stream: Arc::clone(&stream),
             shutdown: shutdown_receiver,
         })
-        .await
-        .unwrap();
+        .await;
 
     while let Some(line) = lines.next().await {
         let line = line?;
@@ -76,28 +73,36 @@ async fn connection_loop(mut broker: Sender<Event>, stream: TcpStream) -> Result
                 to: dest,
                 msg,
             })
-            .await
-            .unwrap();
+            .await;
     }
 
     Ok(())
 }
 
+#[derive(Debug)]
+enum ConnectionWriterEvent {
+    Message(String),
+    Shutdown
+}
+
 async fn connection_writer_loop(
     messages: &mut Receiver<String>,
     stream: Arc<TcpStream>,
-    mut shutdown: Receiver<Void>,
+    shutdown: Receiver<Void>,
 ) -> Result<()> {
     let mut stream = &*stream;
-    loop {
-        select! {
-            msg = messages.next().fuse() => match msg {
-                Some(msg) => stream.write_all(msg.as_bytes()).await?,
-                None => break,
-            },
-            void = shutdown.next().fuse() => match void {
-                Some(void) => match void {},
-                None => break,
+    let messages = messages.map(ConnectionWriterEvent::Message);
+    let shutdown = shutdown.map(|_| ConnectionWriterEvent::Shutdown).chain(stream::once(ConnectionWriterEvent::Shutdown));
+
+    let mut events = shutdown.merge(messages);
+
+    while let Some(event) = events.next().await {
+        match event {
+            ConnectionWriterEvent::Message(msg) => {
+                stream.write_all(msg.as_bytes()).await?;
+            }
+            ConnectionWriterEvent::Shutdown => {
+                break
             }
         }
     }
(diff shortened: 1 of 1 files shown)


examples/a-chat/server.rs
```diff
@@ -76,28 +73,36 @@ async fn connection_loop(mut broker: Sender<Event>, stream: TcpStream) -> Result
...
-        select! {
-            msg = messages.next().fuse() => match msg {
-                Some(msg) => stream.write_all(msg.as_bytes()).await?,
-                None => break,
-            },
-            void = shutdown.next().fuse() => match void {
-                Some(void) => match void {},
-                None => break,
+    let messages = messages.map(ConnectionWriterEvent::Message);
+    let shutdown = shutdown.map(|_| ConnectionWriterEvent::Shutdown).chain(stream::once(ConnectionWriterEvent::Shutdown));
```
matklad: That's smart! Although I wonder if there's some simpler cooperative shutdown idiom...

matklad: Ok, I think I've conjured something up:

```rust
fn with_shutdown<S, T>(stream: S, shutdown: Receiver<Void>) -> impl Stream<Item=T>
where
    S: Stream<Item=T>,
    T: Unpin,
{
    let items = stream.map(Some);
    let shutdown = shutdown.map(|void| match void {}).chain(stream::once(None));
    items.merge(shutdown).scan((), |&mut (), item| item)
}
```

The loop can then be written as

```rust
let mut messages = with_shutdown(messages, shutdown);
while let Some(msg) = messages.next().await {

}
```

Note that this also allows pushing cancellation completely out of this function and onto the call site (at the cost of making this generic over `messages` stream)!

matklad: I think `with_shutdown` should probably take `shutdown: impl Future<Item = !>` rather than a specific receiver

examples/a-chat/server.rs
```diff
@@ -26,7 +24,7 @@ pub(crate) fn main() -> Result<()> {
 async fn accept_loop(addr: impl ToSocketAddrs) -> Result<()> {
     let listener = TcpListener::bind(addr).await?;
 
-    let (broker_sender, broker_receiver) = mpsc::unbounded();
+    let (broker_sender, broker_receiver) = channel(10);
```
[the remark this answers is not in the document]
skade: Maybe @stjepang can answer that?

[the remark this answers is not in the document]
stjepang: I think you should not think about buffer capacities too much. Capacities of 1, 10, and 100 are all valid here, it doesn't really matter.

Ideally, we'd use capacity of 0 as a safe default choice (like Go does, i.e. `make(chan int)` constructs a 0-capacity channel). However, the problem is async channels based on futures can't really have capacity 0.

A channel can truly have capacity of 0 only in the context of preemptible threads. With Go channels and `crossbeam-channel`, send and receive operation need to pair up, at which point we flip an atomic value and both sending and receiving side agree that a message has been sent.

With futures-based channel, we can't do that. Imagine a receive operation is pending and registered in the channel. Then comes a send operation and sends a message. What happens if the receiving side then wakes up and cancels its receive operation? Did the message get through or not? Doesn't matter if your answer is "yes" or "no", we'll run into some inconsistencies either way. I guess the point is that we can only create an illusion of 0-capacity channels, but the channel will in some ways behave as if the capacity was 1.

examples/a-chat/server.rs
yoshuawuyts: @matklad I feel like the `with_shutdown` method you've written here might be a bit daunting for people trying to understand everything that's going on. I don't quite understand what `shutdown.map(|void| match void {})` does, though I do trust it's _correct_.

I feel like in the case of this example we should probably try to err on keeping code as simple as possible. Never types, and empty matches could be tricky for people to pick up on. Personally I'd prefer to keep what is currently in the PR.

examples/a-chat/client.rs
```diff
@@ -1,45 +1,43 @@
...
+             println!("{}", message);
+         }
+         Ok(())
+    });
+
+    let outgoing = task::spawn(async move {
+         let mut stdin = BufReader::new(stdin()).lines();
+
+         while let Some(line) = stdin.next().await {
+             let line = line?;
```
yoshuawuyts: I think we can collapse some of these lines. Wrote a quick gist of some change I think we could do for the client: https://gist.github.com/yoshuawuyts/0e0b259882a280316fe21e6928590d08.

examples/a-chat/server.rs
stjepang: Same! I see how the "void" pattern is useful but is definitely something that is pretty unusal and gives me pause :)

matklad: 100% agree with a reasoning that this bit of code is too smart for the tutorial (or even for application code). At the same time, I feel strongly that:

* termination via explicit message is a huge code-smell in Rust, and in large projects it leads to problems with complexity and shutdown correctness. The problem with `Event::Shutdown` is that it creates two termination code paths when handing shutdown (you might receive the message, or you might just fall off the loop) as well as signaling shutdown (you either send a shutdown message directly, or unwind / early return via `?`). 
* because the pattern is not obvious, it's important that we teach it early

I'll try too see if we can implement both correct and simple shutdown. Some ideas I have for simplifying this:

* remove the `Void` enum: it was not introduced in this PR, and `.map(|_| ConnectionWriterEvent::Shutdown)` , line gave *me* a pause. In fact, `let shutdown = shutdown.map(|_| ConnectionWriterEvent::Shutdown).chain(stream::once(ConnectionWriterEvent::Shutdown));` is exactly the same as `let shutdown = shutdown.map(|void| match void {}).chain(stream::once(None));`, so I'd argue there's no increase in complexity here, it's too complex both before and after. What *adds* complexity is the `scan` call. 
* replace shutdown channel with `impl Future<Output = ()>`
* try to push cancellation to the call site

Finally, I believe the signature

```rust
fn with_shutdown<S, T, F>(stream: S, shutdown: F) -> impl Stream<Item=T>
where
    S: Stream<Item=T>,
    F: Future<Output = ()>
{
```

is right: it's not trivially simple, but it's easy to explain what this function does (terminates stream early based on the shutdown signal), and we want to re-use it twice even in this small file. I hope I can implement it in simpler terms.

See also https://github.com/rust-lang/book/issues/1910

matklad: Note that I am not entirely sure that my claim "message-based cancellation is horrible" is correct. I believe this because of the two code paths vs crash only software theoretic argument, and because I've seen *once* how explicit cancellation breeds complexity. See this series of issues/commits in the language client for kakune editor:

* https://github.com/ul/kak-lsp/pull/106
* https://github.com/ul/kak-lsp/commit/d7ec12a2ec688d9a3652ab9e01ad5074dd656ac9
* https://github.com/ul/kak-lsp/issues/108
* https://github.com/ul/kak-lsp/pull/109

examples/a-chat/client.rs
skade: I'm not a huge fan of using `?` in calls, tbh.

examples/a-chat/server.rs
stjepang: We really do need a robust way of canceling stuff that is not based on channels + select...

matklad: I think I've managed to make cancellation in this case rather easy, at the cost of building a 75-lines long stop_source/stop_token library: https://github.com/matklad/async-std/commit/ab02901dcdc0c6cdb289d78914231de8a132ec9c. I feel this is roughly the place where we want to end-up eventually, but I am not sure we want to do this for tutorial.

I am genuinely don't know what to do with tutorial, I have how there's no simple solution :(

yoshuawuyts: Met up with @skade in person today, and we had some discussions about this. I think we ended up with two conclusions:

- We probably need a counterpart to `Stream::merge` that interleaves 2 streams, but drops them once one of them is exhausted.
- We should look into writing a cancellation library that works for both futures and streams. We bounced around some ideas that were based on the pattern used in this PR. There's probably some design space we could explore here.

Regarding this PR. I think focusing on making it as understandable as possible might be the right direction for now. And once we figure out a more formal strategy for cancellation we can apply that in a follow-up PR.

How does that sound?

matklad: SGTM, I’ve sort-of reaches similar conclusions, specifically:

* merge shortest might be an interesting primitive, thought, in this particular case, it might be enough if merge *drops* a substrram that is exhausted (see https://github.com/rust-lang/rust/issues/66031 for minimizes example)

* cancellation should be in a library, and, as design space is large, it should be outside of async std at least for start. I believe last 75 lines of https://github.com/matklad/async-std/commit/ab02901dcdc0c6cdb289d78914231de8a132ec9c might be a good start actually, should I just publish that to crates.io?

* for this PR, I wonder if we should rip out cancellation completely?

yoshuawuyts: > for this PR, I wonder if we should rip out cancellation completely?

That's definitely an option I think. It's a very cool system to show off though, so I feel like it's a bit of a loss. But given the circumstance maybe it's for the best. 

Also yeah having it on crates.io would be neat; big fan of publishing things!

matklad: @yoshuawuyts published https://github.com/async-rs/stop-token. You are also an owner on crates-io (I was surprised that there's no async-rs-publishes group to which I can grant permissions). There's a high change that I won't be able to properly maintain this crate, but docs are very upfront that it is an experiment, and it's also in a finished state already, so there's hopefully little to main (well, until folks ask for linked tokens and cancellation callbacks...)

PR #420, opened 2019-10-30
correction of a typo

JayatiGoyal:
times was converted to time


stjepang:
Hmmm, I believe simply "... compilation." could work (: but I'll leave this to @yoshuawuyts

yoshuawuyts:
Either or works for me. Going by the lowest friction change, which is to just merge this PR :~)

PR #421, opened 2019-10-30
Refactor the task module

stjepang:
Okay, this is a big PR. Sorry 😅

Nothing is fundamentally different from the previous implementation - I mostly reorganized the modules so that it becomes clearer what is where and what is what. While this doesn't really change our task executor, some small tweaks and the nature of cleaning things up made it a bit more efficient.

My primary motive here was to clearly separate what is the public API (what is visible in the generated docs for `async_std::task`) and what is the task executor, i.e. the "machine" driving the module.

The executor now lives in the `crate::task::driver` module, and everything else is an implementation of the high-level API that really has nothing to do with the executor. For example, `task_local!` and `block_on()` have meaty implementations, but are not a part of the executor itself and therefore that code does not live in the `driver` module.

There is also additional documentation and comments now, which should help contributors navigate the code more easily. In particular, I want people not interested in low-level details to not be lost in the codebase and people who are interested to be able to effortlessly play with the executor implementation without having to mess with anything that is unrelated to task execution.

This is the `task` module: https://github.com/stjepang/async-std/tree/refactor-task/src/task

This is the `task::driver` module: https://github.com/stjepang/async-std/tree/refactor-task/src/task/driver

The scheduler is less than 150 lines of code and this is its main part: https://github.com/stjepang/async-std/blob/refactor-task/src/task/driver/pool.rs


src/task/executor/pool.rs
```diff
@@ -0,0 +1,144 @@
...
+                    .or_else(|| {
+                        // First, pick a random starting point in the list of local queues.
+                        let len = pool.stealers.len();
+                        let start = random(len as u32) as usize;
+
+                        // Try stealing a batch of tasks from each local queue starting from the
+                        // chosen point.
+                        pool.stealers
+                            .iter()
+                            .chain(pool.stealers.iter())
```
yoshuawuyts: Heh, so how does this work? The chaining of `pool.stealers` with itself kind of confuses me.

stjepang: Yeah, it's a bit weird. I changed the code to:
```rust
// Try stealing a batch of tasks from each local queue starting from the
// chosen point.
let (l, r) = pool.stealers.split_at(start);
let rotated = r.iter().chain(l.iter());
rotated.map(|s| s.steal_batch_and_pop(&local)).collect()
```

Hopefully now it's a bit clearer!

yoshuawuyts: Ah, I see now. Yeah that's better. Thank you!

PR #422, opened 2019-10-31
Remove unused pin-utils dep

yoshuawuyts:
pin-utils is no longer used. This removes it from our dependency tree.


yoshuawuyts:
Ah dang, `pin_utils` with underscore didn't show up in search. We do this PR tho. Going to file an issue for this instead.

Issue #430, opened 2019-10-31
Replace `VecDeque` with `stream::from_iter` in examples


yoshuawuyts:
Now that https://github.com/rust-lang-nursery/futures-rs/pull/1930 has landed, we're waiting on https://github.com/async-rs/async-std/pull/415 to be merged. Once it's merged we should replace all uses of `VecDeque` with `stream::from_iter` so we're setup for the `futures@0.3.0` release.

This should be a fairly mechanical change, and easy to pick up!

k-nasa:
Will I do this with #415 ?

yoshuawuyts:
@k-nasa I think we should merge that first, and create a follow-up PR to it that ports over the examples. That way we keep moving forward without having a single massive PR that does both things at once. Would you be willing to write the follow-up PR?

PR #431, opened 2019-11-01
fix doc tests

yoshuawuyts:
It seems #409 broke our tests. The problem is that `VecDeque` implements `Ord` which has a conflicting `min` method. I've disabled the doc test for `min`, in anticipation of #430. We can re-enable it once we land #415. Thanks!


stjepang:
Needs `cargo fmt`, then it should be good.

PR #432, opened 2019-11-01
Unstabilize Arc and Weak

stjepang:
Two users on Discord have been confused by the existence of `async_std::sync::Arc` since they expected it to be somehow different from `std::sync::Arc`, but it's just a re-export. I share similar feelings. Perhaps it'd be wise to unstabilize it before publishing 1.0, just in case.

To elaborate a bit more... It's "obvious" to me how `async_std::fs::File` or `async_std::sync::Mutex` are different from their `std` counterparts. While we do re-export a bunch of types like `std::fs::FileType` or `std::net::Shutdown`, those are not important and only exist as auxiliary types supporting those that *are* important. In my mind, `Arc` is not such an auxiliary type, which is probably why it feels odd to have it re-exported. But that's just my mental model of the library :)

I should also note that it's possible I (and perhaps other users) will get warmed up to the idea of this re-export but it just doesn't sit right with me right now.

Opinions? How do you think about all this? What should we do here?
cc @yoshuawuyts @skade @matklad @taiki-e

### DIFF
--- a/src/sync/barrier.rs
+++ b/src/sync/barrier.rs
@@ -10,7 +10,9 @@ use crate::sync::Mutex;
 /// ```
 /// # async_std::task::block_on(async {
 /// #
-/// use async_std::sync::{Arc, Barrier};
+/// use std::sync::Arc;
+///
+/// use async_std::sync::Barrier;
 /// use async_std::task;
 ///
 /// let mut handles = Vec::with_capacity(10);
@@ -121,7 +123,9 @@ impl Barrier {
     /// ```
     /// # async_std::task::block_on(async {
     /// #
-    /// use async_std::sync::{Arc, Barrier};
+    /// use std::sync::Arc;
+    ///
+    /// use async_std::sync::Barrier;
     /// use async_std::task;
     ///
     /// let mut handles = Vec::with_capacity(10);
--- a/src/sync/mod.rs
+++ b/src/sync/mod.rs
@@ -121,10 +121,6 @@
 //! The following is an overview of the available synchronization
 //! objects:
 //!
-//! - [`Arc`]: Atomically Reference-Counted pointer, which can be used
-//!   in multithreaded environments to prolong the lifetime of some
-//!   data until all the threads have finished using it.
-//!
 //! - [`Barrier`]: Ensures multiple threads will wait for each other
 //!   to reach a point in the program, before continuing execution all
 //!   together.
@@ -142,7 +138,6 @@
 //!   writer at a time. In some cases, this can be more efficient than
 //!   a mutex.
 //!
-//! [`Arc`]: crate::sync::Arc
 //! [`Barrier`]: crate::sync::Barrier
 //! [`Condvar`]: crate::sync::Condvar
 //! [`channel`]: fn.channel.html
@@ -175,6 +170,8 @@
 //! # })
 //! ```
 
+#[cfg(feature = "unstable")]
+#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
 #[doc(inline)]
 pub use std::sync::{Arc, Weak};
 


skade:
I agree with that reasoning. There's a case to be made that you should be able to write all async code just by using `async-std` and `Arc` is kind of an async concern, but I disagree for 2 reasons:

* Most actual codebases will import some amount of code from `std`
* A pure re-export runs into danger of people accidentally believing `async-std`s `Arc` is different from std's `Arc`, leading to confusion with API that expect an `Arc` passed.

yoshuawuyts:
I strongly disagree with this direction. As I stated in https://github.com/async-rs/async-std/pull/403#issuecomment-547141026 ("pedagogy") I think async-std would overall benefit from clearer enumeration of which APIs we export, which APIs are adaptations of std APIs, and which APIs we've designed ourselves.

## Usability

`Mutex` is practically never used without `Arc`, and by splitting the two into separate imports we harm the overall usability of async-std. An example of the awkwardness of having a singular `std::sync::Arc` import can be seen in #419's `client.rs`:

https://github.com/async-rs/async-std/blob/d17197ff728ed12bb3878296b7527f512d0f2de9/examples/a-chat/client.rs#L1-L9

_The single import from `std` stands out. This will be a recurring pattern in async-std user's code if we follow through with this proposed change._

I don't think the right conclusion for us is to feature-flag a `std` re-export, but instead ask clearer questions on how we can communicate the nature of the types we provide. A primary goal of async-std is to enable people to be productive with Async Rust, and I believe this change would regress usability without addressing the underlying cause.

## Feedback asked

Additionally I think saying that "two users were confused on Discord" is not an accurate characterization of the conversations that were had. The fact that the conversations ended in asking a binary choice of: _"Would you want to keep or remove `Arc`"_ is a wrong way to solicit feedback.

> [stjepan to leo60228] given that you saw the discussion, would you like to have Arc in async-std or not?

> [stjepan to axie] what is in your opinion better? keeping Arc in async_std::sync or removing it from there?

Instead I think we should have asked _what_ is confusing people. Which assumptions did they have, and how was reality different? What did _they_ think we could improve, rather than offer a single choice that they could either agree or disagree with.

## Conclusion

I don't think this is the right direction for us, and I'm confused about the framing of the problem. Perhaps from a Rust expert's point of view `Mutex`, `SocketAddr`, and `Arc` can somehow be classified as primary / secondary. But I don't think that framing translates well to the wider Rust audience. Regardless of classification, people will want to know what we've defined, and what's coming from the stdlib.

The reasoning for rejecting #403 has made me wary of marking these APIs as "unstable" because I feel the burden of proof will then be turned into "please provide reasons _why_ we should include this" rather than assuming it should be included by default.

Any confusion we may have perceived from users so far has been symptoms of an underlying issue that we need to communicate better, and unstabilizing these APIs will not change the root cause.

I don't think we'll be able to agree on a direction on this thread, and it's something we should discuss in a more synchronous manner before we release 1.0. I don't want a decision on this PR to be based on whether it's stalled / pushed through -- but instead would like us to align on overall design direction so we can create consensus on where we want to take this. So perhaps let's take some time to sit down and talk this week?

nbdd0121:
Can we just remove `#[doc(inline)]`, so that it communicates clearly to the user that they are re-exported?

taiki-e:
> _The single import from `std` stands out. This will be a recurring pattern in async-std user's code if we follow through with this proposed change._

If there are multiple imports from std, I think it does not look bad. e.g...

```rust
use std::{
    collections::HashMap,
    sync::{atomic::AtomicUsize, Arc}, 
};

use async_std::{ 
    io::{stdin, BufReader}, 
    prelude::*, 
    sync::Mutex, 
}; 
```

stjepang:
@yoshuawuyts and I discussed this and we'd close this PR and leave `Arc` and `Weak` as they are.

If you have *strong* feelings about this, now is the time to speak up. Otherwise, we'll close it. :)

yoshuawuyts:
We've talked about this in person, and we're not going to go forward with this at this time. Thanks all!

PR #434, opened 2019-11-01
Put everything behind a 'stable' feature to avoid future breaking changes

### LINKED ISSUE #417: Create default `stable` feature before 1.0 to avoid future breaking changes
Hey, this comes from a discussion that happened on reddit after a announcement of this crate, basically it would be nice to split this crate in features in the future, to allow using just a fraction of it. This would make it incredibly easy to integrate this crate with others in the async ecosystem.

But splitting into features is a breaking change, since it will break the crate for whoever uses `default-features = false`, so either we do it before 1.0, or we will have to wait for 2.0.

There is a trick tho, we can avoid this breaking change if we crate a `stable` feature right now and put everything stable in it, so `default-features = false` wouldn't work and they would be forced to select the features from the start.

Discussion:
https://www.reddit.com/r/rust/comments/doos98/asyncstd_v09911_released/f5phwor/

paulocsanz:
This fixes #417, although I don't know if this is the ideal way, can anybody point me the best way?

I had to define the macro in `lib.rs` otherwise everything else on `util.rs` would have to be behind the same macro, I could move it to `utils.rs` tho.

Also, since things on `unstable` seem to depend on `stable` (does it really need to?). So for now `unstable` needs `stable`, that could become a problem when specific features are added, since you may need to play feature hide and seek with them to use the `unstable` features. What do you guys think about that?

If we put only the pub exports behind the flag a bunch of "unused" warnings will pop up, is it better than putting everything behind the flag (it may fix the `unstable` problem)? Should we just add a `allow(unused)`?

And for some weird reason `prelude.rs` stopped compiling referencing this PR rust-lang/rust#52234, which seems like a problem so I removed the `pub use`, specially because now the macro won't be in the prelude. Any ideas on that?  (this actually breaks a test, what is the best way to fix it?)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,6 +41,7 @@
 //! features = ["unstable"]
 //! ```
 
+#![cfg(feature = "default")]
 #![cfg_attr(feature = "docs", feature(doc_cfg))]
 #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
 #![allow(clippy::mutex_atomic, clippy::module_inception)]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -49,33 +49,46 @@
 #![doc(html_logo_url = "https://async.rs/images/logo--hero.svg")]
 #![recursion_limit = "2048"]
 
-#[macro_use]
-mod utils;
+/// Declares stable items.
```
stjepang: Can you move this macro into `utils.rs` to group all these macros together?

stjepang: Or, alternatively, I believe you can just put this in `lib.rs` :)

```rust
#![cfg(feature = "stable")]
```

yoshuawuyts:
I think this idea is worthwhile for us to introduce other features down the line (such as a subset of async-std that does not build the runtime). I'm not sure what the right name is, but I like the name `default`, as the only purpose for it is to be disabled through `no-default`, and then re-enable other features -- possibly `runtime` and `core`.

stjepang:
@yoshuawuyts I like that idea!

I believe this whole PR can be reduced to (seems to work for me!):
1. Add `default = []` to `Cargo.toml`.
2. Add `#![cfg(feature = "default")]` to `lib.rs`.

If we add new features like `runtime` or `core`, we'd change that attribute to:

```rust
#![cfg(any(feature = "default", feature = "runtime", feature = "core"))]
```

paulocsanz:
@stjepang @yoshuawuyts How about now?

stjepang:
Looks good to me but I've never seen a crate do this before. Is there someone more familiar with Cargo who could take a quick look?

src/lib.rs
```diff
@@ -49,46 +50,33 @@
...
-    pub mod future;
-    pub mod io;
-    pub mod net;
-    pub mod os;
-    pub mod path;
-    pub mod prelude;
-    pub mod stream;
-    pub mod sync;
-    pub mod task;
+pub mod fs;
```
yoshuawuyts: I think we may want to keep this behind the `default` flag. If we change the behavior of passing `no-default` that should be considered a breaking change.

taiki-e:
I recently confirmed this behavior when working on cargo-hack: https://github.com/taiki-e/cargo-hack/blob/75bff8d3c45e72e2075daf998b7dd179c8f7848c/src/main.rs#L243-L246
```rust
    // run with no default features if the package has other features
    //
    // `default` is not skipped because `cfg(feature = "default")` is work
    // if `default` feature specified.
```

src/lib.rs
stjepang: The whole library is now behind the `default` flag. (or perhaps I misunderstood your point?)

PR #441, opened 2019-11-02
mark stdio-lock structs as unstable

yoshuawuyts:
Follow-up to https://github.com/async-rs/async-std/pull/334, this marks some of the extra stdio structs as "unstable". It was already never possible to create them so merging the patch was already correct. But this adds some extra polish. Thanks heaps!

### DIFF
--- a/src/io/stdout.rs
+++ b/src/io/stdout.rs
@@ -1,4 +1,3 @@
-use std::io::Write as StdWrite;
 use std::pin::Pin;
 use std::sync::Mutex;
 
@@ -8,6 +7,7 @@ use crate::task::{spawn_blocking, Context, JoinHandle, Poll};
 
 cfg_unstable! {
     use once_cell::sync::Lazy;
+    use std::io::Write as _;
 }
 
 /// Constructs a new handle to the standard output of the current process.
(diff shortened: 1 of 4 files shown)


src/io/stdout.rs
```diff
@@ -59,13 +59,19 @@ pub fn stdout() -> Stdout {
...
 /// A locked reference to the Stderr handle.
-/// This handle implements the [`Write`] traits, and is constructed via the [`Stdout::lock`] method.
+///
+/// This handle implements the [`Write`] traits, and is constructed via the [`Stdout::lock`]
+/// method.
 ///
 /// [`Write`]: trait.Read.html
 /// [`Stdout::lock`]: struct.Stdout.html#method.lock
+#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
+#[cfg(any(feature = "unstable", feature = "docs"))]
```
stjepang: I changed how the `unstable` feature flag work recently (now `docs` depends on `unstable`) now it's enough to just do this:

```rust
#[cfg(feature = "unstable")]
#[cfg_attr(feature = "docs", doc(cfg(unstable)))]
```

yoshuawuyts: ah nice, didn't realize. Updated!

PR #442, opened 2019-11-02
add Stream::copied

zhangguyu6:
ref #129, Implement Steam::copied.
# References
* https://doc.rust-lang.org/src/core/iter/adapters/mod.rs.html#136-171
* https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#2298-2323

### DIFF
--- a/src/stream/stream/copied.rs
+++ b/src/stream/stream/copied.rs
@@ -1,0 +1,33 @@
+use crate::stream::Stream;
+use crate::task::{Context, Poll};
+use pin_project_lite::pin_project;
+use std::pin::Pin;
+
+pin_project! {
+    #[doc(hidden)]
+    #[allow(missing_debug_implementations)]
+    pub struct Copied<S> {
+        #[pin]
+        stream: S,
+    }
+}
+
+impl<S> Copied<S> {
+    pub(super) fn new(stream: S) -> Self {
+        Copied { stream }
+    }
+}
+
+impl<'a, S, T: 'a> Stream for Copied<S>
+where
+    S: Stream<Item = &'a T>,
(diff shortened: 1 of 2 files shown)


k-nasa:
This code con't run.
```rust
        use async_std::prelude::*;
        use std::collections::VecDeque;

        let v: VecDeque<_> = vec![&1, &2, &3].into_iter().collect();

        let mut v_copied  = v.copied();

        dbg!(v);
        dbg!(v_copied);
    });
```
I got this error message.
```
src/main.rs:11:14
   |
7  |         let v: VecDeque<_> = vec![&1, &2, &3].into_iter().collect();
   |             - move occurs because `v` has type `std::collections::VecDeque<&i32>`, which does not implement the `Copy` trait
8  |
9  |         let mut v_copied  = v.copied();
   |                             - value moved here
10 |
11 |         dbg!(v);
   |              ^ value used here after move
```
Is the `copied` method meaningful?

stjepang:
Oh no, I think I didn't look closely enough, my bad! :( I'll try to fix in a follow-up PR

zhangguyu6:
copied is the same as .map(|&x| x), consume an stream that copy each element from &T to T.

This code: 
```rust
use async_std::prelude::*;
use async_std::stream;
use std::collections::VecDeque;


fn main() {
    task::block_on(async {
        let v : VecDeque<_> = vec![1, 2, 3].into_iter().collect();

        let mut s = stream::from_iter(v.iter());

        let mut v_copied = s.copied();

        dbg!(v);

    });
}
```

output:
```
[examples/hello-world.rs:15] v = [
    1,
    2,
    3,
  
```

PR #446, opened 2019-11-03
Add Stream cloned

yjhmelody:
#129

### DIFF
--- a/src/stream/stream/cloned.rs
+++ b/src/stream/stream/cloned.rs
@@ -4,7 +4,7 @@ use pin_project_lite::pin_project;
 use std::pin::Pin;
 
 pin_project! {
-    #[allow(missing_debug_implementations)]
+    #[derive(Debug)]
     pub struct Cloned<S> {
         #[pin]
         stream: S,
--- a/src/stream/stream/cloned.rs
+++ b/src/stream/stream/cloned.rs
@@ -4,7 +4,6 @@ use pin_project_lite::pin_project;
 use std::pin::Pin;
 
 pin_project! {
-    #[doc(hidden)]
     #[allow(missing_debug_implementations)]
     pub struct Cloned<S> {
         #[pin]
--- a/src/stream/stream/cloned.rs
+++ b/src/stream/stream/cloned.rs
@@ -1,0 +1,33 @@
+use crate::stream::Stream;
+use crate::task::{Context, Poll};
+use pin_project_lite::pin_project;
+use std::pin::Pin;
+
+pin_project! {
+    #[doc(hidden)]
+    #[allow(missing_debug_implementations)]
+    pub struct Cloned<S> {
+        #[pin]
+        stream: S,
+    }
+}
+
+impl<S> Cloned<S> {
+    pub(super) fn new(stream: S) -> Self {
+        Self { stream }
+    }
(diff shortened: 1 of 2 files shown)


src/stream/stream/cloned.rs
```diff
@@ -0,0 +1,33 @@
+use crate::stream::Stream;
+use crate::task::{Context, Poll};
+use pin_project_lite::pin_project;
+use std::pin::Pin;
+
+pin_project! {
+    #[doc(hidden)]
```
k-nasa:
```suggestion

```

src/stream/stream/cloned.rs
```diff
@@ -0,0 +1,33 @@
...
+    #[allow(missing_debug_implementations)]
```
k-nasa:
```suggestion
    #[derive(Debug)]
```

src/stream/stream/mod.rs
```diff
@@ -373,6 +375,40 @@ extension_trait! {
...
+            assert_eq!(v_cloned.next().await, Some(1));
+            assert_eq!(v_cloned.next().await, Some(2));
+            assert_eq!(v_cloned.next().await, Some(3));
+            assert_eq!(v_cloned.next().await, None);
+
+            #
+            # }) }
+            ```
+        "#]
+        fn cloned<'a,T>(self) -> Cloned<Self>
```
k-nasa: I tryed running this code.
```rust
fn main() -> std::io::Result<()> {
        use async_std::prelude::*;
        use std::collections::VecDeque;
        let v: VecDeque<_> = vec![&1, &2, &3].into_iter().collect();
        let v_cloned = v.cloned();
        dbg!(v);
        dbg!(v_cloned);
    });

    Ok(())
}
```


I god this error.
```console
error[E0382]: use of moved value: `v`
 --> src/main.rs:7:14
  |
5 |         let v: VecDeque<_> = vec![&1, &2, &3].into_iter().collect();
  |             - move occurs because `v` has type `std::collections::VecDeque<&i32>`, which does not implement the `Copy` trait
6 |         let v_cloned = v.cloned();
  |                        - value moved here
7 |         dbg!(v);
  |              ^ value used here after move
```

Is it appropriate to move ownership?

taiki-e: > Is it appropriate to move ownership?

This should be fine because `cloned` clones each element of the stream, not the stream itself.

Refs: [std::iter::Iterator::cloned](https://doc.rust-lang.org/nightly/core/iter/trait.Iterator.html#method.cloned)

k-nasa: thanks!

src/stream/stream/mod.rs
```diff
@@ -372,6 +374,40 @@ extension_trait! {
...
+
+            Basic usage:
+
+            ```
+            # fn main() { async_std::task::block_on(async {
+            #
+            use async_std::prelude::*;
+            use std::collections::VecDeque;
+
+            let v: VecDeque<_> = vec![&1, &2, &3].into_iter().collect();
```
taiki-e: Could you replace `VecDeque` with `stream::from_iter`? (like https://github.com/async-rs/async-std/pull/447)

PR #447, opened 2019-11-03
Replace `VecDeque` with `stream::from_iter` in examples

k-nasa:
close: https://github.com/async-rs/async-std/issues/430


yoshuawuyts:
@k-nasa thank you so much for making this PR! This was one of the biggest obstacles we had to clear for the upcoming 1.0 release, and thanks to you we're now good! Really thankful for the work you've put in!

PR #448, opened 2019-11-03
Clarify blocking in channel docs

### LINKED ISSUE #444: Doc: Sender state that send will be blocked
Documentation states that `Sender::send` will "block the current task until the send operation can proceed". The word "block" is potentially misleading, because it is usually applied as synonymous to "will block thread". Could it be reformulated with word "await" somehow? For example: "If the channel is full, this method will await until operation can proceed."
https://github.com/async-rs/async-std/blob/fa91d7f856a9f0afc9689371ced6f1c941c346bc/src/sync/channel.rs#L105

stjepang:
Closes #444 

@vchekan Does this look better? :)

### DIFF
--- a/src/sync/channel.rs
+++ b/src/sync/channel.rs
@@ -22,7 +22,7 @@ use crate::sync::WakerSet;
 ///
 /// Senders and receivers can be cloned. When all senders associated with a channel get dropped, it
 /// becomes closed. Receive operations on a closed and empty channel return `None` instead of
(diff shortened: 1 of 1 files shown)


vchekan:
Yes, thanks! Now I do not have to write tests to confirm that it does what I think it should do :)

stjepang:
Thanks for filing the issue and reviewing this PR! :)

Issue #453, opened 2019-11-04
Expose future types from future::ready and future::pending


yoshuawuyts:
Today I was working on Tide, and came across the need for having the future type returned by `future::ready`. Namely I wanted to implement a no-op as part of a trait that expects a future to be returned.

The signature is:
```rust
    type ConnectionFuture: Future;

    fn connect(&self) -> Self::ConnectionFuture {
        ReadyFuture {}
    }
```

`fn connect` returns `ConnectionFuture`. It doesn't do anything, so using `future::ready` seemed like a nice option. Except since it's an async fn we're not able to enumerate it.

For this reason I'd like to propose exposing the futures from `future::ready` and `future::pending` so they can be used as no-ops, or thin wrappers and enumerated inside types. Thanks!

stjepang:
Any reason why not use `Pin<Box<dyn Future<Output = ...>>>` or the `async-trait` crate?

This seems like the classic `async fn` in traits problem :)

yoshuawuyts:
@stjepang I guess the answer is perf. But yeah maybe we just shouldn't. Unsure, heh.

PR #454, opened 2019-11-04
expose future::{Ready,Pending}

yoshuawuyts:
Implements #453. Thanks!

## Screenshots
Screenshot_2019-11-05 async_std future - Rust

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -50,9 +50,9 @@
 pub use async_macros::{join, try_join};
 
 pub use future::Future;
-pub use pending::pending;
+pub use pending::{pending, Pending};
 pub use poll_fn::poll_fn;
-pub use ready::ready;
+pub use ready::{ready, Ready};
 pub use timeout::{timeout, TimeoutError};
 
 pub(crate) mod future;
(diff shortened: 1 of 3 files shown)


stjepang:
If we have `Pending` and `Ready` structs, should we also have `Timeout` and `PollFn`?

yoshuawuyts:
@stjepang possibly?

stjepang:
Yeah, I would say so for consistency.

taiki-e:
> If we have `Pending` and `Ready` structs, should we also have `Timeout` and `PollFn`?

Should we also expose FuturesExt combinators?

stjepang:
@taiki-e That's a good point. If we are exposing these futures, then it would logically follow that we should probably also expose futures returned by `FutureExt` combinators. So either we expose all future types or none.

On the other hand, we could make a rule that only future *constructors* should expose futures, while future *combinators* shouldn't.

@yoshuawuyts What's the status of this PR?

yoshuawuyts:
@stjepang happy to close it — I'm unsure at this point and just not having it seems fine.

PR #464, opened 2019-11-06
Rename stream_extend to extend

stjepang:
I've renamed this method to be more consistent with the standard library and re-exported the `Extend` trait in our prelude.

### DIFF
--- a/src/collections/binary_heap/extend.rs
+++ b/src/collections/binary_heap/extend.rs
@@ -2,10 +2,10 @@ use std::collections::BinaryHeap;
 use std::pin::Pin;
 
 use crate::prelude::*;
-use crate::stream::{Extend, IntoStream};
+use crate::stream::{self, IntoStream};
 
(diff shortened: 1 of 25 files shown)


stjepang:
@yoshuawuyts How do you feel about these ideas (all of them are independent of each other)?

* We don't put our `Extend` into the prelude so that synchronous `foo.extend(bar);` still works.

* Instead of doing `stream::Extend::extend(&mut foo, bar).await`, first import with `use async_std::stream::Extend::extend` and then do `extend(&mut foo, bar).await`.

* Or, perhaps even re-export a convenience method called `extend` into the `stream` module so that one can do `stream::extend(&mut foo, bar).await`.

stjepang:
Here's an idea: what if both `std::iter::extend()` and `async_std::stream::extend()` existed? 

That'd be nice and would make sync/async versions convenient and symmetrical at the same time :)

Let's discuss this offline one more time (briefly) and make a final decision. Thanks for clarifying motivation for this change last time we discussed it, that was helpful!

Issue #365, opened 2019-10-17
Can we simplify io::copy?


stjepang:
The current signature of `async_std::io::copy` is:

```rust
pub async fn copy<R, W>(reader: &mut R, writer: &mut W) -> io::Result<u64>
where
    R: Read + Unpin + ?Sized,
    W: Write + Unpin + ?Sized;
```

Both reader and writer need to be mutable references because we're just mimicking [`std::io::copy`](https://doc.rust-lang.org/std/io/fn.copy.html).

Unfortunately, this API is annoying to use when we want to share `TcpStream`s. Just look at this convoluted `&mut (&stream, &stream)` pattern:

```rust
async fn process(stream: TcpStream) -> io::Result<()> {
    let (reader, writer) = &mut (&stream, &stream);
    io::copy(reader, writer).await?;
    Ok(())
}
```

I think we can do better. What if we didn't follow the API from `std` and had the following instead?

```rust
pub async fn copy<R, W>(reader: R, writer: W) -> io::Result<u64>
where
    R: Read + Unpin,
    W: Write + Unpin;
```

This might look like a less powerful APIs than the previous one, but I believe it is functionally the same. Note that we have these blanket impls of `Read` and `Write`:

```rust
impl<T: Read + Unpin + ?Sized> Read for &mut T {}
impl<T: Write + Unpin + ?Sized> Write for &mut T {}
```

That means if we can pass `&mut T` into the previous API, then it should be totally fine to pass it into the new one too! In other words, the new API is fully compatible with the previous one (unless I'm missing something here).

So the cool thing is that while it might seem we're deviating from the `std` APIs, we kind of aren't. :)

Here's how we can write an echo TCP server using the new `async_std::io::copy`:

```rust
async fn process(stream: TcpStream) -> io::Result<()> {
    io::copy(&stream, &stream).await?;
    Ok(())
}
```

And here's how we do it on `Arc<TcpStream>`:

```rust
async fn process(stream: Arc<TcpStream>) -> io::Result<()> {
    io::copy(&*stream, &*stream).await?;
    Ok(())
}
```

This change could make sharing streams a lot easier, and we've seen plenty of people struggle with this problem. Wdyt?

stjepang:
cc @skade @yoshuawuyts

stjepang:
I believe we could make the same change in the standard library. *scratches head*

yoshuawuyts:
[...]

PR #471, opened 2019-11-07
Unstable feature: copy takes arguments by value

stjepang:
This just makes our `io::copy()` more generic so that one can do this:

```rust
io::copy(&stream, &stream).await?;
```

instead of:

```rust
let (reader, writer) = &mut (&stream, &stream);
io::copy(reader, writer).await?;
```

This feature is only available when `unstable` feature flag is turned out.

cc #365

### DIFF
--- a/src/io/copy.rs
+++ b/src/io/copy.rs
@@ -43,10 +43,98 @@ use crate::task::{Context, Poll};
 /// #
 /// # Ok(()) }) }
 /// ```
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
@stjepang could you say more about why both versions are kept? I was thinking we keep only a single version around, but mark it unstable.

I don't think we'd be able to stabilize this way because it would break existing code wouldn't it? Or is the current trait forward-compatible?

PR #474, opened 2019-11-07
Fix deadlock when all receivers are dropped

### LINKED ISSUE #472: Drop the channel receiver first, then the sender await forever
Whatever the capacity is, drop the receiver first, then the sender await forever, test with release `v0.99.11`.

```rust
    task::block_on(async {
        let (s, r) = channel(10);
        drop(r);
        s.send(1).await;
    });
```

stjepang:
Closes #472

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,14 +74,14 @@ jobs:
     - name: Docs
       run: cargo doc --features docs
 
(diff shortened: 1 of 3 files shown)


stjepang:
I've disabled clippy for now and added a comment to the code. Hopefully it's a bit clearer now but we can revisit later if not.

PR #475, opened 2019-11-07
Spawn more than one blocking thread

stjepang:
This fixes a bug where only one thread was ever spawned :)


stjepang:
Just did some interactive testing by spawning a bunch of blocking tasks that sleep for a random number of time and it seems the blocking pool now does exactly what it should: more threads are spawned on demand and we stop threads when they have no work to do.

PR #476, opened 2019-11-07
Add "std" feature flag

yoshuawuyts:
This adds a "std" feature flag, allowing to opt-in to only the core traits and types. No runtime. Thanks!

## Screenshot

__default__
Screenshot_2019-11-08 async_std - Rust

__std__
Screenshot_2019-11-08 async_std - Rust(1)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,11 @@ features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
-default = []
+default = ["core"]
 docs = ["unstable", "attributes"]
 unstable = ["broadcaster"]
 attributes = ["async-attributes"]
+core = []
 
 [dependencies]
 async-attributes = { version = "1.1.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,33 +21,55 @@ features = ["docs"]
 rustdoc-args = ["--cfg", "feature=\"docs\""]
 
 [features]
-default = ["core"]
-docs = ["unstable", "attributes"]
-unstable = ["broadcaster"]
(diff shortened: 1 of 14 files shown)


yoshuawuyts:
only thing I couldn't figure out was how to keep exporting task_local from the prelude; suggestions welcome!

Cargo.toml
```diff
@@ -21,32 +21,55 @@ features = ["docs"]
...
+  "mio",
+  "mio-uds",
+  "num_cpus",
+  "pin-project-lite",
+]
+docs = ["unstable"]
+unstable = ["default", "broadcaster"]
+std = [
+  "async-macros",
+  "crossbeam-channel",
```
stjepang: I believe we can delete `crossbeam-channel` from here and move it into `default = [ ... ]`

Cargo.toml
```diff
@@ -21,32 +21,55 @@ features = ["docs"]
...
+  "crossbeam-utils",
+  "futures-core",
+  "futures-io",
+  "futures-timer",
```
stjepang: Timers should also belong in `runtime` since they need a background thread... which is a runtime :)

src/sync/waker_set.rs
```diff
@@ -8,11 +8,11 @@ use std::cell::UnsafeCell;
 use std::ops::{Deref, DerefMut};
 use std::sync::atomic::{AtomicUsize, Ordering};
```
stjepang:
```suggestion
```

stjepang: (can delete line)

stjepang:
I resolved my comments and put `Future::delay()` behind `#[cfg(feature = "default")]`.

Issue #470, opened 2019-11-07
Add Future::join and Future::try_join


yoshuawuyts:
Similar to #405's addition of `Future::race` we should expose `Future::join` and `Future::try_join` as well. However unlike #405 we would keep the macro around because it's the only way of creating a variable-sized tuple when joining multiple results ("return-type polymorphism").

This would be a chance as well to update the `future` module-level docs to talk about `Future::join` vs `Future::race`, making it more symmetrical.

This should be a fun challenge to pick up for whoever is interested. Marking it as "good first issue". Thanks!

derekdreery:
I think this issue has been resolved?

yoshuawuyts:
@derekdreery indeed it is. Thanks for pointing it out!

PR #478, opened 2019-11-08
Add Future::join and Future::try_join

portgasd666:
Addresses #470 .

@yoshuawuyts Regarding the module level documentation changes, is the plan to replace `join!` and `try_join!` in the documentation with `join` and `try_join` ?

### DIFF
--- a/src/future/future/join.rs
+++ b/src/future/future/join.rs
@@ -1,0 +1,62 @@
+use std::pin::Pin;
+
+use async_macros::MaybeDone;
+use pin_project_lite::pin_project;
+
+use crate::task::{Context, Poll};
+use std::future::Future;
+
+pin_project! {
+    #[allow(missing_docs)]
+    #[allow(missing_debug_implementations)]
+    pub struct Join<L, R>
+    where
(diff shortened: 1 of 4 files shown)


stjepang:
I like this change. 👍

To be honest, I don't find `join!` and `try_join!` all that useful nor expect those macros to pop up in practice. It seems a lot of people are confused by future joining in general (see [here](https://www.reddit.com/r/rust/comments/dt6u0s/joined_futures_will_run_concurrently_not_in/)).

Instead, I expect people will spawn futures to make them run in paralle/concurrently rather than joining futures.

wdyt, @yoshuawuyts?

yoshuawuyts:
@stjepang you might be right actually. Agree we can probably remove them without much loss.

portgasd666:
I have changed the reference to `join!` and `try_join!` in the module documentation to refer to 'join' and 'try_join` functions instead.

stjepang:
@yoshuawuyts @portgasd666 In that case, let's also remove macros `join!` and `try_join!`? :)

PR #481, opened 2019-11-08
fix attributes feature

yoshuawuyts:
This fixes the attributes feature. Apparently I forgot to push a patch yesterday as part of #476, which has introduced a bug. This fixes that. Thanks!


stjepang:
Let's also check `--features attributes` in CI to prevent this in the future :)

PR #487, opened 2019-11-09
Cleanup stream traits

stjepang:


stjepang:
@yoshuawuyts do you think that traits like `Extend`, `FromStream`, and `IntoStream` should be implementable by users or should we hide return types behind `ImplFuture`?

yoshuawuyts:
@stjepang dang, good question. I think allowing people to implement them is okay. I think once we have async traits we may want to push another major anyway, so we can change that constraint at that point.

PR #489, opened 2019-11-09
Implement FromStr for PathBuf

jaysonsantos:
This makes PathBuf compatible with std version as you can simply call:
```rust
let path: PathBuf = FromStr::from_str(s).unwrap()
```


yoshuawuyts:
Reference from std: https://doc.rust-lang.org/src/std/path.rs.html#1542-1548

PR #491, opened 2019-11-09
note on Stream::merge ordering

yoshuawuyts:
Adds a note on `Stream::merge` ordering. Thanks!

### DIFF
--- a/src/stream/stream/mod.rs
+++ b/src/stream/stream/mod.rs
@@ -1656,7 +1656,7 @@ extension_trait! {
             Combines multiple streams into a single stream of all their outputs.
 
             Items are yielded as soon as they're received, and the stream continues yield until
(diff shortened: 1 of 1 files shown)


src/stream/stream/mod.rs
```diff
@@ -1655,8 +1655,8 @@ extension_trait! {
         #[doc = r#"
             Combines multiple streams into a single stream of all their outputs.
 
-            Items are yielded as soon as they're received, and the stream continues yield until both
-            streams have been exhausted.
+            Items are yielded as soon as they're received, and the stream continues yield until
+            both streams have been exhausted. The output ordering between stream is not guaranteed.
```
stjepang:
```suggestion
            both streams have been exhausted. The output ordering between streams is not guaranteed.
```

Issue #490, opened 2019-11-09
Throughput improvement to Stream::merge


yoshuawuyts:
In #437 we fixed bugs to `Stream::merge`. There is another improvement we may want to apply here:

## Randomization

If two items from `Stream::merge` are ready at the same time, we will always pick the item from the first stream first. We should probably apply some randomization here to make it so we don't exhaust one stream completely before starting the other. This should reduce overall latency in some cases, as it becomes more "fair".

We can use `utils::random` for this purpose.

portgasd666:
If this were implemented what would be the correct way to handle the non-determinism for the purpose of tests ?

Razican:
Hi I would like to work on this if nobody is working on it :)

yoshuawuyts:
This has been merged as part of #503. Thanks!

PR #503, opened 2019-11-11
Randomize Stream::merge to improve the throughput.

Razican:
This implements #490. Let me know of any improvements that can be done!

### DIFF
--- a/src/stream/stream/merge.rs
+++ b/src/stream/stream/merge.rs
@@ -45,26 +45,25 @@ where
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let this = self.project();
         if utils::random(1) == 1 {
-            poll_next_in_order(cx, this.left, this.right)
+            match this.left.poll_next(cx) {
+                Poll::Ready(Some(item)) => Poll::Ready(Some(item)),
+                Poll::Ready(None) => this.right.poll_next(cx),
+                Poll::Pending => match this.right.poll_next(cx) {
+                    Poll::Ready(Some(item)) => Poll::Ready(Some(item)),
+                    Poll::Ready(None) => Poll::Pending,
+                    Poll::Pending => Poll::Pending,
+                },
+            }
         } else {
-            poll_next_in_order(cx, this.right, this.left)
+            match this.right.poll_next(cx) {
+                Poll::Ready(Some(item)) => Poll::Ready(Some(item)),
+                Poll::Ready(None) => this.left.poll_next(cx),
+                Poll::Pending => match this.left.poll_next(cx) {
+                    Poll::Ready(Some(item)) => Poll::Ready(Some(item)),
+                    Poll::Ready(None) => Poll::Pending,
+                    Poll::Pending => Poll::Pending,
+                },
+            }
         }
-    }
-}
-
-/// Pools the next item, trying in order, first the first item, then the second one.
-fn poll_next_in_order<F, S, T>(cx: &mut Context<'_>, first: F, second: S) -> Poll<Option<T>>
-where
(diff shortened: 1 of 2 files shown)


Razican:
Hi, so I made 3 commits. The first one was a bit of a fail. Tests passed on my laoptop because this was an unstable function.

The second commit is what I think it should look like: one function to not duplicate code, but I was getting an error saying that `first`/`second` didn't implement `Stream`, which they do. So I don't understand the issue.

Duplicating the code in the two branches of the `if` works, though.

Do you know why the second commit might be failing?

Razican:
Hi, I fixed the deduplication of the code, so I think this is ready for merging.

src/stream/stream/merge.rs
```diff
@@ -40,14 +44,30 @@ where
...
+    first: Pin<&mut F>,
+    second: Pin<&mut S>,
+    cx: &mut Context<'_>,
+) -> Poll<Option<I>>
+where
+    F: Stream<Item = I>,
+    S: Stream<Item = I>,
+{
+    match first.poll_next(cx) {
+        Poll::Ready(Some(item)) => Poll::Ready(Some(item)),
```
k-nasa:
```suggestion
        Poll::Ready(item) => Poll::Ready(item),
```
I think either is fine, but I thought it was a little redundant.
However, the above changes will not match None, so the order must be changed.
Which do you think is better?

```rust
Poll::Ready(None) => second.poll_next(cx),
Poll::Ready(item) => Poll::Ready(item),
```

src/stream/stream/merge.rs
```diff
@@ -40,14 +44,30 @@ where
...
+        Poll::Ready(None) => second.poll_next(cx),
+        Poll::Pending => match second.poll_next(cx) {
+            Poll::Ready(Some(item)) => Poll::Ready(Some(item)),
```
k-nasa:
```suggestion
            Poll::Ready(item) => Poll::Ready(item),
```
The same applies here

src/stream/stream/mod.rs
```diff
@@ -1670,11 +1670,14 @@ extension_trait! {
...
-            assert_eq!(s.next().await, Some(1u8));
-            assert_eq!(s.next().await, Some(2u8));
-            assert_eq!(s.next().await, Some(3u8));
-            assert_eq!(s.next().await, None);
+            while let Some(n) = s.next().await {
+                lst.push(n)
+            }
+
+            lst.sort_unstable();
+            assert_eq!(&lst, &[1u8, 2u8, 3u8]);
```
k-nasa: 👍

src/stream/stream/mod.rs
```diff
@@ -1670,11 +1670,14 @@ extension_trait! {
...
             let mut s = a.merge(b).merge(c);
+            let mut lst = Vec::new();
 
-            assert_eq!(s.next().await, Some(1u8));
-            assert_eq!(s.next().await, Some(2u8));
-            assert_eq!(s.next().await, Some(3u8));
-            assert_eq!(s.next().await, None);
+            while let Some(n) = s.next().await {
+                lst.push(n)
+            }
```
k-nasa: I think I can write like this.

```rust
use async_std::stream::FromStream;
let v = Vec::from_stream(stream).await;
```

Razican: Yes, this makes it more clear :)

src/stream/stream/merge.rs
Razican: Hmmm yes, I will review this, better if we remove redundancy. I just used the code that was already in place, but this change looks better.

Razican:
I made the requested changes and reduced the number of match arms in the second match (two arms had the same value).

Issue #495, opened 2019-11-10
§3.4 (Sending Messages) in the book requires a different version of futures


Ppjet6:
## Steps to reproduce

- `futures-preview = { version = "0.3.0-alpha.19", features = [ "async-await" ] }` in `Cargo.toml` as recommended in §3.1.
- Try to build the example in https://book.async.rs/tutorial/sending_messages.html

```rust

use futures::channel::mpsc; // 1
use futures::sink::SinkExt;
use std::sync::Arc;

type Sender<T> = mpsc::UnboundedSender<T>; // 2
type Receiver<T> = mpsc::UnboundedReceiver<T>;

async fn connection_writer_loop(
    mut messages: Receiver<String>,
    stream: Arc<TcpStream>, // 3
) -> Result<()> {
    let mut stream = &*stream;
    while let Some(msg) = messages.next().await {
        stream.write_all(msg.as_bytes()).await?;
    }
    Ok(())
}
```

## What happens

```
error[E0599]: no method named `next` found for type `futures_channel::mpsc::UnboundedReceiver<std::string::String>` in the current scope
  --> core/src/main.rs:89:36
   |
89 |     while let Some(msg) = messages.next().await {
   |                                    ^^^^ method not found in `futures_channel::mpsc::UnboundedReceiver<std::string::String>`
   |
   = note: the method `next` exists but the following trait bounds were not satisfied:
           `futures_channel::mpsc::UnboundedReceiver<std::string::String> : async_std::stream::stream::StreamExt`
   = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
   |
7  | use futures_util::stream::StreamExt;
   |
```

It seems to work with `futures = "0.3.0"`, and it's indeed what is being used in the `a-chat` example in the repo as of 122e87364bef463c5afbf7681ec3ce35a3a7f577.

yoshuawuyts:
@Ppjet6 Thanks for filing this! We should indeed fix this. I've marked this as "good first issue" as it should be relatively easy to update.

dunnock:
I like to take it if nobody is working on it now

Enselic:
As far as I can tell, this issue can be closed now, because this was fixed by https://github.com/async-rs/async-std/commit/4aa9928ece3d968ea229842cbdaf53dd81fc11c2#diff-a66f64c4985b3614f93e8bba0d7b7cd879f8db631a2ee2f4afd55973ad21de9a.

PR #504, opened 2019-11-11
fixing tutorial requirements #495

dunnock:
Updated dependencies and rustc version in docs / tutorial, so that example would compile and run


dunnock:
It seems was fixed in another PR

Issue #499, opened 2019-11-10
Missing properties on net::TcpStream


dignifiedquire:
When working with `TcpStream` I realized some things are still missing, this is a list for tracking those

- [x] `connect`
- [ ] `connect_timeout`
- [x] `local_addr`
- [ ] `nodelay`
- [ ] `peek`
- [x] `peer_addr`
- [ ] `read_timeout`
- [x] `set_nodelay`
- [ ] `set_nonblocking`
- [ ] `set_read_timeout`
- [x] `set_ttl`
- [ ] `set_write_timeout`
- [x] `shutdown`
- [ ] `take_error`
- [ ] `try_clone` (I remember some discussion, so might not be happening)
- [x] `ttl`
- [ ] `write_timeout`

tomaka:
Aren't all methods that end with `_timeout` hacks due to the synchronous nature of these operations in the stdlib? Do we really need to reproduce them in an asynchronous context?

dignifiedquire:
@tomaka they could still be useful, as convenience methods. Alternatively adding some documentation on how to achieve the same effect with the new tooling might be a better way to go.

ArtemGr:
Regarding the read calls.
Let's say that I've asked to read 7 bytes and the other side has sent me 6 bytes ("foobar").

Will the `TcpStream::poll_read` immediately return me `Ready(Ok(6))`?
Can I be sure that it won't be `Pending` until a seventh byte arrives?

Similarly, will an `await` on the `TcpStream::read` return the 6 bytes soon as they arrive? Or will it keep waiting for the 7th byte?

This is usually under-specified in, say, POSIX `read` calls and hence we have to read byte-by-byte from some streams if we want to access the incoming data as soon as it is available from the network interface. Would be nice to avoid these byte-by-byte reads though.

SuperHacker-liuan:
keepalive & set_keepalive is useful when peer suddenly lose its access to network.

Fishrock123:
`keepalive` is an HTTP thing, not a tcp thing?

SuperHacker-liuan:
> `keepalive` is an HTTP thing, not a tcp thing?

FYR, https://tools.ietf.org/html/rfc1122#page-101

PR #507, opened 2019-11-11
feat(net): implement more methods for TcpStream

dignifiedquire:
- `connect_timeout`
- `take_error`

Ref #499


src/net/tcp/stream.rs
```diff
@@ -96,6 +97,18 @@ impl TcpStream {
         }))
     }
 
+    /// Opens a TCP connection to a remote host with a timeout.
+    ///
+    /// Unlike `connect`, `connect_timeout` takes a single `SocketAddr` since
+    /// timeout must be applied to individual addresses.
+    ///
+    /// It is an error to pass a zero `Duration` to this function.
```
stjepang: This doc comment is missing an example. :)

PR #509, opened 2019-11-11
docs(readme): add ci status badge

fundon:


README.md
```diff
@@ -8,6 +8,11 @@
 <br />
 
 <div align="center">
+   <!-- CI status -->
+  <a href="https://github.com/async-rs/async-std/actions">
+    <img src="https://github.com/async-rs/async-std/workflows/CI/badge.svg"
```
stjepang: It would be nice if we could choose `?style=flat-square` here but I don't think GitHub's badges have configurable style. Maybe I'm wrong though.

fundon: Currently, GitHub cant customize style. Maybe wait https://github.com/badges/shields/issues/2574.

taiki-e:
Which branch's status is pointed?
As we are not running CI on the master branch, if this is reporting the status of master, the status is pointing to the old commit.
On the other hand, if it points to another branch, the status may be from a PR that has not landed.

(I confirmed the above behavior, I removed the GitHub actions badge from all my projects [about a month ago](https://github.com/taiki-e/futures-async-stream/commit/ad5e2f9385ad050c3314d2a643ee0a149df1c214), but it may have changed now)

fundon:
More details see this: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow

taiki-e:
@fundon Thanks, I'll look that.

taiki-e:
Docs says:

> By default, badges display the status of your default branch (usually master).

As we are not running CI on the master branch, the status of one month ago is currently displayed.
So, to activate the badge I think we need a branch where CI is always run (i.e., need to enable CI on `master`).

dignifiedquire:
Thank you

PR #512, opened 2019-11-11
Add proof of concept serde support with PathBuf

jamesmunns:
This is a proof of concept for supporting types from `async_std` to be used with Serde. For now I have only implemented support for PathBuf, but if you like this, we can start rolling it out to all of the wrapped types.

```rust
// src/lib.rs:
use async_std;
use serde::{self, Serialize, Deserialize};

#[derive(Serialize, Deserialize)]
struct Hmmm {
    ex: async_std::path::PathBuf,
}

fn main() {
    println!("Hello, world!");
}
```

```toml
# Cargo.toml
[package]
name = "serde-ex"
version = "0.1.0"
authors = ["James Munns <james.munns@ferrous-systems.com>"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies.async-std]
version = "1.0"
path = "../async-std"

[dependencies.serde]
features = ["derive"]
version = "1.0"
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,10 +33,14 @@ default = [
   "mio-uds",
   "num_cpus",
   "pin-project-lite",
+  "serde-remotes",
 ]
 docs = ["attributes", "unstable"]
 unstable = ["default", "broadcaster"]
 attributes = ["async-attributes"]
+serde-support = [
+  "serde",
+]
 std = [
   "async-macros",
   "crossbeam-utils",
@@ -70,6 +74,7 @@ once_cell = { version = "1.2.0", optional = true }
 pin-project-lite = { version = "0.1", optional = true }
 pin-utils = { version = "0.1.0-alpha.4", optional = true }
 slab = { version = "0.4.2", optional = true }
+serde = { version = "1.0.0", optional = true, features = ["derive"] }
 
 [dev-dependencies]
 femme = "1.2.0"
--- a/src/path/pathbuf.rs
+++ b/src/path/pathbuf.rs
@@ -13,12 +13,16 @@ use crate::path::Path;
 use crate::prelude::*;
 #[cfg(feature = "unstable")]
 use crate::stream::{self, FromStream, IntoStream};
+#[cfg(feature = "serde-support")]
+use serde::{Serialize, Deserialize};
 
 /// This struct is an async version of [`std::path::PathBuf`].
 ///
 /// [`std::path::Path`]: https://doc.rust-lang.org/std/path/struct.PathBuf.html
+#[cfg_attr(feature = "serde-support", derive(Serialize, Deserialize))]
 #[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct PathBuf {
+    #[cfg_attr(feature = "serde-support", serde(flatten))]
     inner: std::path::PathBuf,
 }
 


yoshuawuyts:
@jamesmunns would it be possible to use [`serde-derive`](https://docs.rs/crate/serde_derive/1.0.102) here instead so that we can call the feature `serde`?

jamesmunns:
@yoshuawuyts I think so! It's a habit of mine to just use `serde` as a dependency, but it's likely not needed here, just `serde-derive`.

jamesmunns:
Things that need to be decided now:

1. What should the feature name be? just `serde` if we can depend on `serde-derive`? Something else?
2. Should this be behind the `unstable` feature? e.g. require `serde-support` AND `unstable` be set for now?
3. Are there any cases where a simple `flatten` won't work? Maybe wrapper types with more than one field?

Once we decide and correct those, we can probably land this with just PathBuf supported, then start incrementally covering more data types. Or we can hit them all at once in this PR.

yoshuawuyts:
Another thing we may want to consider is adding an entry to our "features" section in `lib.rs`: https://docs.rs/async-std/1.0.0/async_std/#features. This seems like something people may want to be aware of.

Cargo.toml
```diff
@@ -33,10 +33,14 @@ default = [
...
   "num_cpus",
   "pin-project-lite",
+  "serde-support",
 ]
 docs = ["attributes", "unstable"]
 unstable = ["default", "broadcaster"]
 attributes = ["async-attributes"]
+serde-support = [
+  "serde",
+]
```
yoshuawuyts: I think this feature might need to depend on `"default"`, as the `path` submodule not available on `"std"`. That way people don't run into any surprises.

jamesmunns:
@yoshuawuyts ah, no, I think we need a `serde` dependency as well. Otherwise I get:

```
    Checking async-std v1.0.0 (/tmp/async-std)
error[E0463]: can't find crate for `serde`
  --> /tmp/async-std/src/path/pathbuf.rs:22:49
   |
22 | #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
   |                                                 ^^^^^^^^^^^ can't find crate
```

Cargo.toml
jamesmunns: Added!

I think this could be relaxed, if there are serializable types that are present in non-`std` configurations. That should be possible to do down the road though.

stjepang:
Should we put serde support behind a new feature flag, perhaps named `serde1`?

See also: https://github.com/rust-lang/api-guidelines/issues/180

jamesmunns:
Hmm, this needs a deeper look. My first approach won't work.

```rust
use async_std;
use serde_json::to_string;
use serde::{self, Serialize, Deserialize};

#[derive(Serialize, Deserialize)]
struct Hmmm {
    ex: async_std::path::PathBuf,
}

fn main() {
    dbg!(to_string(&Hmmm {
        ex: async_std::path::PathBuf::new(),
    }).unwrap());
}
```

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("can only flatten structs and maps (got a string)", line: 0, column: 0)', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

There's likely still a way, just probably a bit less elegant. Merits further investigation.

jamesmunns:
But it does work for newtypes! https://github.com/async-rs/async-std/pull/512/commits/c70f00e915053c9dc43ef10cbe72fe3fc7a3a990 switches `PathBuf` to be a newtype, rather than a struct containing an `inner` field.

```
[src/main.rs:11] to_string(&Hmmm{ex: async_std::path::PathBuf::new(),}).unwrap() = "{\"ex\":\"\"}"
```

I don't think this needs to be a breaking change? And for fields where a newtype doesn't work, we can always manually impl `Serialize`/`Deserialize`

yoshuawuyts:
> I don't think this needs to be a breaking change?

We've never provided guarantees about internals, so indeed don't think it's a breaking change.

jamesmunns:
@stjepang just let me know if you'd prefer `serde` or `serde1` as the feature flag, then I think we can land this, and start adding serde support to other structures.

stjepang:
@dtolnay Do you think we should name the feature flag `serde` or `serde1`?

yoshuawuyts:
> I am skeptical of making this a default feature. (...)

Agreed; this should only be available when the `serde` feature is enabled. We currently don't depend on `serde` anywhere else, so it's indeed a new dep.

> I am skeptical of using serde_derive. (...)

Oh, yeah a manual impl def sounds better here. Thanks for suggesting!

maisiliymBridj:
Status?
Is anyone working on a PR to convert all wrapper types to newtypes, with serde feature?

hoijui:
How is the state of this?
As I understood from the conversation here, it is ready to be merged.. no?

poly2it:
It would be great to have this merged. I am also in need of this feature for PathBuf.

PR #514, opened 2019-11-12
stabilize task::yield_now

yoshuawuyts:
Stabilize `task::yield_now`. It doesn't perform any io, and doesn't rely on any reactor implementations either, so it can be stabilized as part of the `std` feature. Thanks!

### DIFF
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -124,6 +124,9 @@ cfg_std! {
 
     #[doc(inline)]
     pub use async_macros::ready;
+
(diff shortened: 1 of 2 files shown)


skade:
Can we ensure that we still hold that behind a feature flag so that it doesn't accidentally land in a patch release?

stjepang:
@skade Any reason why we shouldn't publish 1.1.0 as the next release?

skade:
@stjepang Mainly as an exercise, we might have situations where we do patch releases in between.

PR #520, opened 2019-11-12
Cleaning up stream pinning.

gierlachg:
Seems like ~none~ most of stream pinning (`pin_utils::pin_mut!(stream);`) is not required - removed macro usages where necessary


skade:
Thank you for the PR!

PR #523, opened 2019-11-12
polish lib.rs examples

### LINKED ISSUE #522: Extend examples to show more than just a single `task::block_on`
Can somebody add an example of what this would look like:

```javascript
const foo = async function () { return 1 }
const bar = async function () { return 1 }
const results = await Promise.all([foo(), bar()])
```

I think having just a single `task::block_on` in each example isn't very "real world"

yoshuawuyts:
This updates our lib.rs examples, showing off more of what async-std can do, and somewhat simplifying what we were already doing. Additionally it should make it nicer for people from other languages to understand what async rust is about.

Closes #522. Thanks!

## Screenshots

2019-11-12-230749_1920x1080

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,10 @@ syntax.
 
 ## Examples
 
+All examples require the [`"attributes"` feature] to be enabled.  This feature
+is not enabled by default because it significantly impacts compile times. See
+[`task::block_on`] for an alternative way to start executing tasks.
+
 ```rust
 async fn say_hello() {
     println!("Hello, world!");
@@ -90,6 +94,8 @@ More examples, including networking and file access, can be found in our
 
 [`examples`]: https://github.com/async-rs/async-std/tree/master/examples
 [documentation]: https://docs.rs/async-std#examples
+[`task::block_on`]: task/fn.block_on.html
+[`"attributes"` feature]: https://docs.rs/async-std/#features
 
 ## Philosophy
 
--- a/README.md
+++ b/README.md
@@ -75,19 +75,21 @@ syntax.
 ## Examples
 
 ```rust
-use async_std::task;
+async fn say_hello() {
+    println!("Hello, world!");
+}
 
-fn main() {
-    task::block_on(async {
-        println!("Hello, world!");
-    })
+#[async_std::main]
+async fn main() {
+    say_hello().await;
 }
 ```
 
 More examples, including networking and file access, can be found in our
-[`examples`] directory.
+[`examples`] directory and in our [documentation].
 
 [`examples`]: https://github.com/async-rs/async-std/tree/master/examples
+[documentation]: https://docs.rs/async-std#examples
 
 ## Philosophy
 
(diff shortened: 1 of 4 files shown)


brandonros:
I don't know if this is sufficient to close the issue.

```javascript
const main = async function () {
  const rows = await queryDatabase()
  const apiResponse = await fetchJson()
}
```

I think an example that showcases multiple uses of `spawn`ing tasks into something that `main` can then block on is much more realistic.

README.md
```diff
@@ -75,19 +75,21 @@ syntax.
...
-use async_std::task;
+async fn say_hello() {
+    println!("Hello, world!");
+}
 
-fn main() {
-    task::block_on(async {
-        println!("Hello, world!");
-    })
+#[async_std::main]
```
k-nasa: I thought it would be nice to show that the README needs to enable the 'attributes' feature as well as the documentation.

src/lib.rs
```diff
@@ -131,17 +131,55 @@
 //!
 //! # Examples
 //!
-//! Spawn a task and block the current thread on its result:
+//! All examples require the [`"attributes"` feature](#features) to be enabled.
```
k-nasa: 👍

src/lib.rs
```diff
@@ -131,17 +131,55 @@
...
+//! }
+//! ```
+//!
+//! Await two futures concurrently, and return a tuple of their output:
+//!
+//! ```
+//! #[async_std::main]
+//! async fn main() {
+//!     let a = || async move { 1u8 };
+//!     let b = || async move { 2u8 };
```
stjepang: I believe those two should not be closures?
```rust
let a = async { 1u8 };
let b = async { 2u8 };
```

omac777:
Here is an example that shows passing 3 arguments within a task::block_on, 2 string arguments along with a vector of strings.
It compiles/runs successfully.
```
use std::env::args;
use async_std::fs::File;
use async_std::io::{self, BufReader};
use async_std::prelude::*;
use async_std::task;

async fn say_hi( mypath1_ : &str, mypath2_ : &str, myFilePaths_ : &mut Vec<String>) -> io::Result<()> {
   
    task::block_on(async {
        println!("mypath1_:<<{}>>",mypath1_);
        println!("mypath2_:<<{}>>",mypath2_);
        for oneFilePath2 in myFilePaths_ {
            println!("oneFilePath2:<<{}>>", oneFilePath2);
        }

        Ok(())
    })    
}

fn main() -> io::Result<()> {
    let path1 = args().nth(1).expect("missing path argument");
    let path2 = args().nth(2).expect("missing path argument");

    let mut vsMyFilePaths : Vec<String> = Vec::new();
    vsMyFilePaths.push("blah1".to_string());
    vsMyFilePaths.push("blah2".to_string());
    vsMyFilePaths.push("blah3".to_string());
    
    task::block_on(say_hi(&path1, &path2, &mut vsMyFilePaths))
}
```

PR #525, opened 2019-11-13
Add io write by_ref

yjhmelody:


yoshuawuyts:
I tried doing this as part of #314 a while ago. I think the right way to go about this is to add the methods to the futures-rs implementations so they can be directly targeted. I think this would indeed be a breaking change if we merged it here directly.

PR #526, opened 2019-11-13
refactor io dir to be same with std and export IntoInnerError

yjhmelody:


yjhmelody:
std export the `IntoInnerError`, so do we.
But I'm stranged that when I change to use std's `IntoInnerError`, the compiler reported errors. Though they are the same structure.

src/io/mod.rs
```diff
@@ -269,13 +269,15 @@
 //! [`Result`]: https://doc.rust-lang.org/std/result/enum.Result.html
 //! [`.unwrap()`]: https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap
 
+const DEFAULT_BUF_SIZE: usize = 8 * 1024;
```
yoshuawuyts: should this const be marked `pub(crate)`? Is that even possible?

I guess I'm a bit confused how you're accessing this from other crates without it having a visibility specifier haha. Feel like I'm missing something about how consts work :P

yjhmelody: Well, I think sub mod always can access father mod.

yjhmelody: Just like write `tests` in sub mod even for test some private functions.
I think pub(crate) is for being public to current crate and private to extern crate.

yjhmelody: See `std::io`,  I just imitate its way of writing.
image

PR #527, opened 2019-11-13
Add Stream count

yjhmelody:
#129


k-nasa:
This apparently does the same thing as #368 .
I want to give priority to the one submitted first, but is there any objection?

yoshuawuyts:
@k-nasa no objection; both PRs are really good, but sadly we can only merge one.

PR #530, opened 2019-11-13
add mutex benches vs std + parking_lot

yoshuawuyts:
Entirely for fun, but sets us up against `std` and `parking_lot` mutexes, which will soon be the same thing. Figured it'd be nice to remind us of the baseline we're measuring against while benchmarking. Because it doesn't seem like perf will fluctuate much there.

Thanks!

## Sample Output

```txt
test async_std::contention      ... bench:     342,769 ns/iter (+/- 59,938)
test async_std::create          ... bench:           5 ns/iter (+/- 0)
test async_std::no_contention   ... bench:     576,119 ns/iter (+/- 9,641)
test parking_lot::contention    ... bench:     257,292 ns/iter (+/- 66,530)
test parking_lot::create        ... bench:           0 ns/iter (+/- 0)
test parking_lot::no_contention ... bench:     342,655 ns/iter (+/- 23,643)
test std::contention            ... bench:   1,560,041 ns/iter (+/- 733,017)
test std::create                ... bench:          24 ns/iter (+/- 1)
test std::no_contention         ... bench:     339,644 ns/iter (+/- 39,173)
test result: ok. 0 passed; 0 failed; 0 ignored; 9 measured; 0 filtered out
```
_note that "no contention" is never faster than "contention" because we're running 10x the amount of tasks. They cannot quite be compared (:_

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ crossbeam-deque = { version = "0.7.1", optional = true }
 crossbeam-utils = { version = "0.6.6", optional = true }
 futures-core = { version = "0.3.0", optional = true }
 futures-io = { version = "0.3.0", optional = true }
-futures-timer = { version = "1.0.2", optional = true }
+futures-timer = { version = "2.0.0", optional = true }
 kv-log-macro = { version = "1.0.4", optional = true }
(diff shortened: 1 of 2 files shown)


c410-f3r:
Maybe also add `futures::lock::Mutex` to the list?

PR #534, opened 2019-11-14
Make module "channel" pub

### LINKED ISSUE #532: Cannot access the "Sender" of a channel
I am having problems passing a sender as a parameter to a function.

The following code:

```rust
use async_std::sync::channel;
use async_std::io;
use async_std::prelude::*;
use async_std::task;


fn main() -> io::Result<()> {
    task::block_on(async {
        let (s, r) = channel::<i32>(1);
        let webcam = start_service(&s);
        Ok(())
    })
}

async fn start_service(s : &async_std::sync::channel::Sender<i32>) -> () {
}
```

gives the following error:
```
error[E0603]: module `channel` is private
  --> src\main.rs:15:46
   |
15 | async fn start_service(s : &async_std::sync::channel::Sender) -> () {
   |                                              ^^^^^^^
```

Is it an error or am I using it incorrectly?

Thanks for an awesome crate!

mrpink76:
Tried to fix my own issue. Hope someone can review the fix!

Fixes #532

### DIFF
--- a/src/sync/mod.rs
+++ b/src/sync/mod.rs
@@ -187,7 +187,7 @@ cfg_unstable! {
     pub use channel::{channel, Sender, Receiver};
 
     mod barrier;
-    mod channel;
+    pub mod channel;
 }
 
 pub(crate) mod waker_set;


yoshuawuyts:
@mrpink76 Thanks for your PR; however this is already public as [`async_std::sync::Sender`](https://docs.rs/async-std/1.0.1/async_std/sync/struct.Sender.html)

PR #538, opened 2019-11-15
 Add stream by_ref

k-nasa:
ref: https://github.com/async-rs/async-std/issues/129

### DIFF
--- a/src/stream/stream/mod.rs
+++ b/src/stream/stream/mod.rs
@@ -1405,6 +1405,52 @@ extension_trait! {
         }
 
         #[doc = r#"
+            Borrows an stream, rather than consuming it.
+
+            This is useful to allow applying stream adaptors while still retaining ownership of the original stream.
+
+            # Examples
+
+            ```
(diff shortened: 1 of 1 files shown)


src/stream/stream/mod.rs
```diff
@@ -1404,6 +1404,52 @@ extension_trait! {
...
+
+            assert_eq!(sum, 3);
+
+            // now this is just fine:
+            assert_eq!(stream.next().await, Some(3));
+            assert_eq!(stream.next().await, None);
+            #
+            # }) }
+            ```
+        "#]
```
yoshuawuyts: Could we maybe mark this as "unstable"?

I'm quite sure this is fine, but just to be sure *all* new APIs introduced are marked unstable first, so we can stabilize them later. Thanks!

k-nasa: Sorry! I forget it always...

yoshuawuyts: @k-nasa no worries! Thanks for fixing!

Issue #539, opened 2019-11-15
To implement ToSocketAddrs for a custom type the ToSocketAddrsFuture should be pub use


bandheight:

yoshuawuyts:
Ref https://docs.rs/async-std/1.0.1/async_std/net/trait.ToSocketAddrs.html. From a glance it seems this is indeed accurate. Since the result type from `ToSocketAddrs` is opaque, we could probably change it to be similar to [`FromStream::from_stream`](https://docs.rs/async-std/1.0.1/async_std/stream/trait.FromStream.html#tymethod.from_stream) without breaking anything.

@bandheight out of curiosity: what's your use case for implementing `ToSocketAddrs`?

stjepang:
I'd be okay with changing the return type of `to_socket_addrs` to `Pin<Box<dyn Future>>` here. Performance impact of the change would be effectively none.

bandheight:
@yoshuawuyts I implemented an `enum` type to represent multiple address types.

```rust
pub enum Address {
    /// IP Address
    SocketAddr(SocketAddr),
    /// Domain name address, eg. example.com:8080
    Domain(String, u16),
}
```
If this type implement the trait `ToSocketAddrs`,  it will be passed directly to `TcpListener::bind` and other similar functions.

yoshuawuyts:
@bandheight does the following work for you?

```rust
async fn bind_socket(addr: Address) -> io::Result<TcpListener> {
    match addr {
        SocketAddr(addr) => TcpListener::bind(addr).await?,
        SocketAddr((addr, port) => TcpListener::bind((&*addr, port)).await?,
    }
}
```

If we don't already support `ToSocketAddr` for `(String, u16)` then we should add that!

bandheight:
@yoshuawuyts This code is worked. 
And it is also my current solution, but I want to make my code look more neat and elegant.
So implement the trait `ToSocketAddrs` is probably the best solution.

```rust
pub struct DomainName(pub String, pub u16);

...
// implement Deserialize for DomainName
...

#[derive(Deserialize)]
pub enum Address {
    /// IP Address
    SocketAddr(SocketAddr),
    /// Domain name address, eg. example.com:8080
    Domain(DomainName),
}

#[derive(Deserialize)]
pub struct Config {
    pub addr: Address,
}

...
let config: Config = <deserialize from config file>;
...

// This is neat and elegant.
let listener = TcpListener::bind(config.addr);
```

[...]

PR #540, opened 2019-11-15
Make ToSocketAddrsFuture available for pub use.

bandheight:
If we want to implement ToSocketAddrs for a custom type,
the ToSocketAddrsFuture should be pub use.

Signed-off-by: Hosun Zhu <hosun@linux.com>


yoshuawuyts:
Ref #539

PR #542, opened 2019-11-15
update doc url

yjhmelody:
Maybe it's better to use latest version url for futures lib?
Just need to replace `0.3` with `latest` in url


src/io/seek/mod.rs
```diff
@@ -27,7 +27,7 @@ extension_trait! {
 
         [`std::io::Seek`]: https://doc.rust-lang.org/std/io/trait.Seek.html
         [`futures::io::AsyncSeek`]:
-        https://docs.rs/futures-preview/0.3.0-alpha.17/futures/io/trait.AsyncSeek.html
+        https://docs.rs/futures/0.3/futures/stream/trait.Stream.html
```
taiki-e:
```suggestion
        https://docs.rs/futures/0.3/futures/io/trait.AsyncSeek.html
```

PR #544, opened 2019-11-15
update futures-timer & other deps

yoshuawuyts:
Updates futures-timer to `2.x`, removing some unneeded deps which should improve compile times. Also for good measure ran `cargo upgrade` to upgrade all other deps. Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,33 +50,33 @@ std = [
 ]
 
 [dependencies]
-async-attributes = { version = "1.1.0", optional = true }
+async-attributes = { version = "1.1.1", optional = true }
 async-macros = { version = "2.0.0", optional = true }
 async-task = { version = "1.0.0", optional = true }
 broadcaster = { version = "0.2.6", optional = true, default-features = false, features = ["default-channels"] }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -50,33 +50,33 @@ std = [
...
 broadcaster = { version = "0.2.6", optional = true, default-features = false, features = ["default-channels"] }
-crossbeam-channel = { version = "0.3.9", optional = true }
-crossbeam-deque = { version = "0.7.1", optional = true }
-crossbeam-utils = { version = "0.6.6", optional = true }
-futures-core = { version = "0.3.0", optional = true }
-futures-io = { version = "0.3.0", optional = true }
-futures-timer = { version = "1.0.2", optional = true }
+crossbeam-channel = { version = "0.4.0", optional = true }
+crossbeam-deque = { version = "0.7.2", optional = true }
+crossbeam-utils = { version = "0.7.0", optional = true }
```
taiki-e: Filed https://github.com/async-rs/async-task/pull/7 to avoid duplicate dependencies.

PR #546, opened 2019-11-15
Fix clippy warn

k-nasa:


yoshuawuyts:
This patch indeed seems to fix the clippy issues; but @stjepang and I have been talking over the past week, and a lot of the clippy lints are starting to feel a bit *tedious*. In addition to that things seem to be breaking fairly regularly on the clippy side, which made us question if we should even keep it.

@k-nasa how do you feel about clippy?

k-nasa:
umm.  This problem is difficult. 
Certainly working with code conventions is tedious. 
The problems I have fixed now are most that don't have to be fixed. 
 However, clippy sometimes displays important warnings. 

For example this: https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone

So I think it is difficult to ignore it completely.
In addition, it is cumbersome to choose which items to fix and which to ignore.

I don't have conclution for this problem.
but, I think someone needs to face clippy and make the appropriate corrections.

I'm interested in what you and @stjepang  think.

PR #548, opened 2019-11-16
fix stream code style

yjhmelody:
I think the `pub` $vis in stream also are  style inconsisitent.
Because like `new` fn is only used by super module, `pub(crate)` / `pub(super)` / `pub` are all no different.

### DIFF
--- a/src/stream/stream/all.rs
+++ b/src/stream/stream/all.rs
@@ -14,6 +14,17 @@ pub struct AllFuture<'a, S, F, T> {
     pub(crate) _marker: PhantomData<T>,
 }
 
+impl<'a, S, F, T> AllFuture<'a, S, F, T> {
+    pub(crate) fn new(stream: &'a mut S, f: F) -> Self {
+        Self {
+            stream,
+            f,
+            result: true, // the default if the empty stream
+            _marker: PhantomData,
+        }
+    }
+}
+
 impl<S: Unpin, F, T> Unpin for AllFuture<'_, S, F, T> {}
 
 impl<S, F> Future for AllFuture<'_, S, F, S::Item>
--- a/src/stream/stream/any.rs
+++ b/src/stream/stream/any.rs
@@ -14,6 +14,17 @@ pub struct AnyFuture<'a, S, F, T> {
     pub(crate) _marker: PhantomData<T>,
 }
 
+impl<'a, S, F, T> AnyFuture<'a, S, F, T> {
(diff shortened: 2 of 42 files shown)


src/stream/stream/all.rs
```diff
@@ -14,6 +14,17 @@ pub struct AllFuture<'a, S, F, T> {
     pub(crate) _marker: PhantomData<T>,
 }
 
+impl<'a, S, F, T> AllFuture<'a, S, F, T> {
+    pub fn new(stream: &'a mut S, f: F) -> Self {
```
taiki-e:
```suggestion
    pub(crate) fn new(stream: &'a mut S, f: F) -> Self {
```

I don't think this is public at this time, but [public methods can avoid the dead_code warning](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e613e0113ead88b7038308455ff9523a), so I think it is preferable to make it private (avoid to compile unused private API).

src/stream/stream/any.rs
```diff
@@ -14,6 +14,17 @@ pub struct AnyFuture<'a, S, F, T> {
     pub(crate) _marker: PhantomData<T>,
 }
 
+impl<'a, S, F, T> AnyFuture<'a, S, F, T> {
+    pub fn new(stream: &'a mut S, f: F) -> Self {
```
taiki-e:
```suggestion
    pub(crate) fn new(stream: &'a mut S, f: F) -> Self {
```

src/stream/stream/count.rs
```diff
@@ -19,8 +19,8 @@ pin_project! {
 }
 
 impl<S> CountFuture<S> {
-    pub(crate) fn new(stream: S) -> Self {
-        CountFuture { stream, count: 0 }
+    pub fn new(stream: S) -> Self {
```
taiki-e:
```suggestion
    pub(crate) fn new(stream: S) -> Self {
```

src/stream/stream/cycle.rs
```diff
@@ -15,8 +15,8 @@ impl<S> Cycle<S>
 where
     S: Stream + Clone,
 {
-    pub fn new(source: S) -> Cycle<S> {
-        Cycle {
+    pub fn new(source: S) -> Self {
```
taiki-e:
```suggestion
    pub(crate) fn new(source: S) -> Self {
```

yjhmelody:
@taiki-e  why use `pub(crate)`?

PR #551, opened 2019-11-17
verbose errors feature

killercup:
This adds a new "verbose-errors" feature flag to async-std that enables
wrapping certain errors in structures with more context. As an example,
we use it in `fs::File::{open,create}` to add the given path to the
error message (something that is lacking in std to annoyance of many).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ std = [
   "pin-utils",
   "slab",
 ]
+verbose-errors = []
 
 [dependencies]
 async-attributes = { version = "1.1.1", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,6 @@ std = [
   "pin-utils",
   "slab",
 ]
-verbose-errors = []
 
 [dependencies]
 async-attributes = { version = "1.1.1", optional = true }
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -53,17 +53,8 @@ pub fn random(n: u32) -> u32 {
 }
 
 /// Add additional context to errors
-///
-/// *Note for implementors:* The given closure must only be executed when
-/// `verbose-errors` feature is enabled for this crate!
-pub(crate) trait VerboseErrorExt: Sized {
-    #[cfg(feature = "verbose-errors")]
-    fn verbose_context(self, message: impl Fn() -> String) -> Self;
-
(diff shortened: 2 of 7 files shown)


killercup:
We discussed this a while back, and I wanted to provide something small that can be a basis for discussion. The main thing here is to provide an internal mechanism to easily add optional contexts. I hope–but have not proven–that the closure get compiled away in release modes without the feature flag set.

stjepang:
This is great, thank you! :) I think we don't even need to introduce a new feature flag for this and can just provide verbose errors by default. Is there any real downside to providing this by default?

Wdyt, @yoshuawuyts ?

killercup:
It costs a string allocation in the error case :shrug:

stjepang:
Error cases don't need to be fast :)

Besides, we already have a string allocation for the `path`, a task allocation for `spawn`, a push into the queue, and so on... so the cost of this additional error context is practically zero.

src/utils.rs
```diff
@@ -52,6 +52,20 @@ pub fn random(n: u32) -> u32 {
     })
 }
 
+/// Add additional context to errors
+///
+/// *Note for implementors:* The given closure must only be executed when
+/// `verbose-errors` feature is enabled for this crate!
+pub(crate) trait VerboseErrorExt: Sized {
```
yoshuawuyts: I think we could follow `anyhow` / `fehler`'s lead here and call it `Context`, with a method named `context`.

Cargo.toml
```diff
@@ -48,6 +48,7 @@ std = [
   "pin-utils",
   "slab",
 ]
+verbose-errors = []
```
yoshuawuyts: Agree we don't need a feature for this (:

src/utils.rs
killercup: Agreed now that it's no longer the name of the feature flag :)

tests/verbose_errors.rs
```diff
@@ -0,0 +1,20 @@
+#[cfg(feature = "verbose-errors")]
```
otavio: The flag was removed, wasn't it?

killercup: A sorry, forgot this one. Fixed.

killercup: A sorry, forgot this one. Fixed.

killercup:
🎉 let's maybe open an issue to track more places we can add this

PR #552, opened 2019-11-18
add os::windows::symlink_{dir,file}

hhggit:


src/os/windows/mod.rs
```diff
@@ -2,4 +2,5 @@
 
 cfg_std! {
     pub mod io;
+    pub mod fs;
```
yoshuawuyts: This should probably part of a `cfg_default! {}` block so it's only included when the full runtime is enabled. Our "std" feature means "traits and types only, but no threadpool, executor, or other bindings". I suspect this is what's failing CI right now.

yoshuawuyts:
Oh, one last thing. We should probably mark these APIs as "unstable". But I can do that in a follow-up PR. Thanks heaps!

PR #562, opened 2019-11-20
DoubleEndedStream extension

felipesere:
This is a first stab at a skeleton implementation of the `DoubleEndedStream` extension.
Docs are mostly still outstanding, I'll add them in based on their Stream variation as this evolves.

I also added a dummy "Sample" implementation of `DoubleEndedStream` for a vec.
While not the worst, this probably needs a better name and place,
possibly even `double_ended_stream::from_iter(...)`?

Looking forward for any feedback and advice :) 

#129


src/stream/double_ended/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+
+            let mut s = Sample::from(vec![7u8]);
+
+            assert_eq!(s.next().await, Some(7));
+            assert_eq!(s.next().await, None);
+            #
+            # }) }
+            ```
+        "#]
+        fn next(&mut self) -> impl Future<Output = Option<Self::Item>> + '_ [NextBackFuture<'_, Self>]
```
yoshuawuyts:
```suggestion
        fn next_back(&mut self) -> impl Future<Output = Option<Self::Item>> + '_ [NextBackFuture<'_, Self>]
```

src/stream/double_ended/nth_back.rs
```diff
@@ -0,0 +1,39 @@
...
+            Some(v) => match self.n {
+                0 => Poll::Ready(Some(v)),
+                _ => {
+                    self.n -= 1;
+                    cx.waker().wake_by_ref();
+                    Poll::Pending
+                }
+            },
+            None => Poll::Ready(None),
+        }
```
yoshuawuyts: This block can probably be flattened into:

```rust
match next {
    Some(0) => Poll::Ready(Some(v)),
    Some(_) => {
        self.n -= 1;
        cx.waker().wake_by_ref();
        Poll::Pending
    },
    None => Poll::Ready(None),
}
```

src/stream/double_ended/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+
+extension_trait! {
+    use crate::stream::Stream;
+
+    use std::pin::Pin;
+    use std::task::{Context, Poll};
+
+
+    #[doc = r#"
+    Something fancy
```
k-nasa:
```suggestion
            Something fancy
```
Although it is a detailed place

src/stream/double_ended/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+
+            # Examples
+
+            ```
+            # fn main() { async_std::task::block_on(async {
+            #
+            use async_std::stream::Sample;
+            use async_std::stream::double_ended::DoubleEndedStreamExt;
+
+            let mut s = Sample::from(vec![7u8]);
```
k-nasa: Isn't it better to implement `DoubleEndedStream` in` FromIter`?
Was there any problem?

src/stream/double_ended/nth_back.rs
felipesere: Almost :) next can not be matched against `Some(0)`, as the type is ` Option<S::Item>`.
The Some(0) comes from matching against self.n

src/stream/double_ended/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+
+            # Examples
+
+            Basic usage:
+
+            ```
+            # fn main() { async_std::task::block_on(async {
+            #
+            use async_std::stream::Sample;
+            use async_std::stream::double_ended::DoubleEndedStreamExt;
```
yjhmelody: whether should we re-export it to `async_std::stream::DoubleEndedStreamExt` rather than `async_std::stream::double_ended::DoubleEndedStreamExt`?

felipesere: I still need to figure this out, but its good input :)

felipesere:
I'm kind of stuck with figuring out how and where to reexport different chunks of the API.
The first thing is that I am not 100% sure what the import structure should look like and then how to map it to files/modles/uses.

Could someone nudge me in the right direction?

yoshuawuyts:
@felipesere could you maybe share more details where you're getting stuck? What are you trying to do that isn't working? This is a rather large PR, and I don't have the information quite ready the way you do I think :sweat_smile:

yoshuawuyts:
Oops, I accidentally pushed changes onto this branch, then reset the changes with a force push. It should be all good now. If anyone's wondering what had happened; sorry about that!.

src/stream/double_ended_stream/from_iter.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+        return Poll::Ready(None);
+    }
+}
+
+impl<T> DoubleEndedStream for FromIter<T> {
+    fn poll_next_back(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        Poll::Ready(self.inner.pop())
+    }
+}
```
yoshuawuyts: I'm a bit confused about this file; can't [`stream::FromIter`](https://docs.rs/async-std/1.2.0/async_std/stream/struct.FromIter.html) implement `DoubleEndedStream` if `T: DoubleEndedIterator`? I don't think this function would still be needed then, right?

PR #568, opened 2019-11-20
Fix rng use in Stream::merge

stjepang:
`random(n)` generates a random number in `0..n`, which means `random(1)` always returns 0. :)

This PR changes `random(1)` to `random(2)`.


yoshuawuyts:
dang, I should've caught that. Oops, sorry!

PR #570, opened 2019-11-20
Add ring benchmark example

stjepang:
This is the code from https://github.com/stjepang/ring-benchmark

The list of examples is getting a bit unwieldy. I wonder if it would make sense to move the list into the book and just link to it from here. After all, the growing list of  examples is slowly becoming a cookbook already.

@skade, any ideas on how to integrate these examples with the book?


stjepang:
I'll try turning this into a microbenchmark. :)

Issue #569, opened 2019-11-20
Make more errors verbose


killercup:
In #551 I added the ability to easily add details to `std::io::Error`s while not changing the type. This can make many error cases much more user friendly:

https://github.com/async-rs/async-std/blob/b3d30de4a11d785929a086e458ec1da4aec34bf3/src/fs/file.rs#L117-L118

I only added this in two places so far, though, `File::open` and `File::create`. Let's use this issue to track more places we should do this!

gilescope:
Would love that std's errors could be just as helpful! Great work.

killercup:
Thanks! That also reminds me: this issue can probably be closed by now. Feel free to enhance more errors of course!

PR #571, opened 2019-11-20
Add context to more errors

killercup:
Pretty mechanical additions and by no means exhaustive: I bascially just wanted to cover the APIs I end up using quite often :)

For a future PR it might be interesting to add a test that shows that our errors include the correct sources, and that they might even work well with std backtraces.

cc #569

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -9,11 +9,11 @@ use std::sync::{Arc, Mutex};
 
 use crate::fs::{Metadata, Permissions};
 use crate::future;
-use crate::utils::Context as _;
 use crate::io::{self, Read, Seek, SeekFrom, Write};
 use crate::path::Path;
 use crate::prelude::*;
 use crate::task::{self, spawn_blocking, Context, Poll, Waker};
+use crate::utils::Context as _;
 
 /// An open file on the filesystem.
 ///
@@ -114,8 +114,7 @@ impl File {
     pub async fn open<P: AsRef<Path>>(path: P) -> io::Result<File> {
         let path = path.as_ref().to_owned();
         let file = spawn_blocking(move || {
-            std::fs::File::open(&path)
-                .context(|| format!("Could not open {}", path.display()))
+            std::fs::File::open(&path).context(|| format!("Could not open `{}`", path.display()))
         })
         .await?;
         Ok(File::new(file, true))
@@ -154,7 +153,7 @@ impl File {
         let path = path.as_ref().to_owned();
         let file = spawn_blocking(move || {
             std::fs::File::create(&path)
-                .context(|| format!("Could not create {}", path.display()))
+                .context(|| format!("Could not create `{}`", path.display()))
         })
         .await?;
         Ok(File::new(file, true))
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -114,7 +114,7 @@ impl File {
     pub async fn open<P: AsRef<Path>>(path: P) -> io::Result<File> {
         let path = path.as_ref().to_owned();
         let file = spawn_blocking(move || {
-            std::fs::File::open(&path).context(|| format!("Could not open `{}`", path.display()))
+            std::fs::File::open(&path).context(|| format!("could not open `{}`", path.display()))
         })
         .await?;
         Ok(File::new(file, true))
@@ -153,7 +153,7 @@ impl File {
         let path = path.as_ref().to_owned();
         let file = spawn_blocking(move || {
             std::fs::File::create(&path)
-                .context(|| format!("Could not create `{}`", path.display()))
+                .context(|| format!("could not create `{}`", path.display()))
         })
         .await?;
         Ok(File::new(file, true))
--- a/src/io/stdin.rs
+++ b/src/io/stdin.rs
@@ -1,10 +1,11 @@
+use std::future::Future;
 use std::pin::Pin;
 use std::sync::Mutex;
-use std::future::Future;
 
 use crate::future;
 use crate::io::{self, Read};
 use crate::task::{spawn_blocking, Context, JoinHandle, Poll};
+use crate::utils::Context as _;
 
 cfg_unstable! {
     use once_cell::sync::Lazy;
@@ -162,6 +163,7 @@ impl Stdin {
             }
         })
         .await
+        .context(|| String::from("Could not read line on stdin"))
     }
 
     /// Locks this handle to the standard input stream, returning a readable guard.
--- a/src/io/stdin.rs
+++ b/src/io/stdin.rs
@@ -163,7 +163,7 @@ impl Stdin {
             }
         })
         .await
-        .context(|| String::from("Could not read line on stdin"))
+        .context(|| String::from("could not read line on stdin"))
     }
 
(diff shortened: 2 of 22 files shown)


src/fs/file.rs
```diff
@@ -114,8 +114,7 @@ impl File {
     pub async fn open<P: AsRef<Path>>(path: P) -> io::Result<File> {
         let path = path.as_ref().to_owned();
         let file = spawn_blocking(move || {
-            std::fs::File::open(&path)
-                .context(|| format!("Could not open {}", path.display()))
+            std::fs::File::open(&path).context(|| format!("Could not open `{}`", path.display()))
```
yoshuawuyts:
```suggestion
            std::fs::File::open(&path).context(|| format!("could not open `{}`", path.display()))
```

src/fs/file.rs
```diff
@@ -154,7 +153,7 @@ impl File {
         let path = path.as_ref().to_owned();
         let file = spawn_blocking(move || {
             std::fs::File::create(&path)
-                .context(|| format!("Could not create {}", path.display()))
+                .context(|| format!("Could not create `{}`", path.display()))
```
yoshuawuyts:
```suggestion
                .context(|| format!("could not create `{}`", path.display()))
```

src/io/stdin.rs
```diff
@@ -162,6 +163,7 @@ impl Stdin {
             }
         })
         .await
+        .context(|| String::from("Could not read line on stdin"))
```
yoshuawuyts:
```suggestion
        .context(|| String::from("could not read line on stdin"))
```

src/net/tcp/listener.rs
```diff
@@ -75,8 +76,12 @@ impl TcpListener {
     /// [`local_addr`]: #method.local_addr
     pub async fn bind<A: ToSocketAddrs>(addrs: A) -> io::Result<TcpListener> {
         let mut last_err = None;
+        let addrs = addrs
+            .to_socket_addrs()
+            .await
+            .context(|| String::from("could not resolve addresses"))?;
```
yoshuawuyts: It'd be great to know _which_ addresses couldn't be resolved. I feel this is one of the most common cases folks are left wondering about.

src/net/tcp/stream.rs
```diff
@@ -71,11 +72,17 @@ impl TcpStream {
     /// ```
     pub async fn connect<A: ToSocketAddrs>(addrs: A) -> io::Result<TcpStream> {
         let mut last_err = None;
+        let addrs = addrs
+            .to_socket_addrs()
+            .await
+            .context(|| String::from("could not resolve addresses"))?;
```
yoshuawuyts: It'd be great to know _which_ addresses couldn't be resolved. I feel this is one of the most common cases folks are left wondering about.

src/net/udp/mod.rs
```diff
@@ -66,10 +67,14 @@ impl UdpSocket {
     /// #
     /// # Ok(()) }) }
     /// ```
-    pub async fn bind<A: ToSocketAddrs>(addr: A) -> io::Result<UdpSocket> {
+    pub async fn bind<A: ToSocketAddrs>(addrs: A) -> io::Result<UdpSocket> {
         let mut last_err = None;
+        let addrs = addrs
+            .to_socket_addrs()
+            .await
+            .context(|| String::from("could not resolve addresses"))?;
```
yoshuawuyts: It'd be great to know _which_ addresses couldn't be resolved. I feel this is one of the most common cases folks are left wondering about.

src/net/udp/mod.rs
```diff
@@ -151,6 +159,7 @@ impl UdpSocket {
                 .poll_write_with(cx, |inner| inner.send_to(buf, &addr))
         })
         .await
+        .context(|| format!("Could not send packet to {}", addr))
```
yoshuawuyts:
```suggestion
        .context(|| format!("could not send packet to {}", addr))
```

src/net/udp/mod.rs
```diff
@@ -178,6 +187,17 @@ impl UdpSocket {
                 .poll_read_with(cx, |inner| inner.recv_from(buf))
         })
         .await
+        .context(|| {
+            use std::fmt::Write;
+
+            let mut error = String::from("Could not receive data on ");
```
yoshuawuyts:
```suggestion
            let mut error = String::from("could not receive data on ");
```

src/net/udp/mod.rs
```diff
@@ -204,8 +224,12 @@ impl UdpSocket {
     /// ```
     pub async fn connect<A: ToSocketAddrs>(&self, addrs: A) -> io::Result<()> {
         let mut last_err = None;
+        let addrs = addrs
+            .to_socket_addrs()
+            .await
+            .context(|| String::from("could not resolve addresses"))?;
```
yoshuawuyts: It'd be great to know _which_ addresses couldn't be resolved. I feel this is one of the most common cases folks are left wondering about.

src/net/udp/mod.rs
```diff
@@ -248,7 +272,19 @@ impl UdpSocket {
     /// # Ok(()) }) }
     /// ```
     pub async fn send(&self, buf: &[u8]) -> io::Result<usize> {
-        future::poll_fn(|cx| self.watcher.poll_write_with(cx, |inner| inner.send(buf))).await
+        future::poll_fn(|cx| self.watcher.poll_write_with(cx, |inner| inner.send(buf)))
+            .await
+            .context(|| {
+                use std::fmt::Write;
+
+                let mut error = String::from("Could not send data on ");
```
yoshuawuyts:
```suggestion
                let mut error = String::from("could not send data on ");
```

src/net/udp/mod.rs
```diff
@@ -271,7 +307,19 @@ impl UdpSocket {
     /// # Ok(()) }) }
     /// ```
     pub async fn recv(&self, buf: &mut [u8]) -> io::Result<usize> {
-        future::poll_fn(|cx| self.watcher.poll_read_with(cx, |inner| inner.recv(buf))).await
+        future::poll_fn(|cx| self.watcher.poll_read_with(cx, |inner| inner.recv(buf)))
+            .await
+            .context(|| {
+                use std::fmt::Write;
+
+                let mut error = String::from("Could not receive data on ");
```
yoshuawuyts:
```suggestion
                let mut error = String::from("could not receive data on ");
```

src/net/tcp/listener.rs
killercup: ~~Yeah, that's what I wanted as well, but we sadly can't do that without adding more constraints on `A` which would be a breaking change~~

Wrong actually! Since is `async_std::net::ToSocketAddrs`, not the trait from std! If you want to add that info, we can actually add a method like `#[doc(hidden)] fn debug_print_addr<D: Debug>(&self) -> Option<T> { None }` without breaking API!

Or, actually even better: Move the `.context` call into `to_socket_addrs`!

That'd be something for a different PR, though :)

killercup: Actually, scratch all that. It was easier than expected and I added a new commit :)

PR #573, opened 2019-11-21
Fix a-chat tutorial issues

unleashed:
These fix a few typos and grammar issues as well as minor things in the code as presented, such as slightly wrong bounds, calling twice `trim`, or wrong or missing use clauses as the tutorial advances.

### DIFF
--- a/docs/src/tutorial/receiving_messages.md
+++ b/docs/src/tutorial/receiving_messages.md
@@ -46,7 +46,7 @@ async fn connection_loop(stream: TcpStream) -> Result<()> {
             Some(idx) => (&line[..idx], line[idx + 1 ..].trim()),
         };
         let dest: Vec<String> = dest.split(',').map(|name| name.trim().to_string()).collect();
-        let msg: String = msg.trim().to_string();
+        let msg: String = msg.to_string();
     }
     Ok(())
 }
--- a/docs/src/tutorial/receiving_messages.md
+++ b/docs/src/tutorial/receiving_messages.md
@@ -130,7 +130,7 @@ So let's use a helper function for this:
 # };
 fn spawn_and_log_error<F>(fut: F) -> task::JoinHandle<()>
 where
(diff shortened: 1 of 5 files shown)


docs/src/tutorial/receiving_messages.md
```diff
@@ -130,7 +134,7 @@ So let's use a helper function for this:
 # };
 fn spawn_and_log_error<F>(fut: F) -> task::JoinHandle<()>
 where
-    F: Future<Output = io::Result<()>> + Send + 'static,
+    F: Future<Output = Result<()>> + Send + 'static,
```
stjepang: Why this change, where does this `Result` type come from?

docs/src/tutorial/receiving_messages.md
```diff
@@ -46,7 +50,7 @@ async fn connection_loop(stream: TcpStream) -> Result<()> {
             Some(idx) => (&line[..idx], line[idx + 1 ..].trim()),
         };
         let dest: Vec<String> = dest.split(',').map(|name| name.trim().to_string()).collect();
-        let msg: String = msg.trim().to_string();
+        let msg: String = msg.to_string();
```
stjepang: Why no trimming?

docs/src/tutorial/receiving_messages.md
unleashed: From:

> type Result<T> = std::result::Result<T, Box<dyn std::error::Error + Send + Sync>>;

which was defined previously in the tutorial. The `F` bound on `Future<Output = io::Result<()>>` is a type and does not compile.

docs/src/tutorial/receiving_messages.md
unleashed: Because `msg` is trimmed already 4 lines above. I would have preferred the `trim()` to happen here for clarity, but that's how it is presented in other samples of the code down the road (ie. no `trim()` in the following sections).

PR #577, opened 2019-11-23
Refactor join type

k-nasa:
I think the `future :: join` type constraint is strong.
How about making it possible to `join` different future types?

### DIFF
--- a/src/future/future/join.rs
+++ b/src/future/future/join.rs
@@ -12,7 +12,7 @@ pin_project! {
     pub struct Join<L, R>
     where
         L: Future,
-        R: Future<Output = L::Output>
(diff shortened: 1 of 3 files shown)


stjepang:
I like this change. Can we perhaps also do the same for `try_join`?

killercup:
Doesn't this break inference in some cases? I think that's not technically considered a breaking change but if users can notice it, it'd be nice to get a heads up in release notes or the docs.

stjepang:
@killercup I'm afraid it can break inference, yes. These APIs are still unstable and it's not really a breaking change, so I think we're okay. It would be nice to add a heads up to the changelog, though.

PR #581, opened 2019-11-24
feat: implement Barrier using Condvar

### LINKED ISSUE #565: Rewrite Barrier using WakerSet
The `broadcaster` library has taken us very far, but unfortunately has dependencies on `futures_channel`. Now that we have `WakerSet` available internally (for now still, see #549) we should probably consider rewriting `Barrier` to use it.

It should both improve our compilation times, and performance. This would put us in a place where we could consider stabilizing the API. Thanks!

cc/ @tmccombs @stjepang 

## Refs

- #369 
- #549 
- #440 
- #217

tmccombs:
This depends on #369 and fixes #565

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ default = [
   "pin-project-lite",
 ]
 docs = ["attributes", "unstable", "default"]
-unstable = ["std", "broadcaster"]
+unstable = ["std"]
 attributes = ["async-attributes"]
 std = [
   "alloc",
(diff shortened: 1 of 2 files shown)


dignifiedquire:
@tmccombs what is the state of this, why is this only a draft?

tmccombs:
Mostly because it is dependent on an open pull request.

dignifiedquire:
@tmccombs I think you should be able to finish this, now that #369 is merged

dignifiedquire:
thank you @tmccombs

PR #585, opened 2019-11-25
expose try_recv and try_send on channels

### LINKED ISSUE #579: Expose `Receiver::try_recv` API
The current channels are not analogous to the libstd ones. A notable missing feature is `Receiver::try_recv` for taking an item from the channel in a non blocking fashion.

I have a service where background workers conjure fairly expensive resources (start virtual machines, all of which are identical) and they then pass handles to these via a channel. On the other end is a user facing API where access to one such resource can be requested. I want to use `try_recv` to get an available resource from the channel. Because if there are no resources available I want the request to fail immediately. The API should yield immediate results, and no long queues should be built up.

yoshuawuyts:
Exposes `Receiver::try_recv` and `Sender::try_send` on channels. Closes #579. I'm proposing we put the types in the crate's top-level because we're still undecided. I'd like to move to unblock people first, and then consider how we might want to move types around. Thanks!

## Screenshot

Screenshot_2019-11-26 async_std sync - Rust

### DIFF
--- a/src/sync/channel.rs
+++ b/src/sync/channel.rs
@@ -32,6 +32,7 @@ use crate::sync::WakerSet;
 /// # Examples
 ///
 /// ```
+/// # fn main() -> Result<(), async_std::sync::RecvError> {
 /// # async_std::task::block_on(async {
 /// #
 /// use std::time::Duration;
@@ -51,10 +52,11 @@ use crate::sync::WakerSet;
 /// });
 ///
 /// task::sleep(Duration::from_secs(1)).await;
-/// assert_eq!(r.recv().await, Some(1));
-/// assert_eq!(r.recv().await, Some(2));
+/// assert_eq!(r.recv().await?, 1);
+/// assert_eq!(r.recv().await?, 2);
+/// # Ok(())
 /// #
-/// # })
+/// # }) }
 /// ```
 #[cfg(feature = "unstable")]
 #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
@@ -113,6 +115,7 @@ impl<T> Sender<T> {
     /// # Examples
     ///
     /// ```
+    /// # fn main() -> Result<(), async_std::sync::RecvError> {
     /// # async_std::task::block_on(async {
     /// #
     /// use async_std::sync::channel;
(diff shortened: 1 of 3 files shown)


stjepang:
How do you feel about this mild inconsistency where we sometimes use `Option` and sometimes `Result`?

```rust
fn try_recv(&self) -> Result<T, TryRecvError>;
fn recv(&self) -> Option<T>;
fn next(&self) -> Option<T>;
```

Or, to put it differently, should `recv` return a `Result` too?

yoshuawuyts:
@stjepang I think the inconsistency is okay here. `None` implies "no more items", where `Result` implies "something went wrong here". Arguably the `try_` variants could return `Option` too, but then it wouldn't be possible to distinguish between error kinds.

On the other hand [`mpsc::Receiver::recv`](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv) *does* return a Result. The way items are taken out without being fallible is by using `iter`, which returns `None` when done.

This is interesting; hadn't thought of this. I was going to say that the current state is good, but now looking at what `std` provides I'd almost say that following their lead might make more sense? [`crossbeam_channel::Receiver::recv`](https://docs.rs/crossbeam-channel/0.4.0/crossbeam_channel/struct.Receiver.html#method.recv) works the same way too!

stjepang:
Yeah... in that case, are you okay with `recv()` and `next()` (coming from `Stream`) differing in `Result` vs `Option`? They do the same thing, except `recv` is slightly more efficient that `next`.

```rust
fn recv(&self) -> Result<T, RecvError>;
fn next(&self) -> Option<T>;
```

yoshuawuyts:
@stjepang yeah, that sounds good. Seems to be the same in `crossbeam` and `std` as well (:

yoshuawuyts:
Note from triage: we should remove the `Disconnected` case from the `TrySendError` enum.

leo60228:
Is this a good alternative?
```rust
use futures::prelude::FutureExt;

fn try_send<T>(chan: Sender<T>, x: T) -> bool {
    chan.send(x).now_or_never().is_some()
}
```

yoshuawuyts:
@stjepang actually I'm unsure how to remove that case. Maybe let's leave it in for now and consider changing it as part of stabilization?

yoshuawuyts:
ping @stjepang

yoshuawuyts:
@dignifiedquire fixed the doctests, fingers crossed it passes now.

PR #588, opened 2019-11-26
Fix the docs and Debug output of BufWriter.

### LINKED ISSUE #586: BufWriter not flushed
If you run the attached code, connect a client, and then terminate it the time stamp is not flushed to disk (as expected?). Adding an explicit call to flush() fixed it but seems like it should not be necessary.

https://gist.github.com/mtthw-meyer/34d85197e01d686dacca5cf1cd6b28b3

withoutboats:
The BufWriter docs inaccurately stated that it flushes on drop, which it does not do. This PR changes the docs, as well as the example, to highlight that the user must explicitly flush a bufwriter.

There were also two places where the BufWriter code referred to it as a BufReader: in the link to the std docs, and in the Debug output. Those have also been fixed.

Closes #586

### DIFF
--- a/src/io/buf_writer.rs
+++ b/src/io/buf_writer.rs
@@ -22,14 +22,14 @@ pin_project! {
     /// times. It also provides no advantage when writing to a destination that is
     /// in memory, like a `Vec<u8>`.
     ///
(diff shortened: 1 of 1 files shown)


withoutboats:
Gladly :)

PR #597, opened 2019-11-28
Fix failed doc test and enable doc test on CI

k-nasa:
I fix doc test, because some documentation test is failed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,12 @@ jobs:
         command: test
         args: --all --features "unstable attributes"
 
+    - name: documentation test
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -63,7 +63,13 @@ jobs:
       uses: actions-rs/cargo@v1
       with:
         command: test
-        args: --all --features unstable attributes
+        args: --all --features "unstable attributes"
```
k-nasa: The argument here was wrong.
The argument here was wrong.
"attributes" was not recognized as an argument to features.

k-nasa:
May I merge this?
@yoshuawuyts @stjepang

Issue #564, opened 2019-11-20
Add Future::timeout


yoshuawuyts:
Similar to [`future::timeout`](https://docs.rs/async-std/1.0.1/async_std/future/fn.timeout.html) and [`Stream::timeout`](https://docs.rs/async-std/1.0.1/async_std/stream/trait.Stream.html#method.timeout) we should add a `Future::timeout` method.

We already have `Future::delay`, and timeout would be a great counterpart. This is similar to stdlib's [`max`](https://doc.rust-lang.org/std/cmp/fn.max.html) which is both an inherent method and a free function. Thanks!

miker1423:
I like to take this if no one is working on it.

stjepang:
@miker1423 Go for it :)

miker1423:
Sorry for the absence.
As I understood, I should add a new extension method to the Future trait?

miker1423:
Here is the implementation https://github.com/miker1423/async-std/commit/c1f7be5d42f9fbeb63250e2b4f75986d2ae94fec, I know the docs are incomplete, I just wrote to take away the warning.

yoshuawuyts:
@miker1423 yeah that's in the right direction. I don't believe the pub modifiers are needed; `pub(crate)` should suffice. If you could submit the PR it'd make it easier to review. Thanks heaps for your work1

miker1423:
This should be closed with the merged PR #600, thanks!

PR #600, opened 2019-11-30
Adding timeout extension method to Future trait

miker1423:
This would close issue #564


src/future/future/mod.rs
```diff
@@ -355,6 +356,18 @@ extension_trait! {
         {
             TryJoin::new(self, other)
         }
+
+        #[doc = r#"
+            Waits for both the future and a timeout, if the timeout completes before
+            the future, it returns an TimeoutError.
```
k-nasa: Can you write test code following other methods?

https://docs.rs/async-std/1.2.0/async_std/future/trait.Future.html#method.race

miker1423:
Hello! I got two errors like the following on the rust fmt check, but I don't understand what I should do to solve them.
```
 #[test]
 fn should_not_timeout() {
-    task::block_on(async {        
+    task::block_on(async {
         let fut = future::ready(0);
         let dur = Duration::from_millis(100);
         let res = fut.timeout(dur).await;
Diff in /home/runner/work/async-std/async-std/tests/timeout_future.rs at line 25:

```

tests/timeout_future.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+#[test]
+fn should_not_timeout() {
+    task::block_on(async {        
+        let fut = future::ready(0);
+        let dur = Duration::from_millis(100);
+        let res = fut.timeout(dur).await;
+        assert!(res.is_ok());
+    });
```
k-nasa: It is nice! However, I would like to write this as a **doc test**.
I think it will be easier to use by leaving it as a document.

I think the document comments are as follows
```rust
#[doc = r#"
    Waits for both the future and a timeout, if the timeout completes before
    the future, it returns an TimeoutError.

    # Examples

    ```
    # async_std::task::block_on(async {
    let fut = future::pending::<()>();
    let dur = Duration::from_millis(100);
    assert!(fut.timeout(dur).await.is_err());

    let fut = future::ready(0);
    let dur = Duration::from_millis(100);
    assert!(fut.timeout(dur).await.is_ok());
    # });
    ```
"#]
```

k-nasa:
@miker1423 
My editor highlights unnecessary spaces at the end of lines. Maybe this is the cause of the error?
Can I fix it by running `cargo fmt`?

スクリーンショット 2019-12-06 11 37 12

tests/timeout_future.rs
miker1423: Yes, I wrote it after I had pushed earlier, but I was stuck with the format haha

tests/timeout_future.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[test]
+fn should_not_timeout() {
+    task::block_on(async {
+        let fut = future::ready(0);
+        let dur = Duration::from_millis(100);
+        let res = fut.timeout(dur).await;
+        assert!(res.is_ok());
+    });
+}
```
k-nasa: This test is a duplicate of the doc test.
I don't think it is necessary, but what about it?

miker1423: Sure, I can remove them.

src/future/future/mod.rs
```diff
@@ -355,6 +356,33 @@ extension_trait! {
...
+                    let fut = future::ready(0);
+                    let dur = Duration::from_millis(100);
+                    let res = fut.timeout(dur).await;
+                    assert!(res.is_ok());
+
+                    let fut = future::ready(0);
+                    let dur = Duration::from_millis(100);
+                    let res = fut.timeout(dur).await;
+                    assert!(res.is_ok())
+            # });
```
k-nasa: Please change like this
```rust
# async_std::task::block_on(async {
#
let fut = future::ready(0);
let dur = Duration::from_millis(100);
let res = fut.timeout(dur).await;
assert!(res.is_ok());

let fut = future::ready(0);
let dur = Duration::from_millis(100);
let res = fut.timeout(dur).await;
assert!(res.is_ok())
#
# });
```

The current document looks like this
スクリーンショット 2019-12-16 18 38 36

src/future/future/mod.rs
```diff
@@ -355,6 +356,40 @@ extension_trait! {
...
+
+            use async_std::prelude::*;
+            use async_std::future;
+            
+            let fut = future::ready(0);
+            let dur = Duration::from_millis(100);
+            let res = fut.timeout(dur).await;
+            assert!(res.is_ok());
+
+            let fut = future::ready(0);
```
k-nasa:
```suggestion
            let fut = future::pending::<()>();
```

src/future/future/mod.rs
```diff
@@ -355,6 +356,40 @@ extension_trait! {
...
+            let dur = Duration::from_millis(100);
+            let res = fut.timeout(dur).await;
+            assert!(res.is_ok())
```
k-nasa:
```suggestion
            assert!(res.is_err());
```

PR #606, opened 2019-12-04
fix tests/open_file

yjhmelody:


yjhmelody:
I found someone has fixed the ci and tests.

PR #615, opened 2019-12-11
Add an implementation of pending stream

lqf96:
This pull request adds an implementation of pending stream, which always returns `Poll::Pending` and never completes. The implementation is largely copy-pasted from [`futures_util::stream::pending`](https://github.com/rust-lang/futures-rs/blob/master/futures-util/src/stream/pending.rs), with trait implementations for `DoubleEndedStream`, `FusedStream` and `ExactSizeStream` added.

### DIFF
--- a/src/stream/pending.rs
+++ b/src/stream/pending.rs
@@ -28,10 +28,6 @@ impl<T> Stream for Pending<T> {
     fn poll_next(self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<Option<T>> {
         Poll::Pending
     }
-
-    fn size_hint(&self) -> (usize, Option<usize>) {
-        (0, Some(0))
-    }
 }
 
 impl<T> DoubleEndedStream for Pending<T> {
--- a/src/stream/pending.rs
+++ b/src/stream/pending.rs
@@ -1,0 +1,49 @@
+use std::marker::PhantomData;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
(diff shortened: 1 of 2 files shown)


src/stream/pending.rs
```diff
@@ -0,0 +1,45 @@
...
+/// 
+/// [`pending`]: fn.pending.html
+#[derive(Debug)]
+pub struct Pending<T> {
+    _marker: PhantomData<T>
+}
+
+/// Creates a stream that never returns any items.
+/// 
+/// The returned stream will always return `Pending` when polled.
```
yoshuawuyts: Could we add a usage example here?

lqf96: The reason I haven't added an example here is that the pending stream will never return anything. It is possible to add an example that polls instead of `await`s the stream but I think the behavior of pending stream is straightforward enough that such an example isn't needed. Plus, the [`futures_util::stream::pending`](https://docs.rs/futures-util/0.3.1/futures_util/stream/fn.pending.html) function does not come with an example.

yoshuawuyts: > Plus, the futures_util::stream::pending function does not come with an example.

We have an example for [`future::pending`](https://docs.rs/async-std/1.3.0/async_std/future/fn.pending.html) and given `stream::pending` fills much the same role I'd like us to have an example here.

The example we're using there is in combination with a `timeout`, showing the timeout is always hit as the stream will never resolve. That's probably a useful practical use case as well, and showcasing it for stream would be useful as well I think (:

k-nasa:
@lqf96 
This appears to have stopped.
If you're busy, I'll take over for this.

Issue #620, opened 2019-12-13
tests on master failing


yoshuawuyts:
Looks like #562 broke the tests; the double ended `from_iter` function was removed, and now tests on master are broken. https://github.com/async-rs/async-std/runs/347468653

cc/ @felipesere would you mind taking a look? Ideally `stream::FromIter` would implement: 

```rust
impl <T: DoubleEndedStream> DoubleEndedStream for FromIter<T>;
```

so that things just work.

felipesere:
@yoshuawuyts this can be closed now? 😄 
Thanks for fixing it.

skade:
Should be. Reopen if not true.

PR #621, opened 2019-12-13
fix stream doc hiccup

yoshuawuyts:
This removes https://docs.rs/async-std/1.3.0/async_std/stream/stream/index.html from the docs which was accidentally introduced in #562. Thanks!


yoshuawuyts:
tests will fail until we resolve #620

yoshuawuyts:
Tests are broken; merging this as part of a scavenger hunt to fix them.

PR #622, opened 2019-12-13
Fix ci

### LINKED ISSUE #620: tests on master failing

felipesere:
Resolves #620 by helping the compiler infer types, use the stream::from_iter(...) function since it was move there from double_ended_stream, and finally mark the `double_ended_stream` module as not public.


yoshuawuyts:
Superseded by #630. Thanks!

PR #625, opened 2019-12-14
Use ?Sized in Mutex and RwLock

### LINKED ISSUE #609: Allow unsized types for Mutex?
I'm trying to use a `dyn Trait` with a `Mutex` while moving from `std::sync::Mutex` to `async_std::sync::Mutex`. Is it possible for the `Mutex` in this crate to have `T: ?Sized`? The the std version allows for `?Sized` types, but I don't know if this would be an issue for async_std.

nbdd0121:
Fix #609

### DIFF
--- a/src/sync/mutex.rs
+++ b/src/sync/mutex.rs
@@ -39,14 +39,14 @@ use crate::task::{Context, Poll};
 /// #
 /// # })
 /// ```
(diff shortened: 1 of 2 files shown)


nbdd0121:
Sorry, this is intended to be a draft PR

PR #626, opened 2019-12-14
Add std::os::unix::{register, unregister, Interest}

### LINKED ISSUE #293: Expose net::driver::Watcher mechanism
Crates implementing `Evented` should be able to register a reactor.

Follow up from discussion #65

leo60228:
Closes #293, based on https://github.com/async-rs/async-std/issues/293#issuecomment-549049417

### DIFF
--- a/src/net/driver/mod.rs
+++ b/src/net/driver/mod.rs
@@ -1,6 +1,5 @@
 use std::fmt;
 use std::sync::{Arc, Mutex};
-use std::os::raw::c_int;
 
 use mio::{self, Evented};
 use once_cell::sync::Lazy;
@@ -10,6 +9,12 @@ use crate::io;
 use crate::task::{Context, Poll, Waker};
 use crate::utils::abort_on_panic;
 
+#[cfg(unix)]
+pub use std::os::unix::io::RawFd;
+
+#[cfg(not(unix))]
+pub enum RawFd {}
+
 /// Data associated with a registered I/O handle.
 #[derive(Debug)]
 pub struct Entry {
@@ -17,7 +22,7 @@ pub struct Entry {
     token: mio::Token,
 
     /// File descriptor.
-    fd: Option<c_int>,
+    fd: Option<RawFd>,
 
     /// Tasks that are blocked on reading from this I/O handle.
     pub readers: Mutex<Vec<Waker>>,
@@ -62,7 +67,7 @@ impl Reactor {
     }
 
     /// Registers an I/O event source and returns its associated entry.
-    pub fn register(&self, source: &dyn Evented, fd: Option<c_int>) -> io::Result<Arc<Entry>> {
+    pub fn register(&self, source: &dyn Evented, fd: Option<RawFd>) -> io::Result<Arc<Entry>> {
         let mut entries = self.entries.lock().unwrap();
 
         // Reserve a vacant spot in the slab and use its key as the token value.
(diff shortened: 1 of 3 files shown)


src/net/driver/mod.rs
```diff
@@ -51,14 +55,14 @@ impl Reactor {
...
-        let entry = reactor.register(&reactor.notify_reg.0)?;
+        let entry = reactor.register(&reactor.notify_reg.0, None)?;
         reactor.notify_token = entry.token;
 
         Ok(reactor)
     }
 
     /// Registers an I/O event source and returns its associated entry.
-    fn register(&self, source: &dyn Evented) -> io::Result<Arc<Entry>> {
+    pub fn register(&self, source: &dyn Evented, fd: Option<c_int>) -> io::Result<Arc<Entry>> {
```
yoshuawuyts: `Evented` shouldn't be part of our pub interface, and `fd` should be `RawFd`.

src/net/driver/mod.rs
```diff
@@ -82,7 +87,7 @@ impl Reactor {
     }
 
     /// Deregisters an I/O event source associated with an entry.
-    fn deregister(&self, source: &dyn Evented, entry: &Entry) -> io::Result<()> {
+    pub fn deregister(&self, source: &dyn Evented, entry: &Entry) -> io::Result<()> {
```
yoshuawuyts: `Evented` shouldn't be part of our pub interface.

sdroege:
>  I really like the idea of having an `async_std::os::unix` version of our watcher abstraction, and another for `async_std::os::windows`. That way we can differentiate between `RawFd` and `RawSocket` in constructor interfaces.

Note that you will also want to support `RawHandle` on Windows, e.g. for [event objects](https://docs.microsoft.com/en-us/windows/win32/sync/event-objects).

src/net/driver/mod.rs
leo60228: This module is `pub(crate)`.

leo60228:
How does that work? Is there an EventedHandle in mio like there is EventedFd?

leo60228:
mio doesn't seem to have a way to register a RawHandle.

src/net/driver/mod.rs
leo60228: This module is `pub(crate)`, and RawFd is Unix-only. I didn't want to copy/paste the method and struct, though perhaps that's preferable for memory usage? It'd be heavily preferable to use `Option<!>` instead of having to put every call in a cfg, though.

leo60228:
FWIW, I switched from `impl AsRawFd` to `RawFd` because `RawFd` doesn't implement `AsRawFd`.

DemiMarie-parity:
Would it be possible to also expose methods on `UdpSocket` et al?  I really don’t want to write more code than I must for Windows.

Keruspe:
It would be great if we could have access to what was the actual event.
With the current implementation, the waker gets called, but no way of knowing if the event was for read or write (and how are treated the error, read_hup and write_hup events?)

dignifiedquire:
This will need a different approach, now that we are no longer using mio, and use smol as the underlying runtime.

leo60228:
Yeah, this design doesn't make sense anymore, especially since you can just use `smol::Async` directly. It might make sense to re-export that somewhere, but that's about it.

PR #628, opened 2019-12-14
Make WriteFmtFuture must_use

### LINKED ISSUE #627: missing must_use for writeln on File
I'm having a hard time finding the exact definitions that are relevant because everything is macros so I'm not sure where exactly the error is, but I had some code that tried to `writeln!` to an `async_std::fs::File` but forgot to await it, and didn't get any `unused-must-use` lints.

fenhl:
Fixes #627. Thanks to @jebrosen for pointing out the location of the issue.


fenhl:
CI failure appears to be unrelated.

PR #629, opened 2019-12-16
upgrade log, remove kv-log-macro

yoshuawuyts:
Removes `kv-log-macro`, upgrades the `log` crate. Kv logging is now allowed by the macros so this gets rid of an extra dependency. Thanks!


skade:
LGTM, I'm merging.

PR #635, opened 2019-12-17
Revert "upgrade log, remove kv-log-macro"

k-nasa:
Reverts async-rs/async-std#629

Currently it seems to be a compile error.

```
error: failed to select a version for the requirement `log = "^0.4.10"`
  candidate versions found which didn't match: 0.4.8, 0.4.7, 0.4.6, ...
  location searched: crates.io index
required by package `async-std v1.3.0 (/Users/nasa/lab/oss/k-nasa/async-std)`
```

Isn't it because of being yanked?

スクリーンショット 2019-12-18 8 03 02

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ default = [
   "crossbeam-channel",
   "crossbeam-deque",
   "futures-timer",
+  "kv-log-macro",
   "log",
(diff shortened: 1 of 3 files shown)


k-nasa:
Merge as soon as ci passes

PR #636, opened 2019-12-18
Add `UdpSocket::{poll_recv_from, poll_send_to}`

DemiMarie-parity:
This is needed for QUIC implementations based on async-std.  It could be
done with `UdpSocket::{recv_from, send_to}`, but this requires boxing,
reference counting, and `unsafe` code.

### DIFF
--- a/src/net/udp/mod.rs
+++ b/src/net/udp/mod.rs
@@ -1,6 +1,6 @@
 use std::io;
-use std::net::SocketAddr;
-use std::net::{Ipv4Addr, Ipv6Addr};
+use std::net::{Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::task::{Context, Poll};
 
 use crate::future;
 use crate::net::driver::Watcher;
@@ -69,9 +69,7 @@ impl UdpSocket {
     /// ```
     pub async fn bind<A: ToSocketAddrs>(addrs: A) -> io::Result<UdpSocket> {
         let mut last_err = None;
-        let addrs = addrs
-            .to_socket_addrs()
-            .await?;
+        let addrs = addrs.to_socket_addrs().await?;
 
         for addr in addrs {
             match mio::net::UdpSocket::bind(&addr) {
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
@DemiMarie-parity hi, thanks for opening this PR! -- This is a bit of a design departure from where we were at. The team has been away over the holidays, but we're meeting up again next week to catch up and discuss future directions. I'll make sure we discuss this PR then!

Also if Parity is experimenting with async-std, I'm sure I speak for the team we would love to hear more about the directions you're taking to ensure we can support you. We'd love to see you succeed!

stjepang:
> It could be
> done with `UdpSocket::{recv_from, send_to}`, but this requires boxing,
> reference counting, and `unsafe` code.

I believe `poll_recv_from()` can be implemented using `recv_from()` without any problems:

```rust
pub fn poll_recv_from(
    &self,
    cx: &mut Context<'_>,
    buf: &mut [u8],
) -> Poll<io::Result<(usize, SocketAddr)>> {
    Pin::new(&mut self.recv_from(buf)).poll(cx)
}
```

So you could simply do `Pin::new(&mut self.recv_from(buf)).poll(cx)` instead of using `poll_recv_from()`. The reason why this works is that the `recv_from()` future has no state and does nothing on drop -- it simply forwards the call to `watcher.poll_read_with()`.

DemiMarie-parity:
@stjepang Two caveats:

* First, this adds unnecessary overhead.  The compiler may be able to optimize it away, but I would rather not count on that.
* Second, this is contrary to the normal behavior of futures, so it would need to be documented.

yoshuawuyts:
> First, this adds unnecessary overhead. The compiler may be able to optimize it away, but I would rather not count on that.

Pin is a zero-sized type, and LLVM is incredibly good at inlining when compiling Rust with `--release`. Even if this failed to optimize (which seems unlikely) the performance impact would be limited. Overall this doesn't seem like a convincing enough argument to warrant an API addition.

---

> Second, this is contrary to the normal behavior of futures, so it would need to be documented.

Do you have any suggestions to what this could look like? The [pinning chapter](https://rust-lang.github.io/async-book/03_async_await/01_chapter.html) in the async book goes into this a bit. And so do the [`std::pin`](https://doc.rust-lang.org/std/pin/) docs. But not *too well*. 

Maybe there is something there we could do to improve this? Or do you think an async-std specific addition would be more helpful?

DemiMarie-parity:
The biggest issue is that async functions don’t implement `Unpin`, so I need to either box the future (which is actual overhead) or use the unsafe `Pin::new_unchecked` function.  Such use is correct, but users should not be forced to write unsafe code when it is not necessary.

yoshuawuyts:
@DemiMarie-parity [pin-project](https://docs.rs/pin-project/0.4.7/pin_project/) and [pin-project-lite](https://docs.rs/pin-project-lite/0.1.4/pin_project_lite/) are the recommended ways of replacing the unsafe code with safe compile-time invariant checks. The stdlib doesn't include either, but the Pin API was very much designed with these abstractions in mind.

Either way, I think we can have a resolution for this issue. Going to go ahead and close this!

Issue #553, opened 2019-11-18
Implement Clone for TcpStream


yoshuawuyts:
As per https://github.com/async-rs/async-std/issues/365#issuecomment-554988241 we should implement `Clone` for `TcpStream`. The inner fd is already wrapped in an `Arc`, so cloning it should be straightforward. Thanks!

k-nasa:
After all need a source clone, right?
I think you need to call mio :: tcp_stream :: try_clone.
If so, this should be try_clone, not clone.

```
pub struct Watcher<T: Evented> {
    /// Data associated with the I/O handle.
    entry: Arc<Entry>,

    /// The I/O event source.
    source: Option<T>,
}
```

yoshuawuyts:
@k-nasa I believe @stjepang argued this should be `clone` instead because we already have an `Arc<TcpStream>` internally; so there's no need to do an extra syscall to obtain an extra handle to it.

yoshuawuyts:
Seems like this was accidentally closed (:

atbentley:
@yoshuawuyts, it looks like @stjepang is saying that `async_std::net::TcpStream` currently doesn't use an `Arc` internally, but it could?

If that's the case, I started having a look at what it would take to do that. I changed the `TcpStream` struct to be:

```
#[derive(Debug)]
pub struct TcpStream {
    pub(super) watcher: Watcher<Arc<mio::net::TcpStream>>,
}
```

And updated all the references to the inner value of `TcpStream` to respect the `Arc`. I ran into an issue with the `IntoRawFd` trait though. What would this trait look like if we use an Arc? My understanding of rust is still pretty surface level but it seems like `into_raw_fd` will move the value of the fd, which will be impossible here because `Arc` shares the reference to potentially many copies.

```
    impl IntoRawFd for TcpStream {
        fn into_raw_fd(self) -> RawFd {
            self.watcher.into_inner().as_ref().into_raw_fd()
        }
    }
```

So I just commented it out and proceeded with the `Clone` trait. Which I tested using some toy examples and that seemed to work.
 
```
impl Clone for TcpStream {
    fn clone(&self) -> Self {
        TcpStream {
            watcher: Watcher::new(self.watcher.get_ref().clone())
        }
    }
}
```

[...]

PR #642, opened 2019-12-23
Add try_clone for TcpStream

atbentley:
From the discussion in #553 

I also updated the client/server example to use `try_clone` instead of `Arc`, as well as added a small test.

### DIFF
--- a/examples/a-chat/server.rs
+++ b/examples/a-chat/server.rs
@@ -1,7 +1,4 @@
-use std::{
-    collections::hash_map::{Entry, HashMap},
-    sync::Arc,
-};
+use std::collections::hash_map::{Entry, HashMap};
 
 use futures::{channel::mpsc, select, FutureExt, SinkExt};
 
@@ -40,8 +37,7 @@ async fn accept_loop(addr: impl ToSocketAddrs) -> Result<()> {
 }
 
 async fn connection_loop(mut broker: Sender<Event>, stream: TcpStream) -> Result<()> {
-    let stream = Arc::new(stream);
-    let reader = BufReader::new(&*stream);
+    let reader = BufReader::new(&stream);
     let mut lines = reader.lines();
 
     let name = match lines.next().await {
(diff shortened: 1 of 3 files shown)


atbentley:
Not sure why this is failing on windows, is anyone able to point me in a direction here?

yoshuawuyts:
@atbentley I'm unsure, but it seems this might be an issue in `mio`:

```txt
---- cloned_streams stdout ----
thread 'cloned_streams' panicked at 'cannot register an I/O event source: Os { code: 87, kind: Other, message: "The parameter is incorrect." }', src\libcore\result.rs:1188:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```
We're not passing any extra parameters to the underlying OS primitives here; so it feels like it's coming from a lower level. This makes me wonder if going down the `Arc` route isn't the more reliable option.

stjepang:
I believe this is the same problem as in https://github.com/tokio-rs/tokio/issues/774

guiri14:
Definitely looking forward to this. As it stands, I don’t see any way to read a socket on one thread and write it on another on rust stable. I’ve worked around it in the meantime using Arc::get_mut_unchecked, which feels pretty gross.  It does seem to work fine, though.

guiri14:
I attempted creating a std TcpStream, cloning it, and then using async_std::net::TcpStream's From<> impl to convert the std stream. I also received the error

```thread 'main' panicked at 'cannot register an I/O event source: Os { code: 87, kind: Other, message: "The parameter is incorrect." }', C:\Users\jfont\.cargo\registry\src\github.com-1ecc6299db9ec823\async-std-1.4.0\src\net\driver\mod.rs:183:20```

This happens after I've already converted the first one and I'm converting the second one. Here's the demonstration code:
```
let tcp_a = std::net::TcpStream::connect("127.0.0.1:43800")?;
let tcp_b = tcp_a.try_clone()?;
let reader = Arc::new(Mutex::new(async_std::net::TcpStream::from(tcp_a)));
let writer = Arc::new(Mutex::new(async_std::net::TcpStream::from(tcp_b)));
```

PR #647, opened 2019-12-26
feat: do not require default feature for unstable

dignifiedquire:
This allows using the unstable features, without accidentally enabling the runtime.


dignifiedquire:
@yoshuawuyts all fixed

PR #648, opened 2019-12-27
Fixing inaccurate function description in udp::recv

spacekookie:
The recv function only returns the number of bytes. The current
documentation makes the destiction between recv and recv_from
a bit confusing.


yoshuawuyts:
Filed #661 to fix CI.

PR #660, opened 2020-01-07
Showcase the ecosystem

mehcode:
I think it would be a good idea to start calling out crates that you can use with async-std to accomplish tasks that come up in web development: TLS, WebSockets, Web Server, Web Client, Database, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,20 @@ documentation] on how to enable them.
 [cargo-add]: https://github.com/killercup/cargo-edit
 [features documentation]: https://docs.rs/async-std/#features
 
+## Ecosystem
+ 
+ * [async-tls](https://crates.io/crates/async-tls) — Async TLS/SSL streams using **Rustls**. 
+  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,6 +120,18 @@ documentation] on how to enable them.
 [cargo-add]: https://github.com/killercup/cargo-edit
 [features documentation]: https://docs.rs/async-std/#features
 
+## Ecosystem
+ 
+ * [async-native-tls](https://crates.io/crates/async-native-tls) — Native **TLS** for Async. Native TLS for futures and async-std.
```
sdroege: You might also want to add async-tls here :) https://crates.io/crates/async-tls

mehcode: Done. It sure would be nice to be able to universally configure stuff like "which TLS back-end do I want" at a global level in Cargo.toml and stuff like those two crates can merge and not care about stuff like mutually exclusive feature flags.

Issue #639, opened 2019-12-21
StreamExt::collect does not preserve Send


rmanoka:
Hi,

The `StreamExt::collect` in this crate (with `unstable` feature-flag) converts a `Send` stream into future that is not `Send` anymore.  On the other hand, `futures::StreamExt` seems to preserve this auto trait.  This is not a mission-critical issue, but thought I'll bring it to your attention anyway.

Test code to reproduce this, that yields a compilation error:

``` rust
#[async_std::test]
async fn test_send() {
    fn test_send_trait<T: Send>(_: &T) {}

    let stream = futures::stream::pending::<()>();
    test_send_trait(&stream);

    use async_std::prelude::StreamExt;
    let fut = stream.collect::<Vec<_>>();

    // This line triggers a compilation error
    test_send_trait(&fut);
}
```

Tested on 1.4.0

yoshuawuyts:
@rmanoka hi, thanks for raising this! Heh yeah that's no good; we should fix that.

yoshuawuyts:
Okay I'm able to reproduce this. A concrete example of the issue above is:

```rust
use async_std::prelude::*;
use async_std::io;
use async_std::task;

#[test]
fn test_send() -> io::Result<()> {
    task::block_on(async {
        fn test_send_trait<T: Send>(_: &T) {}

        let stream = futures::stream::pending::<()>();
        test_send_trait(&stream);


        task::spawn(async move {
            let _output = stream.collect::<Vec<_>>().await; // compile error here
        });
        Ok(())
    })
}
```

A possible solution here is to add `Send` back to the trait bounds of the Future, which I believe we originally had. Trying that out now.

PR #665, opened 2020-01-10
fix(stream): add send guards on collect

yoshuawuyts:
WIP towards fixing #639. This currently doesn't work yet; needs further investigation. Thanks!


dignifiedquire:
@stjepang any chance you could help with this. This is blocking me from using Streams inside `task::spawn` at the moment. I have tried, but failed to understand the `extension_trait` macro, which I believe needs to be modified in order to add the `Send` bound on `fn collect`

PR #667, opened 2020-01-12
Use `take_while` instead of `scan` in `impl` of `Product`, `Sum` and `FromStream` for `Option` and `Result`

olegnn:
Use of `take_while` instead of `scan` is much clear because
- `take_while` is designed for this kind of tasks
- Current implementation doesn't use `scan` state

### DIFF
--- a/src/option/from_stream.rs
+++ b/src/option/from_stream.rs
@@ -2,7 +2,7 @@ use std::pin::Pin;
 
 use crate::prelude::*;
 use crate::stream::{FromStream, IntoStream};
-use crate::utils::identity;
+use std::convert::identity;
 
 impl<T, V> FromStream<Option<T>> for Option<V>
 where
@@ -33,11 +33,7 @@ where
                 .collect()
                 .await;
 
-            if found_none {
-                None
-            } else {
-                Some(out)
-            }
+            if found_none { None } else { Some(out) }
         })
     }
 }
--- a/src/option/from_stream.rs
+++ b/src/option/from_stream.rs
@@ -17,24 +17,25 @@ where
         let stream = stream.into_stream();
 
         Box::pin(async move {
-            // Using `scan` here because it is able to stop the stream early
+            // Using `take_while` here because it is able to stop the stream early
             // if a failure occurs
-            let mut found_error = false;
+            let mut found_none = false;
             let out: V = stream
-                .scan((), |_, elem| {
-                    match elem {
-                        Some(elem) => Some(elem),
-                        None => {
-                            found_error = true;
-                            // Stop processing the stream on error
-                            None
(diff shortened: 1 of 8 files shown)


src/option/from_stream.rs
```diff
@@ -17,24 +17,26 @@ where
...
-                            None
-                        }
+                .take_while(|elem| {
+                    elem.is_some() || {
+                        found_none = true;
+                        // Stop processing the stream on `None`
+                        false
                     }
                 })
+                .map(Option::unwrap)
```
yoshuawuyts: This pattern here is quite surprising to me, and at a glance I'm not sure what it does. This doesn't feel like a strict improvement in legibility to me.

olegnn: I don't think that `scan` is well suited for this case because it
- Provides unused functionality
- Implicitly converts `Option<T>` to `T` (`.map(Option::unwrap)` does it explicitly)
- Is rarely used (I think much less people know how this combinator works, if compare with `take_while` or `map`).

However I changed previous approach to `filter_map` with placeholder function, hope it's more understandable.

Waiting for your feedback. 😊

P.S. At least first argument in `scan` closure should be `()` instead of `_` to highlight that internal state isn't used at all.

src/utils.rs
```diff
@@ -52,6 +52,14 @@ pub fn random(n: u32) -> u32 {
...
 }
 
+/// Returns given argument without changes.
+#[allow(dead_code)]
+#[doc(hidden)]
+#[inline(always)]
+pub(crate) fn identity<T>(arg: T) -> T {
+    arg
+}
+
```
yoshuawuyts: What is the difference between this function and [`std::convert::identity`](https://doc.rust-lang.org/std/convert/fn.identity.html)?

olegnn: Ah, thanks, I missed that this function was added a year ago already...

src/option/from_stream.rs
yoshuawuyts: What you're saying makes sense!

Question now though: it seems impls are now half `scan` based, and the other half `take_while` based. Would it make sense to move all of them over to be based on `take_while`?

olegnn: The problem with `Result`s implementation is that it takes first `Err` from provided stream.
However `take_while` accepts reference, while `scan` takes ownership. This can be solved by more complex `filter_map`, which also takes ownership.

olegnn: Or I thought about passing `found_error`'s mut reference as `scan` state and changing it in appropriate way.

olegnn:
@yoshuawuyts I finally rewrote `Result`s implementations using `take_while`.

PR #671, opened 2020-01-13
Fix docs for UdpSocket::send 

### LINKED ISSUE #652: docs for UdpSocket::send have same example as send_to
cf. https://docs.rs/async-std/1.4.0/async_std/net/struct.UdpSocket.html#method.send

Noah-Kennedy:
Minor change to alter documentation for `send` to no longer duplicate the documentation for `send_to`.
Fixes #652.

### DIFF
--- a/src/net/udp/mod.rs
+++ b/src/net/udp/mod.rs
@@ -244,9 +244,12 @@ impl UdpSocket {
         }))
     }
 
-    /// Sends data on the socket to the given address.
+    /// Sends data on the socket to the remote address to which it is connected.
     ///
-    /// On success, returns the number of bytes written.
+    /// The [`connect`] method will connect this socket to a remote address.
+    /// This method will fail if the socket is not connected.
+    ///
+    /// [`connect`]: #method.connect
     ///
     /// # Examples
     ///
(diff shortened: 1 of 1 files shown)


src/net/udp/mod.rs
```diff
@@ -297,12 +300,11 @@ impl UdpSocket {
     /// #
     /// use async_std::net::UdpSocket;
     ///
-    /// let socket = UdpSocket::bind("127.0.0.1:0").await?;
+    /// let socket = UdpSocket::bind("127.0.0.1:7878").await?;
     /// socket.connect("127.0.0.1:8080").await?;
+    /// let bytes = socket.send(b"Hi there!").await?;
     ///
-    /// let mut buf = vec![0; 1024];
-    /// let n = socket.recv(&mut buf).await?;
```
k-nasa: Why remove `recv` call?
Is this an example of the `recv` function?

spacekookie:
>-    /// let mut buf = vec![0; 1024];
>-    /// let n = socket.recv(&mut buf).await?;
>
>Why remove `recv` call?
>Is this an example of the `recv` function?

If its not, it should be

src/net/udp/mod.rs
Noah-Kennedy: Good catch. It looks like I put the docs in the wrong spot. I'll correct that issue right now

Noah-Kennedy: Fixed it

PR #674, opened 2020-01-14
lib.rs: fixing various doc tests

spacekookie:
I was gonna work on some other stuff but the doc tests in lib.rs appeared
to be failing, so this is my attempt to fix them.  They're still broken,
I'll just force push this branch later with more fixes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -158,13 +158,13 @@
 //! async fn main() {
 //!     let a = async { 1u8 };
 //!     let b = async { 2u8 };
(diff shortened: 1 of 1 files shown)


spacekookie:
Hmm, ah so turns out I needed to enable `unstable attributes` for those tests, I'll run them again and see if that makes a difference >:P

spacekookie:
I don't know if this still needs to be merged then, seeing as I think I
mostly misunderstood how to run the tests.  Maybe there could be a
static-assert in the tests that require certain features? 🤔

PR #677, opened 2020-01-15
Avoid doctest errors

kornelski:
It seems like proc macros don't work in doctests:

```
test src/lib.rs -  (line 167) ... FAILED
test src/lib.rs -  (line 154) ... FAILED
test src/lib.rs -  (line 141) ... FAILED
```

> error[E0433]: failed to resolve: could not find `main` in `async_std`
> ```
> --> src/lib.rs:146:14
>   |
> 8 | #[async_std::main]
>   |              ^^^^ could not find `main` in `async_std`
> ```

The problem is visible when running `cargo test --all`. Tested on Rust 1.40 and 1.42 nightly.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,7 +138,8 @@
 //!
 //! Call an async function from the main function:
 //!
-//! ```
+#![cfg_attr(feature = "attributes", doc = "```")]
+#![cfg_attr(not(feature = "attributes"), doc = "```rust,ignore")]
 //! async fn say_hello() {
 //!     println!("Hello, world!");
 //! }
@@ -151,7 +152,8 @@
 //!
 //! Await two futures concurrently, and return a tuple of their output:
 //!
-//! ```
+#![cfg_attr(all(feature = "unstable", feature = "attributes"), doc = "```")]
+#![cfg_attr(not(all(feature = "unstable", feature = "attributes")), doc = "```rust,ignore")]
 //! use async_std::prelude::*;
 //!
 //! #[async_std::main]
@@ -164,7 +166,8 @@
 //!
 //! Create a UDP server that echoes back each received message to the sender:
 //!
-//! ```no_run
+#![cfg_attr(feature = "attributes", doc = "```no_run")]
+#![cfg_attr(not(feature = "attributes"), doc = "```rust,ignore")]
 //! use async_std::net::UdpSocket;
 //!
 //! #[async_std::main]


yoshuawuyts:
@kornelski in order to run the tests a few features need to be enabled:

```sh
$ cargo test --all --features 'attributes,unstable'
```

This is the way it's setup in our CI as well. Though as evidenced by this PR it's not discoverable, and we should at the very least document this in the README. I wish we could add this to profiles somehow, but I don't believe we can.

```toml
[profiles.test]
features = [ "attributes", "unstable" ]
```

kornelski:
That's unexpected indeed. 

Doctests are an odd case, but for tests in general I expect them to be gated by the same flags as features are, so that `cargo test` only tests enabled features.

Maybe these doctests could be made conditional? This seems to be a supported syntax:

    #[cfg_attr(feature = "attributes", doc = "#[async_std::main]")]

kornelski:
`cfg_attr` actually works!

yoshuawuyts:
Ah, it seems we're missing a `cargo fmt` pass. CI is failing. Would you mind updating?

kornelski:
rustfmt doesn't understand this syntax and breaks Markdown in the comments.

I'm incredibly disappointed that rustfmt missed the point of gofmt. It unconditionally overrides _all_ human decisions, which makes it a dumb and destructive tool, unlike gofmt which more or less only cleans up known-bad formatting. Because of that I think unconditionally using rustfmt is a bad idea, and I'm not going to use rustfmt.

taiki-e:
Yeah, rustfmt error is bug/limitation of *unstable* `format_code_in_doc_comments` option.

Removing usage of `format_code_in_doc_comments` option should be the simplest fix, if we keep using rustfmt. 
And I prefer to do this for now because of the following reasons:
* this option is unstable 
* (AFAIK) many docs with examples of this crate are in macros and rustfmt doesn't format code in macros, so are the benefits of using this option in this crate should not be very big.

---

fwiw: to keep using `format_code_in_doc_comments`, use cfg on hidden function instead of using `#[doc]` attribute, like:

```rust
/// ```
/// # #[cfg(feature = "attributes")]
/// # fn dox() {
/// examples...
/// # }
/// ```
```

k-nasa:
ping @kornelski

kornelski:
Sorry, I don't like enforced rustfmt checks, and I'm not going to satify rustfmt's arbitrary requirements.

Issue #605, opened 2019-12-03
cfg_default! and its friends interfere with tooling


ypoluektovich:
With the Rust plugin in IntelliJ IDEA, the use of `cfg_default!`, `cfg_std!`, and other similar macros prevents the IDE from discovering the top-level module declarations. That results in absolutely no IDE features working for the types and functions contained within those modules.

- intellij-rust 0.2.111.2147-193
- IntelliJ IDEA 2019.3 (Build #IU-193.5233.102)
- `async-std = "1.2.0"`

Noah-Kennedy:
I know that work has been underway to fix this for quite some time in the IntelliJ Rust plugin. This thread discusses it more: https://github.com/intellij-rust/intellij-rust/issues/1191. As a Clion user myself, I feel your pain, but I think this is a problem for the tool more than the libraries.

PR #684, opened 2020-01-19
Changing task::block_on to park after a single poll

spacekookie:
This was previously discussed in #605 and others as a source of high
CPU load when sleeping tasks because of the overhead created by
retrying a future in short succession.

Personally I've noticed this behaviour as well, but wasn't able to
recreate test settings since making this change yet.  Maybe someone
else has some good benchmarks to run.


spacekookie:
> Looks like there's an unused import of `std::thread` that makes CI
> fail, otherwise looks good to me! :)

I updated the branch.

Issue #678, opened 2020-01-15
Why is TcpStream::connect no longer using mio::net::TcpStream::connect?


twittner:
In https://github.com/async-rs/async-std/commit/d25dae54191c8a922410cd2e3ca39e631d28982b, the implementation of `TcpStream::connect` was switched from using mio's non-blocking connect to the blocking `std::net::TcpStream::connect` which is performed in a background thread created by `spawn_blocking`. May I ask why you decided to no longer use mio's non-blocking connect? It seems the background thread spawned is unaware of any future cancellation, e.g. due to timeouts, but continues to run until all SYN retries have been exhausted which may easily last a couple of minutes. This in turn can cause a high number of threads to be created in the background.

stjepang:
Hmmm, that definitely looks like an oversight during the refactor. I agree with you - we should use mio's non-blocking connect. Happy to review a PR for it!

PR #687, opened 2020-01-23
Use non-blocking connect for TcpStream.

twittner:
Instead of spawning a background thread which is unaware of any timeouts but continues to run until the TCP stack decides that the remote is not reachable we use mio's non-blocking connect.

mio's `TcpStream::connect` returns immediately but the actual connection is usually just in progress and we have to be sure the socket is writeable before we can consider the connection as established.

~The `Watcher` and `Reactor` APIs have changed a bit to (a) allow registration together with an initial set of `Waker`s in order to avoid missing any wakeups that may occur before we even registered a `Waker` with `Watcher::{poll_read_with, poll_write_with}`, and (b) to allow registration with readiness interests other than `PollOpt::all()`.~

**Update**: Following a suggestion of @stjepang we offer methods to check for read/write readiness of a `Watcher` instead of the approach in the previous paragraph to accept a set of `Waker`s when registering an event source. The changes relative to master are smaller and both methods look more useful in other contexts. Also the code is more robust w.r.t. wakeups of the `Waker` from clones outside the `Reactor`.

I am not sure if we need to add protection mechanisms against spurious wakeups from mio. Currently we treat the `Poll::Ready(())` from `Watcher::poll_write_ready` as proof that the non-blocking connect has finished, but if the event from mio was a spurious one, it might still be ongoing.

Context: #678.

### DIFF
--- a/src/net/driver/mod.rs
+++ b/src/net/driver/mod.rs
@@ -59,6 +59,17 @@ impl Reactor {
 
     /// Registers an I/O event source and returns its associated entry.
     fn register(&self, source: &dyn Evented) -> io::Result<Arc<Entry>> {
+        self.register_with(source, Interest::All, Vec::new(), Vec::new())
+    }
+
+    /// Registers an I/O event source and returns its associated entry.
+    fn register_with(
+        &self,
+        source: &dyn Evented,
+        interest: Interest,
+        readers: Vec<Waker>,
+        writers: Vec<Waker>
+    ) -> io::Result<Arc<Entry>> {
         let mut entries = self.entries.lock().unwrap();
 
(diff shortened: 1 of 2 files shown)


tomaka:
@yoshuawuyts @stjepang Would it be possible to release a version including this in the not-too-far future? (it doesn't need to be in the next hours/days, but not too far away either if possible)

In Polkadot we're sometimes reaching a limit to the number of threads using async-std 1.4, which causes a panic in `spawn_blocking`. We did an "emergency" PR https://github.com/paritytech/polkadot/pull/810 that uses async-std master and which works fine.

yoshuawuyts:
@tomaka Ah yeah, for sure. Going to make some time next week to issue a new release.

PR #680, opened 2020-01-16
Some modules support no_std

k-nasa:
The following three modules have been changed to work with no_std.

- task
- future
- stream

PR #696, opened 2020-02-04
Some functions in future module to no_std

k-nasa:
Wait to merge this PR.
https://github.com/async-rs/async-std/pull/680


src/future/poll_fn.rs
```diff
@@ -23,15 +23,18 @@ use crate::task::{Context, Poll};
 /// #
 /// # })
 /// ```
-pub async fn poll_fn<F, T>(f: F) -> T
```
yjhmelody: Why not use `async` syntax ?

taiki-e: the current async-await requires `std`, so cannot use in no-std environment

src/future/poll_fn.rs
```diff
@@ -23,15 +23,18 @@ use crate::task::{Context, Poll};
...
+pub fn poll_fn<F, T>(f: F) -> PollFn<F>
```
yoshuawuyts: Hmm, I'm so-so about this change. Functions such as [`io::copy`](https://docs.rs/async-std/1.5.0/async_std/io/fn.copy.html) don't work on no-std either, but would be really useful to have as well. Having these functions be `async fn` was a bit of a deliberate choice when designing async-std.

With [async await for embedded rust](https://ferrous-systems.com/blog/async-on-embedded/) steadily progressing, I wonder how needed this is. It seems as async/await continues to move beyond mvp status this will be much of a non-issue.

All in all I think it'd be okay to leave this method as-is for now, and we can merge all the other changes. Thanks!

src/future/mod.rs
```diff
@@ -48,17 +48,14 @@
...
-}
-
-cfg_std! {
+    pub use ready::ready;
     pub use pending::pending;
     pub use poll_fn::poll_fn;
-    pub use ready::ready;
 
+    pub(crate) mod future;
+    mod ready;
```
yoshuawuyts: I believe `ready` should be able to work in no-std.

src/future/poll_fn.rs
skade: Hm, the feature won't make the cut in 6 days, so it's at least 18 weeks out. Given that this is no user-visible change, I'd be fine with enabling embrio users to use this currently.

k-nasa: Since the interface seen from the user does not change, I think that it is OK to change this.

PR #697, opened 2020-02-04
Document the core feature

yoshuawuyts:
Follow-up to https://github.com/async-rs/async-std/pull/680, documents how to enable the feature. Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -220,6 +220,16 @@
 //! default-features = false
 //! features = ["std"]
 //! ```
+//! 
+//! And to use async-std on `no_std` targets that only support `alloc` only
+//! enable the `core` Cargo feature:
+//!
+//! ```toml
+//! [dependencies.async-std]
+//! version = "1.0.0"
+//! default-features = false
+//! features = ["core"]
+//! ```
 
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(feature = "docs", feature(doc_cfg))]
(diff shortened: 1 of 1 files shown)


taiki-e:
the feature name is `alloc`

src/lib.rs
```diff
@@ -220,6 +220,16 @@
...
 //! ```
+//! 
+//! And to use async-std on `no_std` targets that only support `alloc` only
+//! enable the `core` Cargo feature:
+//!
+//! ```toml
+//! [dependencies.async-std]
+//! version = "1.0.0"
+//! default-features = false
+//! features = ["core"]
```
k-nasa:
```suggestion
//! features = ["alloc"]
```

src/lib.rs
```diff
@@ -220,6 +220,16 @@
 //! default-features = false
 //! features = ["std"]
 //! ```
+//! 
+//! And to use async-std on `no_std` targets that only support `alloc` only
+//! enable the `core` Cargo feature:
+//!
+//! ```toml
+//! [dependencies.async-std]
+//! version = "1.0.0"
```
k-nasa: There is no `alloc` in this version. (Not in latest 1.5.0)
I think this statement is misleading because it can be used in 1.0.0.

src/lib.rs
```diff
@@ -220,6 +220,16 @@
 //! default-features = false
 //! features = ["std"]
 //! ```
+//! 
+//! And to use async-std on `no_std` targets that only support `alloc` only
+//! enable the `core` Cargo feature:
```
yoshuawuyts:
```suggestion
//! enable the `alloc` Cargo feature:
```

src/lib.rs
yoshuawuyts:
```suggestion
//! version = "1.5.0"
```

PR #703, opened 2020-02-12
channel/recv: improving function docs and code example

spacekookie:
At the moment it's not clear when and why recv returns Option<T>,
instead of just T. This changed comment makes it clear that None will
only be returned once no data will ever be sent again (i.e. after all
senders are gone).


yoshuawuyts:
Related: https://github.com/async-rs/async-std/pull/585

PR #715, opened 2020-02-24
Enforce non zero capacity for channel

### LINKED ISSUE #712: "capacity must be positive" error when sending between tasks
I'm hitting a `capacity must be positive` error when sending between two tasks and I'm a bit unsure what could cause this I think it's a condition that should never be hit?

Licenser:
This aligns the function with it's documentation and will result in easier to resolve errors for problems such as #712. 

closes #712

### DIFF
--- a/src/sync/channel.rs
+++ b/src/sync/channel.rs
@@ -58,7 +58,7 @@ use crate::sync::WakerSet;
 #[cfg(feature = "unstable")]
 #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
 pub fn channel<T>(cap: usize) -> (Sender<T>, Receiver<T>) {
-    assert!(cap >= 1);
+    debug_assert!(cap >= 1);
     let channel = Arc::new(Channel::with_capacity(cap));
     let s = Sender {
         channel: channel.clone(),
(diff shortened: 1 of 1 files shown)


Licenser:
this fails in quite a few places but I don't think those failures are related to my changes

src/sync/channel.rs
```diff
@@ -58,6 +58,7 @@ use crate::sync::WakerSet;
 #[cfg(feature = "unstable")]
 #[cfg_attr(feature = "docs", doc(cfg(unstable)))]
 pub fn channel<T>(cap: usize) -> (Sender<T>, Receiver<T>) {
+    assert!(cap >= 1);
```
yoshuawuyts: Can we change this into a `debug_assert`?
```suggestion
    debug_assert!(cap >= 1);
```

Licenser: We can, the goal here was to pull this assertation:
https://github.com/async-rs/async-std/blob/master/src/sync/channel.rs#L638 forward to the 'offending' function so that it's easier to track the issue.

Licenser:
I've been doing some thinking and perhaps instead of guarding against zero we could use the type checker to make sure that we never get zero by using [`NonZeroUsize`](https://doc.rust-lang.org/std/num/struct.NonZeroUsize.html) what do you think?

Licenser:
ah strike that, it requires to be manually wrapped, however, it could be used for the internal functions? as in the Channel struct

k-nasa:
> this fails in quite a few places but I don't think those failures are related to my changes

I think it can be fixed by rebase master.

Licenser:
Sorry for the late reply / rebase, got dragged into $otherrthings :tm:

PR #721, opened 2020-03-08
update dependence crates

k-nasa:


Cargo.toml
```diff
@@ -54,31 +54,31 @@ alloc = [
...
+once_cell = { version = "1.3.1", optional = true }
+pin-project-lite = { version = "0.1.4", optional = true }
 pin-utils = { version = "0.1.0-alpha.4", optional = true }
 slab = { version = "0.4.2", optional = true }
 
 [dev-dependencies]
 femme = "1.3.0"
 rand = "0.7.3"
-surf = "1.0.3"
+surf = "2.0.0-alpha.0"
```
taiki-e: I'm not sure it is good to depend on pre-release crates. They sometimes break...

k-nasa:
```suggestion
surf = "1.0.3"
```

PR #722, opened 2020-03-08
some stream module to no-std

k-nasa:


dignifiedquire:
Conflict with upstream

k-nasa:
Close by https://github.com/async-rs/async-std/pull/727

PR #723, opened 2020-03-09
feat: initial implementation of std::process

dignifiedquire:
- API is built to match `std::process` as close as possible for now
- Implementation details are built based on https://github.com/alexcrichton/tokio-process

## Outstanding work

- [ ] Review `unsafe` parts in more detail
- [ ] Add documentation examples
- [ ] Port the tests from stdlib over
- [ ] Test on Windows
- [ ] Document original authorship (https://github.com/async-rs/async-std/pull/723#issuecomment-596613123)

If anyone is interested in helping to bring this over the line that would be great, just let me know.

## Example

```rust
use async_std::process::Command;

let output = Command::new("echo")
                      .arg("Hello world")
                      .output()
                      .await
                      .expect("Failed to execute command");

assert_eq!(b"Hello world\n", output.stdout.as_slice());
```

Implements #22


skade:
> Implementation details are built based on https://github.com/alexcrichton/tokio-process

Note that this means we should at some point document the original authorship and credit to Alex Crichton and Ivan Petkov: https://github.com/alexcrichton/tokio-process/blob/master/Cargo.toml#L8

dignifiedquire:
> Note that this means we should at some point document the original authorship and credit to Alex Crichton and Ivan Petkov: [alexcrichton/tokio-process:Cargo.toml@`master`#L8](https://github.com/alexcrichton/tokio-process/blob/master/Cargo.toml#L8) [](/alexcrichton/tokio-process/blob/HEAD@{2020-03-09T15:51:05Z}/Cargo.toml#L8)

Agreed, any thoughts on how to best do that?

skade:
Either on a comment in the main module or an addendum to our LICENSE file. We could also add a section in our in-project documentation.

fenhl:
Has there been a decision on forgetting vs killing the process on drop? I see that the documentation implies kill on drop but references `Child::forget` which does not exist, so I'm wondering how accurate that is.

PR #727, opened 2020-03-12
Revert "Stabilize most stream method and remove unnecessary macros"

k-nasa:
Reverts async-rs/async-std#719

## WHY
https://github.com/async-rs/async-std/issues/711#issuecomment-596848537

I merged a PR. But it has still have some concerns .
I think it would be better to revert.

thx.

cc @taiki-e @yoshuawuyts @skade

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ default = [
   "async-task",
   "crossbeam-channel",
   "crossbeam-deque",
+  "futures-timer",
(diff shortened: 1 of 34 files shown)


yoshuawuyts:
@k-nasa This patch also undoes the CI fixes. Could you perhaps edit it so the fixes remain in place, but the stabilization of the methods is reverted? Thanks!

PR #729, opened 2020-03-18
Fix doc test

k-nasa:
- Fix lib.rs doc test
- Run ignored stream test
- Fix `stream::min` and `stream::max`


src/stream/stream/max.rs
```diff
@@ -11,29 +10,23 @@ use crate::task::{Context, Poll};
 pin_project! {
     #[doc(hidden)]
     #[allow(missing_debug_implementations)]
-    pub struct MaxFuture<S, F, T> {
+    pub struct MaxFuture<S, T> {
```
dignifiedquire, line 13: this seems unrelated, why this change?

k-nasa: I think it is unnecessary type argument. 
Certainly that may be different from what  I'm doing with this PR.

k-nasa:
If pass ci, I merge it.

PR #733, opened 2020-03-20
New scheduler

k-nasa:
This PR is a successor to #631 .

This PR contains only the following parts:
> This PR replaces the current scheduler with a new one. Differences:
> - The executor and reactor are more closely coupled. We don't have a networking driver thread anymore. Instead, executor threads poll mio.

What we did in this PR

- Removed the mechanism to launch a new thread.
- Revived spawn_blocking


src/rt/runtime.rs
```diff
@@ -0,0 +1,342 @@
...
+}
+
+/// A thread running a processor.
+struct Machine {
+    /// Holds the processor until it gets stolen.
+    processor: Processor,
+}
+
+unsafe impl Send for Machine {}
+unsafe impl Sync for Machine {}
```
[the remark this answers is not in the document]
yoshuawuyts: I'm actually a little bit nervous that this wasn't caught by a test. Not sure what to do about it though.

src/rt/runtime.rs
```diff
@@ -102,74 +88,21 @@ impl Runtime {
     /// Runs the runtime on the current thread.
     pub fn run(&self) {
         scope(|s| {
-            let mut idle = 0;
-            let mut delay = 0;
-
-            loop {
```
yoshuawuyts: Why was this loop removed? Doesn't that mean that once we've iterated over all machines once, the loop exits? If this is indeed a bug, we should have a test that catches this.

src/rt/runtime.rs
```diff
@@ -294,13 +204,11 @@ impl Machine {
...
                 rt.quick_poll().unwrap();
 
-                if let Some(p) = self.processor.lock().as_mut() {
-                    if let Steal::Success(task) = p.steal_from_global(rt) {
-                        p.schedule(rt, task);
-                    }
-
-                    p.flush_slot(rt);
+                if let Steal::Success(task) = self.processor.steal_from_global(rt) {
+                    self.processor.schedule(rt, task);
```
yoshuawuyts: I never completely understood the proposed version of the reactor, so I don't know whether this change is accurate. Can you maybe share why the processor lock is no longer needed? I'm trying to understand better what's going on.

k-nasa: Oh, It is I mistaked.
@stjepang created scheduler need to transfer machine's have processor. Because the number of threads changes.
https://github.com/async-rs/async-std/pull/631

However, the mechanism for changing the number of threads has been removed.So I thought I didn't need to lock the processor.

But, Worker haven't implemented Sync trait. Therefore, it was necessary to acquire a lock in order to handle the Processor in a thread-safe manner.

```
struct Processor {
    /// The local task queue.
    /// Not implemented Sync
    worker: Worker<Runnable>,

    /// Contains the next task to run as an optimization that skips the queue.
    slot: Cell<Option<Runnable>>,
}
```

I fixed it 3229111.

src/rt/runtime.rs
k-nasa: Once we started the task processing thread, we thought there was no runtime work.
Now the scheduler is like that.

https://github.com/async-rs/async-std/blob/2dbebe54ede4d2c0a18380f51e785d5306022ac5/src/task/executor/pool.rs#L31

src/rt/runtime.rs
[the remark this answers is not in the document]
k-nasa: Relate to this.
https://github.com/async-rs/async-std/pull/733#discussion_r396076621

yoshuawuyts:
@k-nasa overall this looks good; I can't spot any bugs through review. But still, the fact that a few bugs slipped through CI on an earlier round makes me feel a bit nervous about merging it as-is; I didn't write this part of the code, and don't trust myself to do a full review.

What would help here is if we could add more tests -- at least some we *know* would have failed if we hadn't caught them during review. Having those would provide more confidence that these changes actually work as intended. Thanks!

dignifiedquire:
I think asking folks to test this branch in larger apps or libraries will also help to find bugs and build confidence

k-nasa:
@yoshuawuyts @dignifiedquire What about publishing in version 1.6.0-alpha?

dignifiedquire:
@k-nasa yes, that is the current plan. Cutting an alpha prerelease with this and get folks testing.

dignifiedquire:
This has been merged to master, we will be releasing a `1.6.0-alpha` soon, so please start using either in your projects to test this well.

PR #735, opened 2020-03-21
ci: speed up github actions

k-nasa:


.github/workflows/ci.yml
```diff
@@ -117,15 +129,3 @@ jobs:
 
     - name: Docs
       run: cargo doc --features docs
-
```
dignifiedquire, line 120: why remove clippy?

k-nasa: I don't think there is any point in leaving this comment out. If you need a clippy later, you can just add it again.

dignifiedquire: right, makes sense

dignifiedquire:
Thanks

PR #739, opened 2020-03-30
Update futures-timer to 3.0.2

devashishdxt:


devashishdxt:
I didn't check before creating a PR. Another PR for this is already open https://github.com/async-rs/async-std/pull/738

src/stream/interval.rs
```diff
@@ -71,125 +71,8 @@ impl Stream for Interval {
...
-    let mult = spent_ns / interval_ns + 1;
-    assert!(
-        mult < (1 << 32),
-        "can't skip more than 4 billion intervals of {:?} \
-         (trying to skip {})",
-        interval,
-        mult
-    );
-    prev + interval * (mult as u32)
-}
```
k-nasa: I don't understand this logic, but is it safe to remove it?

devashishdxt: Yes. The logic is complex and there is not much documentation for `next_interval` function. But, as the name suggests, it is trying to find the next `Instant` to update the `Delay` with. As the new version of `Delay` already takes `Duration` as argument, we don't have to go through that complex logic to find the next `Instant`.

Issue #740, opened 2020-04-07
blocking on spawn not work as intended


win-t:
in the [blog] it is said that:
> The runtime measures the time it takes to perform the blocking operation and if it takes a while, a new thread is automatically spawned and replaces the old executor thread. That way, only the current task is blocked on the operation rather than the whole executor thread or the whole runtime. If the blocking operation is quick, we don’t spawn a new thread and therefore no additional cost is inflicted.

so I tried to block all available thread (my machine has 4 core), but after a while, I don't see any new thread is spawned

```rust
use async_std::task; // version 1.5.0
use std::time::Duration;

fn main() {
    task::block_on(async {
        let num = num_cpus::get();
        let mut handlers = Vec::with_capacity(num * 2);

        for i in 1..=num {
            handlers.push(task::spawn(blocking_work(i)));
            handlers.push(task::spawn(normal_work(i)));
        }

        for h in handlers.into_iter() {
            h.await
        }
    })
}

async fn blocking_work(i: usize) {
    loop {
        std::thread::sleep(Duration::from_secs(5));
        println!("blocking_work {}", i)
    }
}

async fn normal_work(i: usize) {
    loop {
        task::sleep(Duration::from_secs(1)).await;
        println!("normal_work {}", i)
    }
}
```

shouldn't runtime spawn more OS thread for this?

[blog]: https://async.rs/blog/stop-worrying-about-blocking-the-new-async-std-runtime/

win-t:
more infomation

no `normal_work` is printed
![image](https://user-images.githubusercontent.com/1589120/78640008-2ad41580-78d9-11ea-8e40-24f9dd9511ef.png)

no extra thread is spawned (image from htop)
![image](https://user-images.githubusercontent.com/1589120/78640071-4e975b80-78d9-11ea-9967-64209619ce00.png)

win-t:
for comparison, same logic but, with golang (because the blog said the runtime is inspired by Go)

```go
package main

// #include <unistd.h>
import "C"

import (
	"fmt"
	"runtime"
	"sync"
	"time"
)

func main() {
	var wg sync.WaitGroup
	num := runtime.GOMAXPROCS(-1)
	for i := 1; i <= num; i++ {
		wg.Add(1)
		go blockingWork(wg.Done, i)

		wg.Add(1)
		go normalWork(wg.Done, i)
	}
	wg.Wait()
}

func blockingWork(done func(), i int) {
	defer done()
	for {
[...]

PR #744, opened 2020-04-09
auto spawn new worker thread

win-t:
Hi guys, this is my proposal for solving https://github.com/async-rs/async-std/issues/740

what changes I made:
- make the runtime to able scale up dan scale down the number of workers thread
  - Wrap stealers with `RwLock` because we need to able to modify it later
  - instead of creating stealer for each worker in `Runtime::new` I combined it with the creation of the thread into `Runtime::start_new_thread`
  - scaling up is by sending a message to runtime to call `Runtime::start_new_thread`
  - scaling down is by pushing exit message to the global queue, so a random worker will pick it later and terminate itself
- Create a monitor thread
  - this thread will run alongside runtime thread
  - this thread will detect blocking call by creating a task that does an infinite loop, this task is an indicator that runtime is still running if this task failed to run, it means that all worker thread is blocked

btw, I need insight on atomic ordering, I don't understand it well, I don't think SeqCst is needed

I don't know how to test the performance overhead of this solution, so I need advice here

### DIFF
--- a/src/rt/runtime.rs
+++ b/src/rt/runtime.rs
@@ -2,12 +2,14 @@ use std::cell::Cell;
 use std::io;
 use std::iter;
 use std::sync::atomic::{self, Ordering};
-use std::sync::{Arc, Mutex};
+use std::sync::atomic::{AtomicBool, AtomicUsize};
+use std::sync::{Arc, Mutex, RwLock};
 use std::thread;
 use std::time::Duration;
 
+use crossbeam_channel::{unbounded, Receiver, Sender};
 use crossbeam_deque::{Injector, Steal, Stealer, Worker};
-use crossbeam_utils::thread::scope;
+use crossbeam_utils::thread::{scope, Scope};
 use once_cell::unsync::OnceCell;
 
 use crate::rt::Reactor;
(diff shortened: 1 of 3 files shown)


src/rt/runtime.rs
```diff
@@ -88,32 +103,88 @@ impl Runtime {
...
                 }
-                Some(m) => m.schedule(&self, task),
+                Some(m) => m.schedule(&self, Task::Runnable(task)),
             }
         });
     }
 
+    /// Start a worker thread.
+    fn start_new_thread<'e, 's: 'e>(&'s self, scope: &Scope<'e>) {
+        let id = self.counter.fetch_add(1, Ordering::SeqCst /* ??? */);
```
win-t, line 116: here

win-t:
adding the capability for the users to set their own monitoring policy, so they can running the monitor to meet specific performance requirement.

skade:
@win-t I would not want to expose an additional public module.

win-t:
yeah, @skade I think so, introducing public API was not my original proposal, let's do it step by step,
for now, just focus on how to handle the blocking part,

I still have doubt about the action: scaling up the runtime or just abort the whole program

that doubt drive me to think that the user should decide it by themself by exposing via new API, but I don't think that is also a good idea,

win-t:
I decide to just scale up the runtime, if the users want to abort, they can create their own monitor thread

src/rt/runtime.rs
```diff
@@ -140,23 +217,34 @@ impl Runtime {
...
+            id,
+            drained: AtomicBool::new(false),
         }
     }
 
     /// Schedules a task onto the machine.
-    fn schedule(&self, rt: &Runtime, task: Runnable) {
-        self.processor.lock().schedule(rt, task);
+    fn schedule(&self, rt: &Runtime, task: Task) {
+        if !self.drained.load(Ordering::SeqCst /* ??? */) {
```
win-t, line 237: this is hot path, and I don't understand what SeqCst mean, maybe other values give better performance

win-t:
Just hold this PR for now until https://github.com/async-rs/async-std/pull/748 is resolved

PR #747, opened 2020-04-09
Fix new scheduler loop

### LINKED ISSUE #746: Performance regression with new scheduler
It turns out that #733 introduced quite the performance regression on various http benchmarks, > 50% on https://github.com/stjepang/minihttp/

I have tracked the issue down to this commit: https://github.com/async-rs/async-std/commit/f9607768461242b129454256ec4528c680c9a946

If I revert it, all tests still pass and performance improves to be even slightly faster than 1.5. 

@k-nasa could you shed some light what exactly you were trying to achieve in this commit and why you introduced it?

dignifiedquire:
This now matches more closely the logic as implemented in #631, and fixes the performance regression as far as I have observed.

Closes #746

### DIFF
--- a/src/rt/runtime.rs
+++ b/src/rt/runtime.rs
@@ -249,6 +249,8 @@ impl Machine {
                 continue;
             }
 
+            let mut sched = rt.sched.lock().unwrap();
+
             // One final check for available tasks while the scheduler is locked.
             if let Some(task) = iter::repeat_with(|| self.find_task(rt))
                 .find(|s| !s.is_retry())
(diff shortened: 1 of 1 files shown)


src/rt/runtime.rs
```diff
@@ -258,19 +260,19 @@ impl Machine {
...
 
-            let mut sched = rt.sched.lock().unwrap();
-
+            // If another thread is already blocked on the reactor, there is no point in keeping
+            // the current thread around since there is too little work to do.
             if sched.polling {
-                thread::sleep(Duration::from_micros(10));
-                continue;
+                break;
             }
```
win-t, line 267: I'm not sure, terminating the thread is a good idea, what will happen when the load is increasing? no new thread will be created

simple sleep like this will end up with single thread
```rust
        task::spawn(async {
            loop {
                task::sleep(Duration::from_secs(1)).await;
                io::stdout().write_all(b"Hello World\n").await.unwrap();
            }
        })
        .await;
```

win-t: I want to bring back auto spawning thread here https://github.com/async-rs/async-std/issues/740 , but it's just as a fallback if the users forget to use `spawn_blocking` (or especially when they don't own full control of the code, like using a library)

dignifiedquire: We can not go back to the sleep & continue, it regresses performance even further, so we need to find a better solution.

dignifiedquire: The correct fix as far as I understand is to bring back the dynamic restarting of machines: https://github.com/async-rs/async-std/pull/748

PR #748, opened 2020-04-10
fix(rt): bring back dynamic machines

dignifiedquire:
Even if we do not make use of the progress blocking, we do need to make use of the dynamic restarting of machines as far as I understand.

Keeps the perf, while removing the regression from #747

### DIFF
--- a/src/rt/runtime.rs
+++ b/src/rt/runtime.rs
@@ -1,6 +1,7 @@
 use std::cell::Cell;
 use std::io;
 use std::iter;
+use std::ptr;
 use std::sync::atomic::{self, Ordering};
 use std::sync::{Arc, Mutex};
 use std::thread;
@@ -26,6 +27,12 @@ thread_local! {
 struct Scheduler {
     /// Set to `true` while a machine is polling the reactor.
     polling: bool,
+
+    /// Idle processors.
+    processors: Vec<Processor>,
+
+    /// Running machines.
+    machines: Vec<Arc<Machine>>,
 }
 
 /// An async runtime.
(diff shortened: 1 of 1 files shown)


src/rt/runtime.rs
```diff
@@ -99,21 +97,57 @@ impl Runtime {
...
+                            })
                         })
-                    })
-                    .expect("cannot start a machine thread");
+                        .expect("cannot start a machine thread");
+                }
+
+                // Sleep for a bit longer if the scheduler state hasn't changed in a while.
+                if idle > 10 {
+                    delay = (delay * 2).min(10_000);
```
yoshuawuyts, line 121: Is perhaps crossbeam utils BackOff a better fit here than rolling it by hand?

k-nasa:
@dignifiedquire 
What is the cost of starting a thread?
The reason I removed the mechanism to increase or decrease the number of threads is to temporarily omit those discussions.

I think it's good to start a thread when you need it and close it when it's no longer needed.

win-t:
> @dignifiedquire
> What is the cost of starting a thread?
> The reason I removed the mechanism to increase or decrease the number of threads is to temporarily omit those discussions.
> 
> I think it's good to start a thread when you need it and close it when it's no longer needed.

can we cache the thread? make the thread is just park instead of terminating it, then when a new thread is needed just unpark it and assume this thread is a fresh one and make sure no resource leak to the new thread (reinitialize all  TLS)

dignifiedquire:
I agree we should investigate parking, but for now this is the original logic and fixes the brokeness on master

src/rt/runtime.rs
dignifiedquire: I figured if it was good enough for @stjepang it'll be good enough for me for now ;)

Issue #664, opened 2020-01-08
Build error on MIPS 32bit


faulesocke:
```
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
 --> [...]/async-std-1.4.0/src/task/task_id.rs:2:25
  |
2 | use std::sync::atomic::{AtomicU64, Ordering};
  |                         ^^^^^^^^^
  |                         |
  |                         no `AtomicU64` in `sync::atomic`
  |                         help: a similar name exists in the module: `AtomicU8
```

Unfortunately, according to the docs, 64bit atomic types are not available on 32bit mips architectures.

I've looked into the source location and two solutions to this come to my mind:

1. use AtomicU32. This might or might not be okay on 32bit architectures.
2. Rewrite this (for mips only) with blocking logic (i.e. keep the counter inside a mutex)

If someone tells me which one is the preferred solution, I can write a PR.

yoshuawuyts:
Related: https://github.com/async-rs/async-std/pull/286

faulesocke:
Yep, this looks like it will solve my problem. Any estimations on when it might be ready?

yoshuawuyts:
@faulesocke this will mostly depend on availability of our team; we'll be sitting down to plan next steps for async-std next week, I'll be sure to bring this up.

faulesocke:
It looks like #286 was closed without being merged. Are there other plans to solve this?

PR #756, opened 2020-04-26
Test and fix 32 bit targets

dignifiedquire:
Based on #286 

Closes #141,#664


dignifiedquire:
Due to the fact that `Mutex::new` is not a `const` function, the fallback failed for now. Switching to `AtomicUsize`. If someone has a better fix please PR it :) But this at least unblocks these systems for a while

bltavares:
❤️ This changeset allowed me to compile async-std on MIPS (should close https://github.com/async-rs/async-std/issues/664 also)

Issue #766, opened 2020-05-08
async-std@1.6.0-beta.1: cannot run an executor inside another executor 


GopherJ:
after upgrading to 1.6.0-beta.1, I start to get this error message. Previous version: 1.5.0

dignifiedquire:
This sounds like you are running `block_on` inside another `block_on`, which was not really well supported before, but is now being detected and not ignored anymore.

GopherJ:
@dignifiedquire  I removed `block_on` inside block on but all my tests still fail, could you please help having a look, I would like to test the new version because maybe wasm compiles.

https://github.com/GopherJ/casbin-rs/tree/upgrade-async-std

GopherJ:
Also this error is a runtime error which makes it harder to identify during compile time, is it possible to locate it during compile time?

andysalerno:
I'm receiving the same error.  At first it was throwing from a `block_on`, so I removed the `block_on` and have none remaining in the codebase.

From inspecting the trace, it looks like there's a `block_on` in async-std-1.6.0-beta.1/src/fs/file.rs:318.

In my code, the `File` is going out of scope, I guess invoking the `drop()` of the `File`, which leads to a stack trace like this, which I guess involves a block-in-a-block (since my `main()` is decorated with the `async_std::main` attribute):

```
  21: async_std::task::block_on::block_on
             at /home/andy/.cargo/registry/src/github.com-1ecc6299db9ec823/async-std-1.6.0-beta.1/src/task/block_on.rs:33
  22: <async_std::fs::file::File as core::ops::drop::Drop>::drop
             at /home/andy/.cargo/registry/src/github.com-1ecc6299db9ec823/async-std-1.6.0-beta.1/src/fs/file.rs:318
  23: core::ptr::drop_in_place
             at /rustc/8d69840ab92ea7f4d323420088dd8c9775f180cd/src/libcore/ptr/mod.rs:177
  24: core::ptr::drop_in_place
             at /rustc/8d69840ab92ea7f4d323420088dd8c9775f180cd/src/libcore/ptr/mod.rs:177
  25: toygrep::buffer::async_line_buffer::AsyncLineBufferReader<R>::take_line_buffer

```

https://github.com/async-rs/async-std/blob/master/src/fs/file.rs#L318

andysalerno:
Did a quick test, this is failing:
```rust
#[async_std::main]
async fn main() {
    async_std::task::block_on(async move {
        println!("Hello, world!");
    });
[...]

PR #768, opened 2020-05-09
fix(fs): use smol::block_on for drop handling of File

dignifiedquire:
This avoids nesting of `task::block_on` calls. 

Ref #766


GopherJ:
I'd like to suggest publishing:  `1.6.0-beta.2` so that more people can test

PR #774, opened 2020-05-11
feat: add env vars to configure the runtime threadpool size and name

jbr:
I read the thread at #690 and this seemed like an easy change to make.

I tried to find a good balance between accurate-but-verbose (ASYNC_STD_RT_POOL_SIZE, ASYNC_STD_RT_THREAD_NAME) and concise-but-generic (RT_THREADS, RT_NAME).

Would people want to be able to optionally include the thread number in the thread name? I'm not certain why they would want to, but I believe I've seen that in other threadpools.  If so, the thread name variable might include some substitution placeholder.

# docs:
image

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -230,6 +230,21 @@
 //! default-features = false
 //! features = ["alloc"]
 //! ```
+//!
+//! # Runtime configuration
+//!
+//! Several environment variables are available to tune the async-std
+//! runtime:
+//!
+//! * `ASYNC_STD_THREAD_COUNT`: The number of threads that the
+//! async-std runtime will start. By default, this is one per logical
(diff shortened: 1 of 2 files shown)


dignifiedquire:
- I think we would want to document this in the top level docs and the task docs.
- This looks good from my side, but it changes things, such that we now allog configuration, would like to get some feedback from @skade and @yoshuawuyts before merging

jbr:
Added docs and a screenshot.

I was unsure if the documentation should explain the difference between parallelism and concurrency in an async environment, or trust that the reader understand that ASYNC_STD_NUM_THREADS isn't the ceiling on concurrency

dignifiedquire:
Thank you @jbr lets get this in and see how it works out

PR #776, opened 2020-05-12
Switches `wasm-timer` for `futures-timer`.

### LINKED ISSUE #775: async-std@1.6-beta.1: wasm-timer `Delay` pauses
Heya, e4df140 uses `wasm_timer::Delay` in `src/utils.rs` for WASM target's `Timer` implementation. When running a countdown async task in a loop with a 50 ms delay, it pauses for 4 seconds every 19 iterations (observed in both Chrome and Firefox):

delay_wasm_timer

Switching it to `futures_timer::Delay` allows it to loop smoothly:

delay_futures_timer

#776 has the change, though `futures-timer` depends on `async-std` in `[dev-dependencies]`, so not sure if you'd like the somewhat cyclic reference.

[`futures-timer`](https://crates.io/crates/futures-timer) backs onto [`gloo_timers`](https://crates.io/crates/gloo-timers), which [calls the browser's `setTimeout`](https://github.com/rustwasm/gloo/blob/master/crates/timers/src/callback.rs#L58) API. I couldn't figure out how `wasm-timer` does it.

azriel91:
Fixes #775.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://book.async.rs/overview
 
 ## [Unreleased]
 
+## Changed
+
+- For `wasm`, switched underlying `Timer` implementation to [`futures-timer`](https://github.com/async-rs/futures-timer). ([#776](https://github.com/async-rs/async-std/pull/776))
+
 # [1.6.0-beta.1] - 2020-05-07
 
 ## Added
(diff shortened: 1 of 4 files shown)


azriel91:
Hm, `cargo test --tests` pass locally (doctests have 3 that fail, but they're the same on `master`).
`./wasm-test.sh` doesn't pass locally on `master` as well.

Perhaps I should reduce the Condvar timeout to 50 ms (down from 100)?

dignifiedquire:
Thank you

PR #802, opened 2020-05-29
Fix wrong slice index when reading a file

jerry73204:
As the title, it fixes the wrong slice index of an internal buffer when reading a `File`.


jerry73204:
I found this issue affects many of my applications. If possible, release 1.6.1 along with this and other critical fix-ups?

dignifiedquire:
yes, I'll try and cut a release tomorrow

PR #804, opened 2020-06-03
Add tokio02 feature flag

yoshuawuyts:
Adds the `tokio02` feature flag. This forwards to `smol`'s `tokio02` feature flag which runs `tokio::rt::enter` whenever a new thread is created. This should allow running `async_std::main` and having `tokio` crates _just work_ ™. Thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,6 +110,17 @@ jobs:
         command:  check
         args: --no-default-features --features alloc --target thumbv7m-none-eabi -Z avoid-dev-deps
 
+  check_tokio_02_feature:
+    name: Check tokio02 feature
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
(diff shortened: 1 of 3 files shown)


skade:
This transitively ties us to the currently internal runtime.

yoshuawuyts:
> This transitively ties us to the currently internal runtime and requests it to have tokio support. 

Adding this feature isn't too difficult even if we were to switch internal engines -- the way this works is calling `tokio::rt::enter` every time we call `thread::spawn`. This feature is really useful for projects currently using `tokio` but wanting to switch away from it. It enables switching over one dependency at the time rather than being forced into a "rewrite the world" scenario.

---

@skade could you maybe say more what you mean by: "[...] the expectations from our side top expose such a feature are."

skade:
I was very tired, so sentence makes no sense.

What I wanted to say is that the feature should be specified from the top down instead of bottom up.

PR #807, opened 2020-06-04
Remove stdio lock methods

### LINKED ISSUE #805: StdoutLock is unsound
Run the following program:

```rust
use async_std::io::prelude::WriteExt;
use async_std::{io as async_io, task};
use std::io::{self, Write};
use std::thread;

#[async_std::main]
async fn main() -> io::Result<()> {
    let mut async_locked = async_io::stdout().lock().await;
    let fut = task::spawn_blocking(|| {
        let stdout = io::stdout();
        let mut locked = stdout.lock();
        writeln!(locked, "sync stdout in thread {:?}", thread::current())?;
        locked.flush()?;
        Ok(())
    });
    async_std::writeln!(
        async_locked,
        "async stdout in thread {:?}",
        thread::current(),
    )
    .await?;
    async_locked.flush().await?;
    fut.await
}
```

Cargo.toml dependencies

```toml
[dependencies]
async-std = { version = "1.6.0", features = [ "attributes", "unstable" ] }
```

This program violates the invariant of `StdoutLock` (there must never be two instances of `StdoutLock` owned by different threads), leading to UB and panics like the following.

```rust
sync stdout in thread Thread { id: ThreadId(10), name: None }
thread 'main' panicked at 'already borrowed: BorrowMutError', /rustc/8d69840ab92ea7f4d323420088dd8c9775f180cd/src/libcore/cell.rs:878:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Considering `lock` is an unstable API, I would recommend removing the API, and suggest usage of `spawn_blocking` instead. I don't think it's possible for async `StdoutLock` to ever be sound with threaded executors

xfix:
Fixes #805.

### DIFF
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -328,12 +328,3 @@ cfg_default! {
     #[cfg(not(target_os = "unknown"))]
     mod stdout;
 }
-
-cfg_unstable_default! {
-    #[cfg(not(target_os = "unknown"))]
-    pub use stderr::StderrLock;
-    #[cfg(not(target_os = "unknown"))]
-    pub use stdin::StdinLock;
-    #[cfg(not(target_os = "unknown"))]
-    pub use stdout::StdoutLock;
-}
--- a/src/io/stderr.rs
+++ b/src/io/stderr.rs
@@ -5,11 +5,6 @@ use std::future::Future;
 use crate::io::{self, Write};
 use crate::task::{spawn_blocking, Context, JoinHandle, Poll};
 
-cfg_unstable! {
-    use once_cell::sync::Lazy;
-    use std::io::Write as _;
-}
-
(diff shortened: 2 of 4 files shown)


xfix:
I personally don't think it's possible to have a correct implementation that inter-ops with standard library's locks because of standard library's choice to have a reentrant mutex.

It may be possible to skip the standard library, but I'm not sure about soundness of doing so, even on POSIX platforms. POSIX says the following in [`read` function documentation](https://pubs.opengroup.org/onlinepubs/9699919799/functions/read.html).

> The behavior of multiple concurrent reads on the same pipe, FIFO, or terminal device is unspecified.

The question is, does "unspecified" in POSIX mean UB.

PR #812, opened 2020-06-07
ci: update actions/cache to v2

thibault-martinez:
For some reason, the cargo index was never cached in Github Actions (e.g. https://github.com/async-rs/async-std/runs/746758946?check_suite_focus=true#step:5:7).
We had the same issue and updating https://github.com/actions/cache to [v2](https://github.com/actions/cache/releases/tag/v2.0.0) seems to have fixed it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,19 +30,19 @@ jobs:
         override: true
 
     - name: Cache cargo registry
-      uses: actions/cache@v1
+      uses: actions/cache@v2
       with:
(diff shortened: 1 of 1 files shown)


thibault-martinez:
From a previous run you can see that the index caching was failing https://github.com/async-rs/async-std/runs/746758946?check_suite_focus=true#step:24:5 and is now working fine https://github.com/async-rs/async-std/pull/812/checks?check_run_id=747446195#step:24:3

Issue #307, opened 2019-10-11
Add UdpSocket::PeerAddr


yoshuawuyts:
This was just stabilized in std; we should consider adding it too.

https://github.com/rust-lang/rust/pull/64728

ansrivas:
@yoshuawuyts I quickly looked in the source code, please correct me if I am wrong, this is going to be currently blocked by mio's new release with `peer_addr` implementation right?

yoshuawuyts:
@ansrivas yeah I guess you're right.

zhaxzhax:
I am a beginner in Rust. And I am really delighted to see there is a issue I can fix. I am gonna take this one.

humancalico:
@yoshuawuyts It seems like this can be closed now.

PR #816, opened 2020-06-14
Add UdpSocket::PeerAddr #307

zhaxzhax:
Add UdpSocket::PeerAddr in #307.
The test is passed.


zhaxzhax:
Thanks for your merge. May I ask that could I look at the std-lib in rust, comparing with async-std, and add all the function that in std but not in async-std? I'd like to do this, because I see that async-std is a async version of std. ( Just my thought, I don't know whether it right or wrong).

PR #822, opened 2020-06-21
refactor: switch to async-mutex for Mutex implementation

dignifiedquire:
Another small step in making async-std more modular.


dignifiedquire:
CI failures seem to be an issue in the current nightl: https://github.com/rust-lang/rust/issues/73592

stjepang:
In `tests/timeout.rs`, I suggest bumping `i * 20` to something larger, perhaps `i * 100`, and changing `(0..100)` to `(0..10)`

PR #827, opened 2020-06-27
feat: new channels

dignifiedquire:
- add new top level `channel` module
- reimplement existing channels using `async-channel`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ std = [
   "wasm-bindgen-futures",
   "futures-channel",
   "async-mutex",
+  "async-channel",
 ]
 alloc = [
   "futures-core/alloc",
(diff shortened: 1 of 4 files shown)


stjepang:
Looks like channel is module is in the wrong place: it's in `async_std::channel`, while it should be `async_std::sync::channel`

yoshuawuyts:
> Looks like channel module is in the wrong place: it's in async_std::channel

@stjepang that's the intent; the libs team has been talking about possibly breaking `std::sync` up into smaller bits anyway, and we figured a `channel` impl would be a good candidate for this. But now that you know it's intentional I'm curious to hear what you think!

stjepang:
Ha, interesting! I actually like `channel` as a top-level module and wish `std` did the same.

dignifiedquire:
Going to close this for now. Will rework later.

PR #828, opened 2020-07-01
Updated dependencies

Tron0xHex:


examples/logging.rs
```diff
@@ -3,7 +3,7 @@
 use async_std::task;
 
 fn main() {
-    femme::start(log::LevelFilter::Trace).unwrap();
+    femme::with_level(log::LevelFilter::Trace);
```
otavio, line 6: I'd prefer if it comes together with the `femme` update, otherwise, it is in an unbuildable state.

Cargo.toml
```diff
@@ -50,48 +47,46 @@ std = [
...
 log = { version = "0.4.8", features = ["kv_unstable"], optional = true }
 memchr = { version = "2.3.3", optional = true }
-num_cpus = { version = "1.12.0", optional = true }
-once_cell = { version = "1.3.1", optional = true }
-pin-project-lite = { version = "0.1.4", optional = true }
-pin-utils = { version = "0.1.0-alpha.4", optional = true }
+num_cpus = { version = "1.13.0", optional = true }
+once_cell = { version = "1.4.0", optional = true }
+pin-project-lite = { version = "0.1.7", optional = true }
+pin-utils = { version = "0.1", optional = true }
```
otavio, line 66: There are a number of changes of the versions which are compatible versions. I recently learned about `--skip-compatible` option of `cargo upgrade` and started to use this. It seems like a good way to reduce unnecessary code changes.

PR #830, opened 2020-07-08
Adds example to async_std::task::spawn

tronta:
As I struggled using the function spawn properly (I didn't get that it can be executed without await :-( ), I'd like to add an other example.

### DIFF
--- a/src/task/spawn.rs
+++ b/src/task/spawn.rs
@@ -23,6 +23,30 @@ use crate::task::{Builder, JoinHandle};
 /// #
 /// # })
 /// ```
+///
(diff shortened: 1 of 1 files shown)


tronta:
Yes, the toplevel description. Would have helped me.
In my opinion, more examples are better than less :grinning: 
What do you think, should be modified? Or do you suggest, this is enough...?

yoshuawuyts:
@tronta hah, yay glad that would've helped -- I would prefer we adapt the docs from `std::thread` for this instead -- putting this behavior into words is probably the right approach here.

Issue #815, opened 2020-06-12
task::spawn_local only available in feature = ["unstable"]


FractalU:
The function 'task::spawn_local' is only behind the 'unstable' feature. However in the docs this function isn't marked as unstable. Is it actually unstable or not?

FractalU:
Here the error message:

   Compiling async_std_unstable v0.1.0 (C:\Users\Richard\Programming\Rust\async_std_unstable)
error[E0432]: unresolved import `async_std::task::spawn_local`
 --> src\main.rs:1:5
  |
1 | use async_std::task::spawn_local;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `spawn_local` in `task`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: could not compile `async_std_unstable`.

To learn more, run the command again with --verbose.

FractalU:
Here the Cargo.toml file:

[package]
............

[dependencies]
async-std = "1.6.1"

FractalU:
<p>Here the relevant portion of the docs:</p>

<table style="display: block;"><tbody><tr class="module-item"><td><a class="fn" href="fn.block_on.html" title="async_std::task::block_on fn">block_on</a></td><td class="docblock-short"><p>Spawns a task and blocks the current thread on its result.</p>
</td></tr><tr class="module-item"><td><a class="fn" href="fn.current.html" title="async_std::task::current fn">current</a></td><td class="docblock-short"><p>Returns a handle to the current task.</p>
</td></tr><tr class="module-item"><td><a class="fn" href="fn.sleep.html" title="async_std::task::sleep fn">sleep</a></td><td class="docblock-short"><p>Sleeps for the specified amount of time.</p>
</td></tr><tr class="module-item"><td><a class="fn" href="fn.spawn.html" title="async_std::task::spawn fn">spawn</a></td><td class="docblock-short"><p>Spawns a task.</p>
</td></tr><tr class="module-item"><td><a class="fn" href="fn.spawn_blocking.html" title="async_std::task::spawn_blocking fn">spawn_blocking</a></td><td class="docblock-short"><span class="stab portability"><code>unstable</code></span><p>Spawns a blocking task.</p>
</td></tr><tr class="module-item"><td><a class="fn" href="fn.spawn_local.html" title="async_std::task::spawn_local fn">spawn_local</a></td><td class="docblock-short"><p>Spawns a task onto the thread-local executor.</p>
[...]

PR #836, opened 2020-07-17
switch to async-io + blocking + multitask

Keruspe:
smol now being divided in smaller crates, I tried porting async-std to those.

The async-io and blocking parts work fine and all tests seem to pass.

As of now, there are some tests hanging, namely because when we spawn a future from inside a block_on, it seems not to be polled, haven't figured out why yet.

The tokio enter function is directly copied from smol

### DIFF
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -148,6 +148,8 @@ cfg_default! {
     mod block_on;
     mod builder;
     mod current;
+    #[cfg(not(target_os = "unknown"))]
+    mod executor;
     mod join_handle;
     mod sleep;
     #[cfg(not(target_os = "unknown"))]
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -149,7 +149,7 @@ cfg_default! {
     mod builder;
     mod current;
     #[cfg(not(target_os = "unknown"))]
-    mod executor;
+    pub(crate) mod executor;
     mod join_handle;
     mod sleep;
     #[cfg(not(target_os = "unknown"))]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,12 @@ jobs:
         command:  check
         args: --features attributes
 
+    - name: build unstable only
+      uses: actions-rs/cargo@v1
+      with:
+        command:  build
+        args: --no-default-features --features unstable
+
     - name: tests
       uses: actions-rs/cargo@v1
       with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ default = [
   "std",
   "async-io",
   "async-task",
+  "blocking",
   "kv-log-macro",
   "log",
   "num_cpus",
@@ -79,6 +80,7 @@ surf = { version = "1.0.3", optional = true }
 
 [target.'cfg(not(target_os = "unknown"))'.dependencies]
 async-io = { version = "0.1.5", optional = true }
+blocking = { version = "0.5.0", optional = true }
 smol = { version = "0.1.17", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ rustdoc-args = ["--cfg", "feature=\"docs\""]
 [features]
 default = [
   "std",
+  "async-io",
   "async-task",
   "kv-log-macro",
   "log",
@@ -77,6 +78,7 @@ futures-timer = { version = "3.0.2", optional = true }
 surf = { version = "1.0.3", optional = true }
 
 [target.'cfg(not(target_os = "unknown"))'.dependencies]
+async-io = { version = "0.1.5", optional = true }
 smol = { version = "0.1.17", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,9 @@ default = [
   "futures-lite",
   "kv-log-macro",
   "log",
+  "multitask",
   "num_cpus",
   "pin-project-lite",
-  "smol",
 ]
 docs = ["attributes", "unstable", "default"]
 unstable = [
@@ -57,7 +57,7 @@ alloc = [
   "futures-core/alloc",
   "pin-project-lite",
 ]
-tokio02 = ["smol/tokio02"]
+tokio02 = ["tokio"]
 
 [dependencies]
 async-attributes = { version = "1.1.1", optional = true }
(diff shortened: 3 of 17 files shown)


Keruspe:
All tests pass here now

Keruspe:
Not sure what's wrong with ppc64 cross compilaion

Keruspe:
#837 just confirmed that the ppc64 failure is unrelated, not sure about the macos one: `(signal: 4, SIGILL: illegal instruction)`

stjepang:
@Keruspe To figure out the crash on macOS, open `.github/workflows/ci.yml` and add `RUST_BACKTRACE: 1` to the `env` section.

Keruspe:
Doesn't seem to give more info, but I think the ci files from master are used. I tried to temporary comment out ppc64 and the ppc64 tests were still run

Keruspe:
I created another project to toy with the CI.

`RUST_BACKTRACE=1` doesn't seem to have an impact of the macOS builder output.

The ubuntu and windows builders pass all the tests just fine. Apart from the ppc64, the other cross compilation targets pass fine too.

I'm not really familiar with debugging macos issues so I don't really know where to start

Keruspe:
The first two commits work fine, the switch to multitask is the one causing the problem on macos

stjepang:
Here are the errors I'm getting on my macOS machine: https://gist.github.com/stjepang/bbe3f07c947b2fd93a47b2e667f4428a

Keruspe:
@stjepang with this last commit, tests pass for OSX.

Looks like `async_io::parking::Parker` calls `Reactor::get()` in its `Drop` impl, which fails on OSX if no IO has been done in the current thread?

Keruspe:
Ok, instead of using the `parking` crate, forcing initialization of the reactor by parking for 0ms seems to work too.

Not sure if there's a more elegant solution, something that we maybe want to fix on the async-io's side of things?

Keruspe:
Just opened https://github.com/stjepang/async-io/pull/4 as an alternative workaround

stjepang:
Thanks for the PR to async-io, this is now fixed in async-io v0.1.4.

Keruspe:
Apart from the ppc64 cross compile which is broken even outside of this PR, everything should work fine now

stjepang:
Looks good to me :) If all goes well, I think it’s time to stabilize crates async-io, blocking, and multitask.

Keruspe:
Rebased on top of master

stjepang:
I think you can now remove `futures-timer` from the `[dependencies]` section in `Cargo.toml`

Keruspe:
It's still used if unstable is enabled and default disabled it seems?

dignifiedquire:
yes it is needed atm when there is no runtime at all imcluded

stjepang:
Right, so we should be able to remove the dependency in the typical case. It pulls in a lot of transitive dependencies...

stjepang:
Right, so we should be able to remove the dependency in the typical case. It pulls in a lot of transitive dependencies...

dignifiedquire:
@stjepang maybe the timers could live in their own crate? that way we could use them in the non runtime build, without pulling in anything else.

stjepang:
@dignifiedquire Timers live in the `async-io` crate. We should exclude that dependency in the non runtime build.

Keruspe:
I don't think there's a way to specify in Cargo.toml that a dep should be there only if an option is unset (wrt futures-timer).

It doesn't seem to bring that much deps when wasm-bindgen isn't enabled though?

Keruspe:
Updated for blocking 0.5

Keruspe:
The nightly CI failures will be fixed with the next nightly: https://github.com/rust-lang/rust/pull/74578

stjepang:
Is the nightly fixed now?

Keruspe:
Should be

Keruspe:
With the switch to async-executor, the code is simpler, we gain more integration with smol 0.3, but the caveat is that tasks spawned onto the multi-threaded global executor cannot run local tasks anymore (since async-executor can only run the multi-threaded executor or the local executor on a thread but not both, so the runtime threads now only runs the global multi-threaded one)

Keruspe:
With something like https://github.com/stjepang/async-executor/pull/3 + reverting the `src/rt/mod.rs` change and dropping `task::executor::run_global` works too and restore the previous behaviour WRT local tasks

stjepang:
What if we restrict `spawn_local()` as suggested here? https://github.com/async-rs/async-std/issues/815#issuecomment-651349994

That would simplify things because only `block_on()` would run a `LocalExecutor`, and spawned threads would only run an `Executor`.

Keruspe:
Then I guess that's more or less what the current port to async-executor does?

Keruspe:
@stjepang What about returning an `Option<Task<T>>` from `Task::local()` instead of panicking? This leaves the choice to the caller to panic or handle the lack of local executor more "softly"

stjepang:
I expect in 99% cases people will just do `Task::local(future).unwrap()`, having the same effect. That said, it would be good to have a way of checking "am I inside an executor?", but I don't think it should block this PR...

src/task/mod.rs
```diff
@@ -148,6 +148,8 @@ cfg_default! {
     mod block_on;
     mod builder;
     mod current;
+    #[cfg(not(target_os = "unknown"))]
```
dignifiedquire, line 151: This also needs to not be enabled when we build without the `default` feature flag, maybe it would be good to introduce an explicit `executor` feature, which groups the deps and this together

Keruspe: Why so? I'm not sure I follow here. All of this module requires a runtime, thus default to be enabled?

Can you point to a code path which would need it?

dignifiedquire: `cargo build --no-default-featurs --features unstable` should work, but I think will not work with the current setup

Keruspe: It already fails on master (#842) and the failures are the same with this PR, it doesn't add more failures

Keruspe: I'll be more than happy to fix things if I'm wrong once master is fixed but I don't think it should block this PR

Keruspe: Build fine now

dignifiedquire: Thank you! Can you do me a favor and add a ci check for this while we are here? otherwise this will break again and again.

Keruspe: Done!

Keruspe:
Anything else holding this PR back?

dignifiedquire:
> Anything else holding this PR back?

- seems CI is sad for some reason
- the code looks good, but I want to give it a test run on a largish codebase I work on to make sure I don't see any surprises

Keruspe:
CI is just windows probably never tested without default and with unstable.

Keruspe:
Squashed those fixes into one commit and rebased

Byron:
Something to be aware of in case you see issues related to Timers: https://github.com/stjepang/async-io/issues/6

zhuxiujia:
> smol now being divided in smaller crates

way smol being divided in smaller crates???

jon-chuang:
Is there discussion on why smol / async-std was preferred to tokio for tremor? I'm trying to understand tremor as well as building some Rust async support for a cloud-native project. The info would be helpful. Thanks.

PR #838, opened 2020-07-18
Streams/Channels documentation for the async-std book

ragona:
Hi friends

This is a first draft of a quick helper section for Streams. I've actually only used Streams in combination with channels, and I wanted to see if I could introduce them independently. I think it's a little odd (and not very concurrent!), and would love any feedback. I'm also not familiar with all the helper methods in `stream` so I would love help adding notes there.

Here are the outstanding items: 

- [ ] Add streams _(Draft)_
- [ ] Add channels
- [ ] Add streams and channels
- [ ] Add links to documentation.
- [ ] Add notes on things like `stream::from_fn` and all the other awesome helper methods.

### DIFF
--- a/docs/src/concepts/streams.md
+++ b/docs/src/concepts/streams.md
@@ -1,1 +1,110 @@
-# TODO: Streams
+# Streams
+
+Streams are a way to evaluate a stream of data; they are the async
+equivalent of iterators.
+
+`async-std` has provided quite a few helper methods to generate streams.
+To get started, check out the `streams` package documentation.
+It contains a bunch of helpful examples for how to create simple streams.
+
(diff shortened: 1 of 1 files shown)


docs/src/concepts/streams.md
```diff
@@ -1 +1,110 @@
...
+If we want to count for less than forever we can use this sort of pattern.
+
+```rust
+for _ in 0..10 {
+	dbg!(counter.next().await);
+}
+```
+
+Or I'm pretty sure there's some way to use the helper methods in `stream`
+to do this, but it's eluding me and I wanted to timebox this little exercise.
```
ragona, line 108: Fix-me!

PR #845, opened 2020-07-31
chore: release v1.6.3

dignifiedquire:

PR #846, opened 2020-08-02
Book: Accept Loop Variable Fixes

ryanbrainard:
In the [example accept loop](https://book.async.rs/patterns/accept-loop.html#handling-errors-manually) in the book, there are a couple of minor variable errors:

1. Should be matching on `result` instead of `stream`.
2. The call to `eprintln!` has a `{}` placeholder, but `e` needs to be passed in.


ryanbrainard:
CI is failing in the same way as described here: https://github.com/async-rs/async-std/pull/845#issuecomment-667351867. This should pass once the root cause of that is fixed.

PR #849, opened 2020-08-11
Fix wrong link in condvar doc

Observer42:


Observer42:
Cause of CI fmt and docs failing: probably the same with https://github.com/async-rs/async-std/pull/845#issuecomment-667351867

Issue #851, opened 2020-08-14
cargo check failing on v1.6.3 tag for wasm32-unknown-unknown


jbr:
see https://github.com/async-rs/async-std/pull/852/checks?check_run_id=983446186

dignifiedquire:
fixed on master

PR #852, opened 2020-08-14
run ci on wasm32-unknown-unknown

jbr:
The test failure on this branch represents an actual build error in 1.6.3 on wasm, breaking any wasm projects that depend on async-std. That build error is filed as #851. This PR tries to avoid this situation in the future by adding wasm to the async-std github CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,26 @@ jobs:
     - name: test
       run: cross test --all --features unstable --target ${{ matrix.target }}
 
+  check_wasm:
+    name: Check wasm targets
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Install nightly with wasm32-unknown-unknown
+      uses: actions-rs/toolchain@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -160,6 +160,26 @@ jobs:
...
       run: cross test --all --features unstable --target ${{ matrix.target }}
 
+  check_wasm:
+    name: Check wasm targets
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Install nightly with wasm32-unknown-unknown
```
dignifiedquire, line 170: probably should use stable for this, there are some current nightlies without wasm32 target support

jbr: addressed in 5342950 by adding a toolchain matrix, but I'm very confused about what happened with the actions. It ran some netlify stuff instead of the file I provided

dignifiedquire: thanks

dignifiedquire:
thanks, merged manually into #863

PR #853, opened 2020-08-15
Add peek{,from} methods to UdpSocket

### LINKED ISSUE #596: Peeking a UDPSocket?
I can use `peek` on a TCPStream to see how many bytes are available to read. I use this to avoid blocking when calling `read`.

With UDPSocket, there doesn't seem to be a way to avoid blocking when using `recv_from` to read data into a buffer. I want something that either reads a UDP packet or immediately returns. Does this exist?

EDIT: `recv` and `recv_from` seem to be exactly the same. `recv` uses `recv_from` in its example.

r3v2d0g:
Fixes #596

### DIFF
--- a/src/net/udp/mod.rs
+++ b/src/net/udp/mod.rs
@@ -206,6 +206,29 @@ impl UdpSocket {
         self.watcher.recv_from(buf).await
     }
 
+    /// Receives data from socket without removing it from the queue.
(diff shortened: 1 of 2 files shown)


yoshuawuyts (approved):
This looks great; thanks so much!

PR #859, opened 2020-08-19
switch from tempdir to tempfile

Keruspe:
Uses a more recent version of rand


Keruspe:
CI failure is spurious

dignifiedquire (approved):

Issue #843, opened 2020-07-25
failed to resolve futures_timer


GopherJ:
```
error[E0433]: failed to resolve: use of undeclared type or module `futures_timer`
  --> /home/cheng/.cargo/registry/src/github.com-1ecc6299db9ec823/async-std-1.6.2/src/utils.rs:84:29
   |
84 |     pub(crate) struct Timer(futures_timer::Delay);
   |                             ^^^^^^^^^^^^^ use of undeclared type or module `futures_timer`

error[E0433]: failed to resolve: use of undeclared type or module `futures_timer`
  --> /home/cheng/.cargo/registry/src/github.com-1ecc6299db9ec823/async-std-1.6.2/src/utils.rs:88:19
   |
88 |             Timer(futures_timer::Delay::new(dur))
   |                   ^^^^^^^^^^^^^ use of undeclared type or module `futures_timer`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0433`.
error: could not compile `async-std`.

To learn more, run the command again with --verbose.

```

with async-std@1.6.2, while run `cargo check` on target `wasm`, it printed this error

dignifiedquire:
fixed on master

PR #863, opened 2020-08-24
fix wasm and nostd builds

### LINKED ISSUE #861: CI check for wasm isn't working
Building async-std for the wasm target is completely broken at the moment:

- https://github.com/async-rs/async-std/issues/851
- https://github.com/async-rs/async-std/issues/843

The reason why it hasn't been detected is that the CI of async-std for wasm isn't working. [Here](https://github.com/async-rs/async-std/runs/1003328709#step:9:1)'s an example.

dignifiedquire:
Closes #861,#843

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,26 @@ jobs:
     - name: test
       run: cross test --all --features unstable --target ${{ matrix.target }}
 
+  check_wasm:
+    name: Check wasm targets
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+
+    - name: Install nightly with wasm32-unknown-unknown
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: nightly
+        target: wasm32-unknown-unknown
+        override: true
+
+    - name: check
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --target wasm32-unknown-unknown
+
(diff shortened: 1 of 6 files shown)


dignifiedquire:
@jbr @yoshuawuyts any ideas about the CI issues, they are quite odd. Suddenly there are missing crates Oo, seems like a CI bug, as this works fine locally.

jbr:
Agreed that it seems to be a CI issue, but I don't think it's related to this PR. See CI failures at https://github.com/async-rs/async-std/pull/853/checks?check_run_id=988260537

Maybe it's related to caching? Does it help to temporarily comment out L32-48 of ci.yml?

jbr:
@dignifiedquire See the above PR for a possible long-term solution to the wasm/no-std problem, but at the cost of requiring all wasm users to enable a `wasm` feature, which would be a breaking change. I have no idea how to get both no-std and wasm compiling simultaneously without that, though, and don't understand how it ever worked

yoshuawuyts (approved):

Fishrock123:
This seems fine to me, although me and @jbr were discussing with @yoshuawuyts last week and Yosh noted that just getting it to build may not be enough and actual was builds could still have issues when run in their environments.

I'm not so invested in that personally but is something that may be good to track, both for async-std and http-rs.

dignifiedquire:
yes we need to run the tests as well, I had wasm-bindgen-test working at some point on my machine, but forgot the magical invocations..

Issue #823, opened 2020-06-24
async std unstable feature compiles wasm crates on non-wasm arch


jbr:
## Versions:
My platform:`x86_64-apple-darwin`
Async-std version: 1.6.2
rust version: both `rustc 1.44.1 (c7087fe00 2020-06-17)` and `rustc 1.46.0-nightly (a8cf39911 2020-06-21)`

## Issue:
This line https://github.com/async-rs/async-std/blob/17ab958ac20537f00a180772d08378619bca2782/Cargo.toml#L81 activates if the `unstable` feature is enabled regardless of the target_arch. This means that crates that use the `unstable` feature such as http-rs/http-types and http-rs/tide have to compile `web-sys` and `js-sys`.

These are the changes to `cargo tree --features unstable` when commenting out that line, which _shouldn't_ be activated on `x86_64-apple-darwin`. I don't assume that the fix is actually to comment this line out. Is this a cargo bug?

```diff
--- master
+++ without line 81
@@ -8,47 +8,9 @@
 ├── futures-core v0.3.5
 ├── futures-io v0.3.5
 ├── futures-timer v3.0.2
-│   ├── gloo-timers v0.2.1
-│   │   ├── futures-channel v0.3.5
-│   │   │   ├── futures-core v0.3.5
-│   │   │   └── futures-sink v0.3.5
-│   │   ├── futures-core v0.3.5
-│   │   ├── js-sys v0.3.40
-│   │   │   └── wasm-bindgen v0.2.63
-│   │   │       ├── cfg-if v0.1.10
-│   │   │       ├── serde v1.0.114
-│   │   │       ├── serde_json v1.0.55
-│   │   │       │   ├── itoa v0.4.6
-│   │   │       │   ├── ryu v1.0.5
-│   │   │       │   └── serde v1.0.114
-│   │   │       └── wasm-bindgen-macro v0.2.63
-│   │   │           ├── quote v1.0.7
-│   │   │           │   └── proc-macro2 v1.0.18
-│   │   │           │       └── unicode-xid v0.2.0
-│   │   │           └── wasm-bindgen-macro-support v0.2.63
-│   │   │               ├── proc-macro2 v1.0.18 (*)
-│   │   │               ├── quote v1.0.7 (*)
-│   │   │               ├── syn v1.0.33
-│   │   │               │   ├── proc-macro2 v1.0.18 (*)
-│   │   │               │   ├── quote v1.0.7 (*)
[...]

PR #864, opened 2020-08-25
fix wasm + no-std cargo interactions

jbr:
This started as an extension of #863 disabling the build cache in ci and turned into an investigation of a cargo bug that pulls in wasm deps for no-std, breaking ci. The proposed fix in this pr also closes https://github.com/async-rs/async-std/issues/823

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,14 +41,6 @@ jobs:
         command:  check
         args: --features unstable --all --bins --examples --tests
 
-    - name: check wasm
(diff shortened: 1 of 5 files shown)


jbr:
@dignifiedquire It seems like this sledgehammer-fix resolves the macos issue, but I think the no-std failure is real. I get it locally as well. Does futures-core/alloc require std?

dignifiedquire:
Got it back working without introducing a feature, so closing this https://github.com/async-rs/async-std/pull/863

PR #866, opened 2020-08-27
update smol dependencies

Keruspe:
Updates all of the subcomponents of smol that we're now using to their latest versions

cc @stjepang

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,9 +80,9 @@ futures-timer = { version = "3.0.2", optional = true }
 surf = { version = "1.0.3", optional = true }
 
 [target.'cfg(not(target_os = "unknown"))'.dependencies]
-async-executor = { version = "0.2.0", optional = true }
-async-io = { version = "0.2.1", optional = true }
-blocking = { version = "0.6.0", optional = true }
+async-executor = { version = "1.0.0", optional = true }
+async-io = { version = "1.0.1", optional = true }
+blocking = { version = "1.0.0", optional = true }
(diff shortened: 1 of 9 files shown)


stjepang (approved):
Lgtm, thanks!

stjepang:
The codebase has a dozen or so instances of this line of code:

```rust
once_cell::sync::Lazy::force(&crate::rt::RUNTIME);
```

With latest version of smol, it is not needed anywhere except in `src/task/builder.rs`. Can you delete it everywhere except in the `task` module?

This way, if libraries like `async-h1` depend on `async-std` and only use its timers, the executor threadpool will not even be initialized. The only thread that gets started is the "async-io" thread.

Keruspe:
Nice catch indeed, fixed

Keruspe:
Updated to newly released 1.0 versions

dignifiedquire (approved):
thanks

Keruspe:
@stjepang async-executor 1.1.0 fails o android:

```
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
  --> /cargo/registry/src/github.com-1ecc6299db9ec823/async-executor-1.1.0/src/lib.rs:29:47
   |
29 | use std::sync::atomic::{AtomicBool, Ordering, AtomicU64, AtomicUsize};
   |                                               ^^^^^^^^^
   |                                               |
   |                                               no `AtomicU64` in `sync::atomic`
   |                                               help: a similar name exists in the module: `AtomicU8`
```

stjepang:
Should be fixed with async-executor v1.1.1

PR #867, opened 2020-08-28
feat: move executor to async-global-executo

Keruspe:
Based on #866 

This uses async-executor under the hood and is highly inspired by both smol and the async-executor implementation inside of async-std.

Pros:
- less code inside async-std, simpler integration
- shares the same executors with any other users of async-global-executor
- allows spawning local tasks on all of the global executor threads

Cons:
- adds one more extra dependency
- ???

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ rustdoc-args = ["--cfg", "feature=\"docs\""]
 [features]
 default = [
   "std",
-  "async-executor",
+  "async-global-executor",
   "async-io",
   "async-task",
   "blocking",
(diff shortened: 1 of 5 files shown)


yoshuawuyts:
Hi, thanks for this PR!

I def see how sharing the global executor between projects could be useful. But in order to accept this PR I'd like to see:
1. `async-global-executor` dual licensed as MIT / Apache-2.0. If ever need to migrate away from `async-global-executor` we shouldn't need to worry about licensing questions.
2. Granting commit/publish access to @dignifiedquire, @stjepang or myself. Or moving the project to the `async-rs` org. That way if something ever needs fixing that is time-critical and for some reason we can't reach you, we're still able to move ahead.

I hope that makes sense. Thanks heaps!

Keruspe:
Done and Done.

I'm fine with moving it to async-rs too if you see fit

Keruspe:
Rebased on top of the 1.0 bumps

dignifiedquire:
Thanks. I was out for the last two weeks, will be back in the next days and take a look at this.

Keruspe:
Rebased

Keruspe:
CI failure should now be fixed with async-executor 1.1.1

dignifiedquire (approved):
Looks fine to me, given that this is pretty much still the same code, but in a different place for now. I do like having it extracted as its own crate.
Thanks

PR #869, opened 2020-08-30
Fix BufWriter documentation: BufWriters do not flush when dropped.

jimblandy:
This was partially fixed in #586, but there's another sentence later that makes
the same claim.


dignifiedquire (approved):
thanks

PR #870, opened 2020-08-31
Fix the Future impl for JoinHandle on wasm

expenses:
On wasm, a `JoinFuture` uses a `futures_channel::oneshot::Receiver` internally, but as this returns `Result<T, Canceled>`, the code doesn't compile on wasm.


jbr:
I believe this is solving the same problem as https://github.com/async-rs/async-std/pull/863

expenses:
Closing due to https://github.com/async-rs/async-std/pull/863.

PR #879, opened 2020-09-18
Drop unused async-task

Keruspe:
We no longer directly depend on async-task.

While at it, pull in latest async-global-executor version with deadlock fix


yoshuawuyts (approved):
:100:

Keruspe:
with async-executor 1.2.0, async-std currently pulls two versions of async-task. Just making a new release will fix this (we might want to bump async-global-executor to 1.1.0 to ensure we pull in async-executor 1.2.0 and async-task 4.0.0)

PR #882, opened 2020-09-22
fix tokio compatibility

### LINKED ISSUE #881: v1.6.4 (regression) `not currently running on the Tokio runtime.` crash
When using the latest async std version:
`async-std = { version = "1.6.4", features = ["attributes", "tokio02"] }` and `tokio02`

I get an error message deep down saying `not currently running on the Tokio runtime`.

I rely on the `tokio02` feature so I can keep using `https://crates.io/crates/cloud-pubsub` that has a dependency on the Tokio runtime. 

If I use 1.6.3 instead, I don't get the error
`async-std = { version = "=1.6.3", features = ["attributes", "tokio02"] }`

I looked at the latest release note and I see nothing related to a potential `tokio02` change that might explain why I now get this crash. 

I'll attach a sentry crash report and happy to provide further details if necessary

Keruspe:
Move it into async-global-executor

Fixes #881


yoshuawuyts (approved):
Looks great!

Do you perhaps also have a link to the `async-global-executor` changes for reference?

Keruspe:
Sure, https://github.com/async-rs/async-global-executor/commit/7260e60a215e89a0b16d8acee8377a4208c2c62b

PR #894, opened 2020-10-15
io::copy: don't poll the reader again after eof while waiting for the writer to flush

jbr:
I'm not entirely sure if this is a bug or just unexpected behavior, but I thought I'd write a quick PR just in case. Prior to this PR, io::copy continues to poll the reader after it has already returned `Poll::Ready(Ok(0))` in the situation where the writer's flush is not immediately Ready.  This resulted in a bug with a Read implementation that did expected not to be read again after it returned eof. I've fixed that Reader elsewhere, but removing the potential polling might save someone else from some confusion

### DIFF
--- a/src/io/copy.rs
+++ b/src/io/copy.rs
@@ -144,6 +144,7 @@ where
             #[pin]
             writer: W,
             amt: u64,
+            reader_eof: bool
         }
     }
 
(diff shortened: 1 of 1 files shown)


yoshuawuyts (approved):

jbr:
Confirmed locally that the test does in fact fail with the changes to async_std::io::copy stashed

joshtriplett (approved):

Fishrock123 (approved):
I think this is a bug fix.

PR #897, opened 2020-10-18
chore: update dependencies

Keruspe:


examples/surf-web.rs
```diff
@@ -1,15 +1,16 @@
 use async_std::task;
 
-fn main() -> Result<(), surf::Exception> {
+fn main() -> Result<(), http_types::Error> {
```
yoshuawuyts, line 3: `surf::Error` should be able to be used here; I don't think we manually need to pull in `http_types`

Keruspe: Indeed, fixed!

PR #899, opened 2020-10-21
serde: seemingly working Path & PathBuf

maisiliym:


hoijui:
what happened here... was there an issue with this PR (apart from not having a description)?

joshtriplett:
@hoijui I don't know, but this was not intentionally closed; I would guess this got auto-closed by GitHub due to branch reorganization. Apologies that it was missed.

hoijui:
should it maybe be reopened again then, or would it need someone to rebase it first?
... and is it worth doing that (as in, does it have a chance to be accepted then)?

joshtriplett:
Given https://github.com/async-rs/async-std/pull/1099 , probably not. We'd recommend switching to smol.

hoijui:
ouhhh thanks for pointing that out... missed that so far!
(and of course.. then it makes little sense)

Issue #880, opened 2020-09-21
Sender::send polls forever if channel is disconnected. 


maffoo:
`Receiver::recv` fails with a `RecvError` if the channel is disconnected, but `Sender::send` will just spin forever in that case with a future that is always pending. I'd like to have a send method that keeps trying if the channel is full but fails with a `SendError` if the channel disconnects. This could be a new method, something like `Sender::send_if_connected`.

maffoo:
Instead of adding a new method (or perhaps in addition) it might be nice to change the behavior of the existing `send` method to finish immediately and drop messages if the channel is disconnected. It is already the case that messages could be added to the channel buffer and then dropped when the channel is disconnected, so `send` completing is no guarantee that messages will be received.

yoshuawuyts:
I think this is closed by #827 which adds a new channel impl.

Fishrock123:
(Closing, we can re-open if it isn't.)

PR #901, opened 2020-10-29
Return SendError from Sender::send on disconnect.

cmaves:
Currently the `send` method for `Sender<T>` returns a Future that never finishes if the channel is disconnected. This issue was brought up and this fix was proposed in issue #880.
This changes brings the `send` method more inline with what the standard library's sender structs do, where they return an error on disconnect. 
If this pull request's fix isn't appropriate then hopefully it can at least get the ball rolling on something else.


Keruspe:
Superseded by #915 I guess?

cmaves:
Yeah, that PR negates the need for this one.

PR #903, opened 2020-11-01
Fix double drop in StreamExt::cycle

taiki-e:
`StreamExt::cycle` drop `source` field without updating field if `source` field's destructor panicked.

https://github.com/async-rs/async-std/blob/11196c853dc42e86d608c4ed29af1a8f0c5c5084/src/stream/stream/cycle.rs#L47-L48

This is unsound because it causes double drop if the user catches panic in the scope where the `Cycle` is alive.

Repro:

```rust
#[test]
#[should_panic(expected = "second")]
fn test() {
    use async_std::stream::*;
    use futures::task::{noop_waker, Context, Poll};
    use std::{panic, pin::Pin, thread};

    #[derive(Clone, Default)]
    struct S {
        dropped: bool,
    }

    impl Stream for S {
        type Item = ();

        fn poll_next(self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<Option<Self::Item>> {
            Poll::Ready(None)
        }
    }

    impl Drop for S {
        fn drop(&mut self) {
            if !thread::panicking() {
                if std::mem::replace(&mut self.dropped, true) {
                    panic!("second")
                } else {
                    panic!("first")
                }
            }
        }
    }

    let mut s = Box::pin(S::default().cycle());
    let w = noop_waker();
    let cx = &mut Context::from_waker(&w);
    let _ = panic::catch_unwind(panic::AssertUnwindSafe(|| {
        let _ = s.as_mut().poll_next(cx); // <-- dropped the first clone of `s`
    }));
    // <-- dropped the first clone of`s` again since `source` field wasn't updated
}
```

### DIFF
--- a/src/stream/stream/cycle.rs
+++ b/src/stream/stream/cycle.rs
@@ -1,14 +1,19 @@
-use core::mem::ManuallyDrop;
 use core::pin::Pin;
+
+use futures_core::ready;
+use pin_project_lite::pin_project;
 
 use crate::stream::Stream;
 use crate::task::{Context, Poll};
 
-/// A stream that will repeatedly yield the same list of elements.
-#[derive(Debug)]
-pub struct Cycle<S> {
-    orig: S,
-    source: ManuallyDrop<S>,
+pin_project! {
+    /// A stream that will repeatedly yield the same list of elements.
+    #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


yoshuawuyts (approved):
Oh, thanks for catching this!

PR #906, opened 2020-11-02
Document task::spawn's support for blocking tasks.

### LINKED ISSUE #905: Docs for task::spawn must mention support for tasks that block
The async-std blog post ["stop worrying about async blocking"](https://async.rs/blog/stop-worrying-about-blocking-the-new-async-std-runtime/) explains that `spawn_blocking` is no longer needed, and that developers should simply use `spawn` instead and let the library decide whether it should keep the thread in the async pool.

If `task::spawn` is the recommended way to spawn tasks that may block, then users will write code that depends on that, and the design decision becomes something async-std can't revert without introducing bugs in its users. This is all fine (it seems like a good design to me), but non-obvious characteristics covered by async-std's stability guarantees should be documented.

jimblandy:
Fixes #905.

### DIFF
--- a/src/task/spawn.rs
+++ b/src/task/spawn.rs
@@ -23,6 +23,24 @@ use crate::task::{Builder, JoinHandle};
 /// #
 /// # })
 /// ```
+///
+/// # Blocking
+///
+/// It is fine to use `spawn` to start tasks that may carry out long-running
+/// computations, use blocking I/O primitives, or otherwise engage in activities
+/// that mean that polling the task's future may not always return quickly.
+///
+/// This function begins executing the given future on a thread pool shared with
+/// other asynchronous tasks. If polling a particular task takes too long, that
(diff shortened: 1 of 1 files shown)


Keruspe:
Currently this is not true anymore.

I'm working on something to maybe bring it back though

jimblandy:
> Currently this is not true anymore.

Oh! Well, hmm. As I say in #905, once you've promised this behavior, taking it back would break all users, so it's only a feature to ship if you're willing to live with it permanently. If we've had to revert it once already, I hope that doesn't mean it's tricky.

jimblandy:
> Currently this is not true anymore.

@Keruspe Can you link to the issue?

joshtriplett:
Since this isn't currently planned, closing this; if in the future we re-add such support, we'll need to document it as well.

PR #924, opened 2020-12-23
Update async-global-executor and add tokio feature for tokio 1.0

Keruspe:


Cargo.toml
```diff
@@ -59,6 +58,7 @@ alloc = [
   "futures-core/alloc",
   "pin-project-lite",
 ]
+tokio = ["async-global-executor/tokio"]
```
yoshuawuyts, line 61: I'd prefer we add a version qualifier to this as well. There are a lot of language features and traits missing in Rust right now which will almost certainly require breaking changes in Tokio. This is the same reality async-std has operated in since our 1.0 release. Ensuring we're prepared for the eventual Tokio 2.0 release would be wise.

I'd think something like:
```suggestion
tokio1 = ["async-global-executor/tokio"]
```

Keruspe: Sure, I was thinking of `tokio` for 1.0 and `tokio2` for 2.0, but I can make the former `tokio1`

src/lib.rs
```diff
@@ -213,6 +213,15 @@
 //! features = ["attributes"]
 //! ```
 //!
+//! Compatibility with the `tokio` runtime is also simultaneously possible
+//! using the `tokio` Cargo feature:
+//!
+//! ```toml
+//! [dependencies.async-std]
+//! version = "1.7.0"
+//! features = ["tokio"]
```
yoshuawuyts, line 222: This needs updating still

Keruspe: Heh, fair enough

skreborn:
@Keruspe Sorry to bother! All checks are green and the reviews are in - and addressed -; is there anything holding this back from being merged?

Keruspe:
@skreborn should land in the next release within a couple of days ;)

dignifiedquire (approved):

PR #925, opened 2020-12-27
feat: use async-lock for RwLock and Barrier

Koxiaet:
We already use `async-mutex` for `async_std::sync::Mutex`, so it seems `async-std` was going in this direction anyway. Using `async-lock` also gives the advantages of `const` constructors. I've checked and the API does not change.


Koxiaet:
CI's failing, but I can't do anything about it because it would be failing on master if it were re-run (as an update to nightly caused it to fail).

taiki-e:
CI failure will be fixed by #927.

dignifiedquire (approved):
thank you

PR #930, opened 2021-01-06
Fix a typo for [sic] FuturesExt trait

micaelbergeron:
The trait that is being referred to here is called `futures::future::FutureExt`.


dignifiedquire (approved):
thanks

PR #933, opened 2021-01-13
remove deprecated sync::channel

dignifiedquire:


Keruspe (approved):
LGTM.

What about stabilizing the new channels?

dignifiedquire:
> What about stabilizing the new channels?

I thought it would be cleaner to do it in its own PR

PR #937, opened 2021-01-15
Implement Clone for File.

sunfishcode:
Implement `Clone` for `File` so that `File`s can be passed into closures
for use in `spawn_blocking`. `File`'s contents are already wrapped in
`Arc`s, so the implementation of `clone` is straightforward.

This also aligns with `TcpStream` which already implements `Clone` using
its internal `Arc`.


yoshuawuyts (approved):
Looks good! -- We discussed this on Zulip, and happy to merge this once tests pass.

PR #941, opened 2021-01-19
Fix vectored IO for TcpStream

theo3:
This change implements `Write::poll_write_vectored` and `Read::poll_read_vectored` on `TcpStream` using the vectored IO methods on the underlying stream. Previously, the `Write` and `Read` traits' default implementations were used, which just called `poll_write` and `poll_read` once for each `IoSlice`.

This change correctly results in `readv` and `writev` syscalls on Linux.


Keruspe (approved):

dignifiedquire (commented):
thanks

PR #948, opened 2021-01-31
Fix some clippy warnings

brightly-salty:
Fixes some clippy warnings like adding `const` and `#[must_use]` where we can, putting Rust identifiers in tick marks in the documentation, etc.

This may or may not involve breaking changes, so feel free to reject this if it is outside of scope for one PR or not good for right now.

### DIFF
--- a/src/fs/dir_builder.rs
+++ b/src/fs/dir_builder.rs
@@ -36,7 +36,8 @@ impl DirBuilder {
     ///
     /// let builder = DirBuilder::new();
     /// ```
-    pub fn new() -> DirBuilder {
+    #[must_use]
+    pub const fn new() -> DirBuilder {
         #[cfg(not(unix))]
         let builder = DirBuilder { recursive: false };
 
--- a/src/fs/read_dir.rs
+++ b/src/fs/read_dir.rs
@@ -80,7 +80,7 @@ enum State {
 
 impl ReadDir {
     /// Creates an asynchronous `ReadDir` from a synchronous handle.
-    pub(crate) fn new(inner: std::fs::ReadDir) -> ReadDir {
+    pub(crate) const fn new(inner: std::fs::ReadDir) -> ReadDir {
         ReadDir(State::Idle(Some(inner)))
     }
 }
--- a/src/io/read/mod.rs
+++ b/src/io/read/mod.rs
@@ -485,7 +485,7 @@ mod tests {
     #[test]
     fn test_read_by_ref() {
         crate::task::block_on(async {
-            let mut f = io::Cursor::new(vec![0u8, 1, 2, 3, 4, 5, 6, 7, 8]);
+            let mut f = io::Cursor::new(vec![0_u8, 1, 2, 3, 4, 5, 6, 7, 8]);
             let mut buffer = Vec::new();
             let mut other_buffer = Vec::new();
 
--- a/src/io/timeout.rs
+++ b/src/io/timeout.rs
@@ -67,7 +67,7 @@ where
         let this = self.project();
         match this.future.poll(cx) {
             Poll::Pending => {}
-            other => return other,
+            other @ Poll::Ready(..) => return other,
         }
 
         if this.timeout.poll(cx).is_ready() {
--- a/src/path/ancestors.rs
+++ b/src/path/ancestors.rs
@@ -21,7 +21,7 @@ use crate::path::Path;
 ///
 /// [`ancestors`]: struct.Path.html#method.ancestors
 /// [`Path`]: struct.Path.html
-#[derive(Copy, Clone, Debug)]
+#[derive(Clone, Debug)]
 pub struct Ancestors<'a> {
     pub(crate) next: Option<&'a Path>,
 }
--- a/src/path/iter.rs
+++ b/src/path/iter.rs
@@ -32,7 +32,7 @@ impl<'a> Iter<'a> {
     ///
     /// assert_eq!(Path::new("foo/bar.txt"), iter.as_path());
     /// ```
-    pub fn as_path(&self) -> &'a Path {
+    #[must_use] pub fn as_path(&self) -> &'a Path {
         self.inner.as_path()
     }
 }
--- a/src/stream/stream/cycle.rs
+++ b/src/stream/stream/cycle.rs
@@ -42,7 +42,7 @@ where
                 this.source.set(this.orig.clone());
                 this.source.poll_next(cx)
             }
-            item => Poll::Ready(item),
+            item @ Some(..) => Poll::Ready(item),
         }
     }
 }
--- a/src/stream/stream/filter_map.rs
+++ b/src/stream/stream/filter_map.rs
@@ -15,7 +15,7 @@ pin_project! {
 }
 
 impl<S, F> FilterMap<S, F> {
-    pub(crate) fn new(stream: S, f: F) -> Self {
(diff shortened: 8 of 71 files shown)


k-nasa (commented):
Thanks for PR.
This PR has many changes, so it takes time to confirm.
I think we can merge quickly only changing the documentation. 
Can you split PR into changing documentation and refactoring?

Fishrock123:
Needs a rebase.

brightly-salty:
Is it good now @Fishrock123 ?

src/fs/dir_builder.rs
```diff
@@ -36,7 +36,8 @@ impl DirBuilder {
     ///
     /// let builder = DirBuilder::new();
     /// ```
-    pub fn new() -> DirBuilder {
+    #[must_use]
+    pub const fn new() -> DirBuilder {
```
Fishrock123, line 40: What is the benefit of making this const?

src/fs/read_dir.rs
```diff
@@ -80,7 +80,7 @@ enum State {
 
 impl ReadDir {
     /// Creates an asynchronous `ReadDir` from a synchronous handle.
-    pub(crate) fn new(inner: std::fs::ReadDir) -> ReadDir {
+    pub(crate) const fn new(inner: std::fs::ReadDir) -> ReadDir {
```
Fishrock123, line 83: Can this even _be_ const? Even `inner` will be different, how would it generate code to account for that?

src/io/read/mod.rs
```diff
@@ -485,7 +485,7 @@ mod tests {
     #[test]
     fn test_read_by_ref() {
         crate::task::block_on(async {
-            let mut f = io::Cursor::new(vec![0u8, 1, 2, 3, 4, 5, 6, 7, 8]);
+            let mut f = io::Cursor::new(vec![0_u8, 1, 2, 3, 4, 5, 6, 7, 8]);
```
Fishrock123, line 488: This seems like churn...

src/io/timeout.rs
```diff
@@ -67,7 +67,7 @@ where
         let this = self.project();
         match this.future.poll(cx) {
             Poll::Pending => {}
-            other => return other,
+            other @ Poll::Ready(..) => return other,
```
Fishrock123, line 70: Unnecessary, if `Poll` were to expand there would be issues anyways and that would almost certainly have to happen in a rust edition.

src/path/ancestors.rs
```diff
@@ -21,7 +21,7 @@ use crate::path::Path;
 ///
 /// [`ancestors`]: struct.Path.html#method.ancestors
 /// [`Path`]: struct.Path.html
-#[derive(Copy, Clone, Debug)]
```
Fishrock123, line 24: This is a breaking change. What is the reasoning?

src/path/iter.rs
```diff
@@ -32,7 +32,7 @@ impl<'a> Iter<'a> {
     ///
     /// assert_eq!(Path::new("foo/bar.txt"), iter.as_path());
     /// ```
-    pub fn as_path(&self) -> &'a Path {
+    #[must_use] pub fn as_path(&self) -> &'a Path {
```
Fishrock123, line 35:
```suggestion
    #[must_use]
    pub fn as_path(&self) -> &'a Path {
```

src/stream/stream/cycle.rs
```diff
@@ -42,7 +42,7 @@ where
                 this.source.set(this.orig.clone());
                 this.source.poll_next(cx)
             }
-            item => Poll::Ready(item),
+            item @ Some(..) => Poll::Ready(item),
```
Fishrock123, line 45: Unnecessary

src/stream/stream/filter_map.rs
```diff
@@ -31,13 +31,13 @@ where
         let this = self.project();
         let next = futures_core::ready!(this.stream.poll_next(cx));
         match next {
-            Some(v) => match (this.f)(v) {
-                Some(b) => Poll::Ready(Some(b)),
-                None => {
+            Some(v) => (this.f)(v).map_or_else(
```
Fishrock123, line 34: Was this clippy? I think this is harder to read.

src/stream/stream/find_map.rs
```diff
@@ -30,13 +30,13 @@ where
         let item = futures_core::ready!(Pin::new(&mut *self.stream).poll_next(cx));
 
         match item {
-            Some(v) => match (&mut self.f)(v) {
-                Some(v) => Poll::Ready(Some(v)),
-                None => {
+            Some(v) => (&mut self.f)(v).map_or_else(
```
Fishrock123, line 33: Same

tests/rwlock.rs
```diff
@@ -49,7 +49,8 @@ fn smoke() {
         let lock = RwLock::new(());
         drop(lock.read().await);
         drop(lock.write().await);
-        drop((lock.read().await, lock.read().await));
```
Fishrock123, line 52: This is behaviorally different than the change. Please undo this. This will wait until both have read access before dropping them.

Fishrock123 (commented):
All the `const`s worry me, I don't know enough what implications those have.

There are some other changes that should be made or undone.

src/stream/stream/filter_map.rs
brightly-salty: This was clippy's suggestion (or after a series of suggestions), removing unnecessary duplication

src/fs/dir_builder.rs
brightly-salty: The benefit of making this const is that `DirBuilder::new` can now be used in const declarations, as well as anything that depends on it also now has a chance to be made const.

src/fs/read_dir.rs
brightly-salty: I'm not sure I understand your reasoning here

src/io/read/mod.rs
brightly-salty: Just more readable, and idiomatic/conventional to have an underscore in Rust. I agree that its not strictly necessary at all

tests/rwlock.rs
brightly-salty: Are you sure about this? clippy said it was a useless call, so it might be a clippy bug if you're right

Fishrock123: I am absolutely sure.

`(lock.read().await, lock.read().await)` means two immutable `read` locks are acquired and simultaneously alive before drop.

brightly-salty: Okay, I'll revert this

src/stream/stream/cycle.rs
brightly-salty: I think the goal here is to make clearer to readers that the only possibility is `Some(..)`, rather than being unclear about the possibly multiple cases it covers.

src/path/ancestors.rs
brightly-salty: I think this might have been a mistake, I'll revert.

src/io/timeout.rs
brightly-salty: I think the goal here is to make clearer to readers that the only possibility is `Poll::Ready(..)`, rather than being unclear about the possibly multiple cases it covers.

jayvdb:
I've run `cargo clippy` and afaics all of the changes in this PR are no longer warnings.

PR #957, opened 2021-02-16
io: export write::*

Keruspe:
We weren't exporting WriteExt.
We already do that with read::*


dignifiedquire (approved):
seems reasonable

Fishrock123 (approved):

Issue #730, opened 2020-03-19
impl IntoRawFd for TcpStream closes the fd


boardwalk:
This code gets takes the TcpStream by value, gets its fd, and immediately destroys the TcpStream, likely destroying the fd:

```rust
    impl IntoRawFd for TcpStream {
        fn into_raw_fd(self) -> RawFd {
            self.as_raw_fd()
        }
    }
```

It looks like the other impls of IntoRawFd do `self.watcher.into_inner().into_raw_fd()` which seems correct comparatively.

rcmgleite:
Even the documentation states that this would transfer ownership of the underlying fd.
https://docs.rs/async-std/1.6.0/async_std/os/unix/io/trait.IntoRawFd.html.

Any updates?

PR #961, opened 2021-02-27
Added TryFrom<Tcp/UnixStream> for RawRd impl

cmaves:
There is no way to get an owned RawFd from TcpStream right now because IntoRawFd is broken (see issue #730). 

I don't think it is possible to correctly implement IntoRawFd in a way that is semantically consistent with how IntoRawFd works in the standard library. `UnixStream` tries to implement this by duping the underlying fd. I'd argue this is semantically inconsistent because it can fail (for example you hit your process fd limit). Also IMO duping is a bad option because it effectively allows you to accidentally create multiple "mutable" references to the same underlying kernel stream, allowing for objects to interfere with each other.

This leaves us with TryFrom<TcpStream> as the best option.

### DIFF
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -1,4 +1,3 @@
-use std::convert::TryFrom;
 use std::io::{IoSlice, IoSliceMut};
 use std::net::SocketAddr;
 use std::pin::Pin;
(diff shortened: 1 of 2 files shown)


joshtriplett:
I'd like to propose holding off on this, in favor of considering an interface based on the new BorrowedFd or OwnedFd types.

cmaves:
Seems fair. I'm gonna go ahead and close this PR since it has been stagnant for so long and may not be the correct path forward anymore.

PR #962, opened 2021-03-04
Fix stdin.rs comments to say "read" instead of "write".

sunfishcode:
This just fixes a few comments that appear to have been copied and
pasted from stdout.rs.

### DIFF
--- a/src/io/stdin.rs
+++ b/src/io/stdin.rs
@@ -16,7 +16,7 @@ use crate::utils::Context as _;
 /// ### Note: Windows Portability Consideration
 ///
 /// When operating in a console, the Windows implementation of this stream does not support
-/// non-UTF-8 byte sequences. Attempting to write bytes that are not valid UTF-8 will return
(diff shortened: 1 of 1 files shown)


sunfishcode:
The CI failure is unrelated:
```
error: unknown `doc` attribute `unstable`
   --> src/utils.rs:145:46
    |
145 |               #[cfg_attr(feature = "docs", doc(unstable))]
    |                                                ^^^^^^^^
    | 
   ::: src/lib.rs:336:1
    |
336 | / cfg_unstable_default! {
337 | |     #[doc(inline)]
338 | |     pub use std::{write, writeln};
339 | | }
    | |_- in this macro invocation
```

sunfishcode:
Rebased, and now the CI is green.

Fishrock123 (approved):
I believe this is correct, though some stdin pipes can also be written to but I don't think that async-std supports that.

PR #965, opened 2021-03-13
Fix WriteFmtFuture not taking into account already written bytes (#964)

### LINKED ISSUE #964: multiple write! on async_std::fs::File can trigger infinite process
By writing two short strings with separate write! invocations we can trigger an infinite process
that will continue writing the start of the second string until the process is killed or the disk is full.

```
    println!("starting ...");

    let mut file = OpenOptions::new()
                .append(true)
                // .write(true)    // also happens with write, write+truncate
                // .truncate(true)
                .create(true)
                .open(filename).await?;

    write!(
        &mut file, 
        "{}", 
        a, 
    ).await?;
    write!(
        &mut file, 
        "{}", 
        b, 
    ).await?;
    
    println!("done.");
```
As long as the second string `b` is shorter than the first one (`a`) we can't reproduce the problem, with a longer than 8 bytes we can reliably reproduce the problem.

Full reproduction in this gist: <https://gist.github.com/hn3000/100460d1e239f22bf3862f1a8f8cfd92>

(Also, I am not sure how to write a failing unit test because the error case does not terminate.)

rkarp:
Fixes #964. The problem is that `amt` (amount of bytes already written) is ignored, so on every poll the whole buffer is sent again if it was not possible to write the whole buffer on the first poll. Interestingly, for some cases other than `File`, all bytes get written on the first poll, which is probably why this bug hasn't been noticed before.

The actual fix is just line 44, but converting from `u64` to `usize` for array indexing requires a checked conversion and I couldn't see any reason in the code why `amt` has to be `u64`, so I changed the type to avoid something like `try_into().unwrap()`.

### DIFF
--- a/src/io/write/write_fmt.rs
+++ b/src/io/write/write_fmt.rs
@@ -11,7 +11,7 @@ pub struct WriteFmtFuture<'a, T: Unpin + ?Sized> {
     pub(crate) writer: &'a mut T,
     pub(crate) res: Option<io::Result<Vec<u8>>>,
     pub(crate) buffer: Option<Vec<u8>>,
-    pub(crate) amt: u64,
+    pub(crate) amt: usize,
 }
 
 impl<T: Write + Unpin + ?Sized> Future for WriteFmtFuture<'_, T> {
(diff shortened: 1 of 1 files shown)


yoshuawuyts (approved):
This is great, thanks so much!

PR #967, opened 2021-03-23
size_hint setup for streams

mich2000:
PR regarding the issue 953. This PR needs a review because it is possible that I have made a lot of errors, because it is not always easy to write size_hint code.

### DIFF
--- a/src/fs/read_dir.rs
+++ b/src/fs/read_dir.rs
@@ -109,4 +109,15 @@ impl Stream for ReadDir {
             }
         }
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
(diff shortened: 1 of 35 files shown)


taiki-e:
I recommend seeing how the same method is implemented in `futures`: https://github.com/rust-lang/futures-rs/pull/1853

mich2000:
Okay I've taken a look an implemented those I thought would make sense.

mich2000:
There is a problem related to the test pipeline powerpc-unknown-linux-gnu, this pipeline needs SO_INCOMING_CPU constant from the libc to complete the build. But with libc 0.2.91 it doesn't have this constant, should we open a issue to address this issue? So that it won't hinder other PR's.

PR #968, opened 2021-03-31
Change Incoming impls to only do one allocation

erickt:
This modifies net::tcp::Incoming and os::net::unix::Incoming to only do one allocation, rather than an allocation for each connection.

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -1,6 +1,6 @@
 use std::fmt;
-use std::future::Future;
 use std::net::SocketAddr;
+use std::net::TcpStream as StdTcpStream;
 use std::pin::Pin;
 
 use async_io::Async;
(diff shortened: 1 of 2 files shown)


erickt:
The powerpc cross compile seems unrelated to my change. It's erroring out on:

```
error[E0425]: cannot find value `SO_INCOMING_CPU` in crate `libc`
    --> /cargo/registry/src/github.com-1ecc6299db9ec823/socket2-0.4.0/src/sys/unix.rs:1201:69
     |
1201 |             getsockopt::<c_int>(self.inner, libc::SOL_SOCKET, libc::SO_INCOMING_CPU)
     |                                                                     ^^^^^^^^^^^^^^^ not found in `libc`

```

skade:
What worries me a little is that all other `unknown-linux-gnu` were cancelled. Do you have a linux machine at hand currently?

PR #986, opened 2021-09-20
fix wrong link.

### LINKED ISSUE #954: Missing or wrong link in documentation to Arc
the [docs page](https://docs.rs/async-std/1.8.0/async_std/task/index.html) contains a link to Arc which is `https://docs.rs/async-std/1.8.0/async_std/gsync/struct.Arc.html` when I believe it should be `https://docs.rs/async-std/1.8.0/async_std/sync/struct.Arc.html`.

kokihonda:
fix #954


yoshuawuyts:
Thanks for this PR! This appears to be a dupe of #987, so I think this can be closed.

PR #987, opened 2021-09-20
fix wrong link.

### LINKED ISSUE #954: Missing or wrong link in documentation to Arc

kokihonda:
fixes #954  The link to Arc in v1.9.0 was also wrong, so I fixed that as well.

### DIFF
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -103,7 +103,7 @@
 //! the desired task name to [`Builder::name`]. To retrieve the task name from within the
 //! task, use [`Task::name`].
 //!
-//! [`Arc`]: ../gsync/struct.Arc.html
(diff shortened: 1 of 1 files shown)


yoshuawuyts (approved):
thank you!

kokihonda:
Thanks for the response!
I'm new to Github, so sorry if I'm wrong.
Prior to v1.10.0, there are typos in the links in the documentation. So if you want to fix the wrong links in earlier versions, I feel you need to replace the tags.

PR #994, opened 2021-10-13
Match error message from sync File::create in async File::create

### LINKED ISSUE #993: Error information from fs operations not displayed when printing errors
With this code:

```
	let file_name = Path::new("/tmp/does_not_exist/test");

	match fs::File::create(file_name) {
		Ok(_) => (),
		Err(e) => {
			println!("Sync: Error opening file {}: {}", file_name.display(), e);
		}
	};

		match async_std::fs::File::create(file_name).await {
			Ok(_) => (),
			Err(e) => {
				println!("Async: Error opening file {}: {}", file_name.display(), e);
			}
		};
	});
```

We get this output:

```
Sync: Error opening file /tmp/does_not_exist/test: No such file or directory (os error 2)
Async: Error opening file /tmp/does_not_exist/test: could not create `/tmp/does_not_exist/test`
```

There are two problems with this:
- The async version doesn't output the actual reason that the operation failed
- The async version adds the filename as context. While arguably the std version should do that as well - it doesn't, meaning that if you use the async API in the same way as the sync API, it prints the filename twice. 

The first point is the most important. Although more error information is available by inspecting `kind` or `source`, I think the `Display` implementation should include the actual cause of the error.

I don't view the second point as critical, as long as async_std supplies the context for every relevant API and also documents that divergance from `std::fs` on the doc page for `async_std::fs` (although I guess a major selling point of async-std is that the API mirrors std).

jaztec:
Removing the context will result in the original `io::Error` and will display as it would in the std library

Fixes #993

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -154,7 +154,6 @@ impl File {
         let path = path.as_ref().to_owned();
         let file = spawn_blocking(move || {
             std::fs::File::create(&path)
-                .context(|| format!("could not create `{}`", path.display()))
         })
         .await?;
         Ok(File::new(file, true))
@@ -902,5 +901,16 @@ mod tests {
             }).await;
             assert_eq!(len as u64, file.metadata().await.unwrap().len());
         });
+    }
+
+    #[test]
+    fn async_file_create_error () {
+        let file_name = Path::new("/tmp/does_not_exist/test");
+        let expect = std::fs::File::create(file_name).unwrap_err();
+
+        crate::task::block_on(async move {
+            let actual = File::create(file_name).await.unwrap_err();
+            assert_eq!(format!("{}", expect), format!("{}", actual));     
+        })
     }
 }


Fishrock123:
(I think this should actually have both - the std library should include the context and the error, ideally, and then we'd copy that.)

jaztec:
It might, I consider myself to green to rust to already have a strong opinion on this. That's why I jumped the ~"good start issue"
 chance :).

But technically, when the std library would add context, omitting it inside the async library would still fix #993, it mostly was about the 2 libraries not being in sync.

heroin-moose:
This PR kinda makes sense because it's not obvious which async_std functions will provide context and which won't. For example, `UdpSocket::bind()` or `TcpStream:connect()` provide only errno so it's up to the user to say what happened. And generally speaking the user has much more context to provide anyway.

PR #995, opened 2021-10-18
Add TcpListener::into_incoming

### LINKED ISSUE #981: Transform TcpListener into Stream of connections
There is [`TcpListener::incoming`](https://docs.rs/async-std/1.9.0/async_std/net/struct.TcpListener.html#method.incoming) that provides this functionality, but it has a lifetime. I cannot simply pass the stream around and manipulate it, because it has `'a` lifetime bound to the `TcpListener`. An alternative that takes over ownership would be appreciated.

Note that this is symmetric to the stdlib, where the same API suffers from the same problem.

piegamesde:
Closes #981.

- The other methods delegate to `async_io` for the implementation. I suggest implementing it here right now until `async_io` adds the method too.
- I used `impl Stream<_>` as return type instead of a custom `IntoIncoming` struct, because the code is cleaner (IMO) and also removes one level of pointer indirection. Also, this is what `async_io` does.

CC @yoshuawuyts

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -150,6 +150,45 @@ impl TcpListener {
         Incoming {
             incoming: Box::pin(self.watcher.incoming()),
         }
+    }    
+
+    /// Turn this into a stream over the connections being received on this
+    /// listener.
+    ///
+    /// The returned stream is infinite and will also not yield
+    /// the peer's [`SocketAddr`] structure. Iterating over it is equivalent to
+    /// calling [`TcpListener::accept`] in a loop.
+    ///
+    /// ## Examples
+    ///
+    /// Merge the incoming connections of multiple sockets into one [`Stream`]:
+    ///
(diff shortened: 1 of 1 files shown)


src/net/tcp/listener.rs
```diff
@@ -150,6 +150,45 @@ impl TcpListener {
...
+    /// let incoming = futures::stream::select_all(
+    ///     listeners.into_iter()
+    ///         .map(TcpListener::into_incoming)
+    ///         .map(Box::pin)
+    /// );
+    /// #
+    /// # Ok(()) }) }
+    /// ```
+    #[cfg(feature = "unstable")]
+    pub fn into_incoming(self) -> impl Stream<Item = io::Result<TcpStream>> + Send {
```
yoshuawuyts, line 187: This ideally would be a named future; but perhaps we can fix that in a follow-up  PR?

yoshuawuyts (approved):

PR #996, opened 2021-10-29
💄 Make TODO chapters in Book as unclickable

s373r:
To prevent opening empty pages this PR makes links unclickable from visual and interaction perspectives. 

Please check the following preview images:

Before             |  After
:-------------------------:|:-------------------------:
before  | after

Inspired by https://github.com/rust-lang/async-book


Fishrock123 (approved):

s373r:
Hello all!

I've made a little investigation why we got a fail as a result of this CI job (and more interested for me why I couldn't get that locally)

### Results
- ["You can't have an empty link."](https://github.com/async-rs/async-std/runs/4071567547?check_suite_focus=true#step:5:8) error check [added in 0.1.0 version of mdBook](https://github.com/rust-lang/mdBook/issues/576)
- [In 0.4.0 version](https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-040) that error check has been [removed](https://github.com/rust-lang/mdBook/pull/1153)
- CI [uses pinned 0.3.7 version](https://github.com/async-rs/async-std/runs/4071567547?check_suite_focus=true#step:4:17) that leads to the error

### Some thoughts

- Since CI is an internal thing I cannot update the version from 0.3.7 to 0.4.X -- also, I don't know the real reasons why the version is pinned currently (it may be some needed workaround 🤔 )

### Possible next steps

1) The version can be updated if we have no strong reason not to do so 

OR

2) This PR can wait for version updating and I'll update it later if needed

---

@Fishrock123 & @yoshuawuyts please see ^^^  if something seems unclear feel free to ping me

s373r:
Added 40741ff

joshtriplett:
We still want to pin a version, we should just update the pin from 0.3 to 0.4.

(Also, there's a typo: "mdbookk" should be "mdbook".)

PR #997, opened 2021-11-07
stream: correct iterators in doc examples

dmitris:
Correct the iterators used in examples to produce the described output.

Currently in https://docs.rs/async-std/1.10.0/async_std/stream/index.html#while-let-loops-and-intostream the text after the code snippet says "This will print the numbers one through five" but because the code uses `stream::repeat(1u8).take(5)`, it will print `1` five times instead - modified to use `from_iter(1..6)`.  Similarly, in the Iinfinity" section https://docs.rs/async-std/1.10.0/async_std/stream/index.html#infinity, changed `stream::repeat(1u8)` to `stream::from_iter(0u8)` to produce the described result, "print the numbers `0` through `4`".

I believe the Section "while let Loops and IntoStream" https://docs.rs/async-std/1.10.0/async_std/stream/index.html#while-let-loops-and-intostream has more problems.  Is the `into_stream` used in this example? The explanation says: "we never called anything on our vector to produce a stream" - but as far as I see, there is no vector in the example, only `async_std::stream::Repeat` created with `repeat` - or a Stream in the modified example. Also `IntoStream` is supported only on unstable - I wonder if it would be better to rename the section to `while let Loops and FromIter` and talk about conversion of iterators into streams with `from_iter` instead of `into_stream`.

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -164,7 +164,7 @@
 //! #
 //! # use async_std::prelude::*;
 //! # use async_std::stream;
-//! let mut values = stream::repeat(1u8).take(5);
+//! let mut values = stream::from_iter(1u8..6);
 //!
 //! while let Some(x) = values.next().await {
 //!     println!("{}", x);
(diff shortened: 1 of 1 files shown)


dmitris:
@yoshuawuyts would appreciate if you could review the suggested change, or point out someone who could do it.  Thanks!

yoshuawuyts (approved):
Looks good; thanks

dmitris:
thanks @yoshuawuyts!  

When will it make its way to https://docs.rs - after the next crate release?

Do you agree with the idea of changing the section header "while let Loops and IntoStream" to "while let Loops and FromIter" and talking about `from_iter()` instead of (or in addition to) `into_stream`? I can raise a PR if that sounds right in principle.

PR #1012, opened 2022-04-23
Remove unused num_cpus dependency (handled by async_global_executor)

joshtriplett:
async-std doesn't use num_cpus directly, only via async_global_executor.


Keruspe:
Thanks!

PR #1014, opened 2022-04-27
Add `TryFrom` impls to convert async types to corresponding sync types

joshtriplett:
Add `TryFrom` implementations to convert `TcpListener`, `TcpStream`,
`UdpSocket`, `UnixDatagram`, `UnixListener`, and `UnixStream` to their
synchronous equivalents, including putting them back into blocking mode.

This does not add such an impl for `File`, because `std::fs::File` does not
provide a `set_nonblocking` method.

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -252,6 +252,16 @@ impl From<std::net::TcpListener> for TcpListener {
     }
 }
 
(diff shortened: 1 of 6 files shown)


yoshuawuyts (approved):
These all look good! - there's a WASM build failure, but I suspect that's unrelated to this patch. Thanks for this!

PR #1017, opened 2022-06-01
Stabilize `std::task::spawn_blocking`

joshtriplett:
Given how widely used spawn_blocking is within async-std itself, and how
useful it is for building other APIs, I think it makes sense to offer it
just as we do `spawn`, even though it isn't standard in Rust itself.


joshtriplett:
https://github.com/async-rs/async-std/pull/1019 fixes the unrelated CI failure.

yoshuawuyts (approved):

PR #1018, opened 2022-06-01
Re-export `task::spawn` and `task::spawn_blocking` at top-level

joshtriplett:
These are fundamental operations that crates will use constantly; make
it possible to write `async_std::spawn` or `async_std::spawn_blocking`
to make examples and tests easier to write without requiring as many
`use` lines.

This PR depends on https://github.com/async-rs/async-std/pull/1017 so that it
can re-export both.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -318,6 +318,11 @@ cfg_default! {
     pub mod net;
     #[cfg(not(target_os = "unknown"))]
     pub(crate) mod rt;
+
+    #[cfg(not(target_os = "unknown"))]
+    pub use crate::task::spawn;
+    #[cfg(not(target_os = "unknown"))]
+    pub use crate::task::spawn_blocking;
 }
 
 cfg_unstable! {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://book.async.rs/overview
 # [Unreleased]
 
 ## Added
+
+- `std::task_spawn_blocking` is now stabilized. We consider it a fundamental API for bridging between blocking code and async code, and we widely use it within async-std's own implementation.
+
 ## Removed
 ## Changed
 
--- a/src/task/mod.rs
+++ b/src/task/mod.rs
@@ -160,11 +160,7 @@ cfg_default! {
     mod task_locals_wrapper;
 
     #[cfg(not(target_os = "unknown"))]
-    #[cfg(any(feature = "unstable", test))]
     pub use spawn_blocking::spawn_blocking;
-    #[cfg(not(target_os = "unknown"))]
-    #[cfg(not(any(feature = "unstable", test)))]
-    pub(crate) use spawn_blocking::spawn_blocking;
 }
 
 cfg_unstable! {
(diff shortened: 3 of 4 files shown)


joshtriplett:
Now passes CI other than the unrelated failure fixed by https://github.com/async-rs/async-std/pull/1019 .

src/lib.rs
```diff
@@ -318,6 +318,11 @@ cfg_default! {
     pub mod net;
     #[cfg(not(target_os = "unknown"))]
     pub(crate) mod rt;
+
+    #[cfg(not(target_os = "unknown"))]
+    pub use crate::task::spawn;
+    #[cfg(not(target_os = "unknown"))]
+    pub use crate::task::spawn_blocking;
```
yoshuawuyts, line 325: A goal we had when designing `async-std`'s was to create a module hierarchy which resembled the hierarchy of the stdlib - but expanding on it with async-only capabilities in certain locations. This was with the expectation that one day we might add these APIs into the stdlib itself. And at least my hope is that we may eventually enable a diff like this:

```diff
- use async_std::task::spawn;
+ use std::task::spawn;
```

You're probably in a better position to answer this, but I kind of assumed that in the stdlib we would likely want to scope the `spawn` family of functions to submodules, and likely not expose them directly from `std`'s root? Meaning that even if we create an `async_std::spawn`, we'd unlikely ever have a `std::spawn`?

## Alternatives

I do think you're touching on an important insight here though: `spawn`ing tasks is a very common operation, and making that convenient to perform seems important. I've recently been experimenting with [making `spawn` a method on `Future`](https://blog.yoshuawuyts.com/postfix-spawn/). This solves the same problem by not requiring any imports at all:

```rust
use tasky::prelude::*;
 
    let res = async { "nori is a horse" }
        .spawn()               // <- `FutureExt::spawn`
        .name("meow".into())
        .await;
    assert_eq!(res, "nori is a horse");
})
```
I've been contemplating whether we could add this to `async-std` as part of `FutureExt`, possibly even deprecating `task::spawn` as part of it. Though there are still a few open questions with this:

1. The path for inclusion in the stdlib seems less clear. The base trait lives in `core::future::Future`, but this extension may only be possible on `std`? Perhaps there is something with `contexts` we could provide, or perhaps the unification of `core` and `std` would help? But I'm not sure what the right path for the stdlib here would be.
2. The [`tasky`](https://docs.rs/tasky/latest/tasky/) prototype of this idea has both `spawn` and `spawn_local` variants of this method. This is probably worth its own blog post, but `spawn_local` doesn't appear to be the right abstraction. Instead we need to find a way to define a spawn method with "the captured values from the scope can be sent to another thread, but once the future is constructed it will no longer move between threads". Much like the `thread::spawn` API works. I'm not sure yet how to do this, as it seems like it needs to be implemented at the `async-task` layer.
3. I'm not sure what to do about `spawn_blocking`. It doesn't really make sense to define it as `FutureExt::spawn_blocking` since it's only intended to be used for non-future workloads. Having a `T::spawn_blocking` feels wrong too. How common is it to use this method? What is the right way to expose this functionality?

joshtriplett: You're right, we're unlikely to ever have a `std::spawn`, so we shouldn't make this change.

In theory, we could make `spawn_blocking` take a future, since it may want to run a mix of sync and async work.

But whether we do that or not, I'm not sure we want to have `spawn` as a method on futures.

joshtriplett:
Closing, since this API wouldn't match future std APIs.

PR #1019, opened 2022-06-02
Fix CI errors about unused-macro-rules

joshtriplett:
float_product and float_sum had unused rules, because they weren't
successfully using their second branch, and weren't successfully
defining wrapping types.

That then led to the discovery that those types *can't* be defined,
because std doesn't actually define any operations on `Wrapping<f32>` or
`Wrapping<f64>`. So, drop those portions of the float macros.

Fix that, and in the process, unify the integer and float macros.


yoshuawuyts (approved):
Fantastic, thank you!

PR #1022, opened 2022-06-02
Remove rustfmt.toml - zero changes to formatting

joshtriplett:
rustfmt's `version = "Two"` option doesn't result in any formatting
changes anywhere in async-std, but it prevents formatting with a stable
toolchain. Remove rustfmt.toml and allow formatting to work on stable
rustfmt.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,1 +1,0 @@
-version = "Two"


joshtriplett:
Definitely seeking comments on this one, in case I've missed something.

yoshuawuyts (approved):
I'm not sure what's going on with this either, so happy to accept this patch!

PR #1036, opened 2022-08-11
Implement I/O-safe traits on types

notgull:
This PR adds a new feature, `io_safety`, which requires Rust 1.63 or higher. This trait implements the `AsFd/AsHandle/AsSocket` family of functions on `async-std`'s types. In addition, several types also implement `From<OwnedFd/OwnedHandle/OwnedSocket>` and `Into<OwnedFd/OwnedHandle/OwnedSocket>`.


.github/workflows/ci.yml
```diff
@@ -108,6 +108,21 @@ jobs:
       with:
         command:  check
         args: --all --features tokio02
+ 
+  check_io_safety_feature:
+    name: Check io_safety feature
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
Fishrock123, line 117: Doesn’t support MacOS?

src/fs/file.rs
```diff
@@ -415,6 +415,8 @@ impl From<std::fs::File> for File {
 cfg_unix! {
     use crate::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 
+    
+
```
Fishrock123, line 419:
```suggestion
```

src/fs/file.rs
```diff
@@ -432,10 +434,36 @@ cfg_unix! {
...
+            }
+        }
+
+        impl From<OwnedFd> for File {
+            fn from(fd: OwnedFd) -> Self {
+                std::fs::File::from(fd).into()
+            }
+        }
+
+        impl From<File> for OwnedFd {
```
Fishrock123, line 457: Should this be `TryFrom`?

src/fs/file.rs
```diff
@@ -955,3 +1009,5 @@ mod tests {
         })
     }
 }
+
+const ARC_TRY_UNWRAP_EXPECT: &str = "cannot acquire ownership of the file handle after drop";
```
Fishrock123, line 1013: Move near the top of file please. Also please insert end of file newlines.

src/net/tcp/listener.rs
```diff
@@ -306,4 +328,26 @@ cfg_windows! {
...
+
+        impl From<OwnedSocket> for TcpListener {
+            fn from(fd: OwnedSocket) -> TcpListener {
+                std::net::TcpListener::from(fd).into()
+            }
+        }
+
+        impl From<TcpListener> for OwnedSocket {
+            fn from(listener: TcpListener) -> OwnedSocket {
+                listener.watcher.into_inner().unwrap().into()
```
Fishrock123, line 349: Should this be `TryFrom`?

src/fs/file.rs
notgull: I was mirroring the previous implementation for `IntoRawFd` for this type. If you would think that it would be better as `TryFrom` I will oblige.

.github/workflows/ci.yml
notgull: The changes I wrote only impact `cfg(unix)` and `cfg(windows)`, so I included a Unix platform (`ubuntu-latest`) and a Windows platform (`windows-latest`). If you would like me to also test on MacOS I can.

src/net/tcp/listener.rs
notgull: I was mirroring the `IntoRawSocket` implementation for this type. Should this be `TryFrom`? If so, I can make it.

yoshuawuyts:
Tagging in @sunfishcode who likely has thoughts on how we should approach this.

sunfishcode:
> _edit:_ I believe file and stdio types may actually be unsound to implement these traits on on windows?

Are you referring to the `OVERLAPPED` issue in Windows? It's my understanding that that's now [fixed](https://github.com/rust-lang/rust/pull/95469).

sunfishcode:
This change looks good to me. Overall, the intention for `From<OwnedFd>` and `From<T> for OwnedFd` impls is to mirror existing `FromRawFd` and `IntoRawFd` impls. In practice, the way `FromRawFd` and `IntoRawFd` are used, users assume they have ownership-transfer semantics, so if there any issues with soundness with the new traits, they'd likely already present with the existing traits.

sunfishcode:
Also, it's currently the case that async-std doesn't currently call `ReadFile` or `NtReadFile` itself; it currently uses [`std::fs::File::read` from within `spawn_blocking`](https://github.com/async-rs/async-std/blob/dfdf56cc9a97c853e7e6caa775fdb439e87a649c/src/fs/file.rs#L719), so async-std wouldn't be doing anything that std itself isn't already doing.

notgull:
What is the current status of this PR?

notgull:
@yoshuawuyts Is there anything I can do to move this PR forwards?

src/fs/file.rs
```diff
@@ -413,7 +415,7 @@ impl From<std::fs::File> for File {
 }
 
 cfg_unix! {
-    use crate::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
+    use crate::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
```
joshtriplett, line 418:
```suggestion
    use crate::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
```

src/fs/file.rs
```diff
@@ -432,10 +434,36 @@ cfg_unix! {
...
+            }
+        }
+
+        impl From<File> for OwnedFd {
+            fn from(val: File) -> OwnedFd {
+                let file = val.file.clone();
+                drop(val);
+                Arc::try_unwrap(file)
+                    .expect(ARC_TRY_UNWRAP_EXPECT)
+                    .into()
```
joshtriplett, line 463: Rather than duplicating this code (and the message), I'd suggest having a single internal helper function that returns the internal File, and then calling that from both of these trait impls.

joshtriplett:
Commented on a couple of minor issues. Happy to merge this with those fixed.

src/fs/file.rs
```diff
@@ -429,11 +440,29 @@ cfg_unix! {
...
+
+        impl From<OwnedFd> for File {
+            fn from(fd: OwnedFd) -> Self {
+                std::fs::File::from(fd).into()
+            }
+        }
+
+        impl From<File> for OwnedFd {
+            fn from(val: File) -> OwnedFd {
+                self.into_std_file()
```
joshtriplett, line 464:
```suggestion
                self.into_std_file().into()
```

Keruspe:
This is completely broken :sob: 
The feature is called io_safety and the cfg checks are done using io-safety as a feature name, so the code actually never got tested and doesn't even build.

joshtriplett:
@Keruspe Argh. :(

Would welcome PRs fixing this; that should happen before a new release.

Keruspe:
Should be fixed in master now, CI is green now

PR #1069, opened 2023-11-23
update async-* dependencies

Keruspe:


.github/workflows/ci.yml
```diff
@@ -132,8 +132,8 @@ jobs:
         target:
           - i686-unknown-linux-gnu
           - powerpc-unknown-linux-gnu
-#          - powerpc64-unknown-linux-gnu
-          - mips-unknown-linux-gnu
+          - powerpc64-unknown-linux-gnu
+#          - mips-unknown-linux-gnu
```
notgull, line 136: Is MIPS broken for the latest smol version?

Keruspe: See https://github.com/async-rs/async-std/actions/runs/7005887907/job/19056537762?pr=1069

Fishrock123 (approved):

Keruspe:
Async-Channel 2 breaks some public API we reexport so we cannot bump it before async-std 2

PR #1083, opened 2024-08-20
Prepare 1.13.0 release

Keruspe:


Keruspe:
If nobody objects I'll probably go ahead with it some time next week (updating the date accordingly)

PR #1104, opened 2026-02-16
Clarify that crate is deprecated not discontinued

xtqqczze:
cc: @joshtriplett


xtqqczze:
I guess this means https://github.com/rustsec/advisory-db/blob/main/crates/async-std/RUSTSEC-2025-0052.md should be updated too.

joshtriplett:
What is the intended distinction you're trying to make here?

